PR #207, opened 2014-10-26
Implement progress to NanAsyncWorker.

brett19:
This builds on NanAsyncWorker by providing a NanAsyncProgressWorker which can dispatch progress information back to the originator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ LINT_SOURCES = \
 	examples/async_pi_estimate/sync.h \
 	nan.h \
 	test/cpp/asyncworker.cpp \
+        test/cpp/asyncprogressworker.cpp \
         test/cpp/asyncworkererror.cpp \
         test/cpp/bufferworkerpersistent.cpp \
         test/cpp/gc.cpp \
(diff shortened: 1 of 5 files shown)


jonathanong:
i'm testing this and it's working pretty well. i don't know much about C++ and stuff but i got it to work with @rvagg's help and it was pretty straight forward. 

so... a merge please!!!

rvagg:
This is slated for a 1.5 (#216) but it does need some docs I believe.

@brett19 do you want to add some docs to the README for this?

@kkoopa what do we need for a 1.5?

brett19:
I am unfortunately extremely busy at the moment, I'll try to get it done next week.

kkoopa:
What we need and what we want are two possibly separate things. Fixes and such for broken code or new version incompatibilities are what is strictly needed. That includes #223 and a new thing: `v8::V8::ContextDisposedNotification` has moved to the current isolate. Additionally, what's already in #216 should also be in the new version.

That should be it. Remaining improvements are probably better off for later versions. Oh yeah, we should test against io.js too.

jonathanong:
:+1:

PR #218, opened 2014-11-08
WeakPersistent Redesign

agnat:
This is a follow-up on #210.

@kkoopa mentioned in a [comment](https://github.com/rvagg/nan/pull/210#issuecomment-61372638) that the weak persistent callback stuff could use some love:

> If you're still looking for a challenge, the weak callback stuff is a bit uglier than might be necessary, should it not be possible to replace the dispatcher with a template function around a functor (would not really want a function pointer, as that is no improvement on the current situation), with the same syntax as currently used?
>
> Also, NAN_WEAK_CALLBACK does not fully work as NAN_METHOD and friends, you cannot use it to do a forward declaration and there is some fudging with class members due to the built-in static keyword which is needed sometimes, but not always, and cannot be set from user code due to it being a template declaration, meaning that static NAN_WEAK_CALLBACK(foo) puts the static keyword in the wrong place. This is something that still has not been fully resolved.

Let's discuss it.

I'm still in the process of wrapping my head around the current stuff but one thing seems obvious: It is not very flexible. Compared to `NanNew<T>(...)` it is totally rigid. I think the right thing to do is to get rid of `NAN_WEAK_CALLBACK` all together. Let users define the callbacks more freely (free function, static member function, functor, member function, what ever) and then deal with it in `NanMakeWeakPersistent(...)`. I'd also like to do away with that ugly argument.

Pseudo:

````
void f0(v8::Handle<v8::Value> trash) {}
void f1(v8::Handle<v8::Object> trash, int * param) {}
struct foo {
  static void bar(v8::Handle<v8::String> trash, foo * param) {}
};

NanMakeWeakPersistent(obj, f0); 
NanMakeWeakPersistent(obj, f1, new int());
NanMakeWeakPersistent(obj, foo::bar, new foo());
````

I might be a bit optimistic but I _think_ this is doable. What do you guys think?


kkoopa:
So, everybody, should we remove the not fully proper `NAN_WEAK_CALLBACK` macro and have users declare their weak callbacks like `void someCallback(NanWeakCallbackData<v8::Object, int> const& data);`?

While I dislike seeing `NAN_WEAK_CALLBACK` disappear from an esthetical point of view (it goes well with `NAN_METHOD` et al.), it has a different structure and simply cannot be made to work exactly the same way as `NAN_METHOD` et al., so I guess I give it a +1.

bnoordhuis:
We should have an abbreviation for "no opinion, do what is best".  Can I coin NODWIB?

kkoopa:
Merged as a65ee94238dbb3d3291c3be8bbcd8f60dfab6418

PR #223, opened 2014-12-14
Make compatible with io.js

### LINKED ISSUE #222: ExternalAsciiStringResource is gone in upstream V8

bnoordhuis:

Use SFINAE to make nan compatible with a range of V8 versions.

Fixes #222.

R=@kkoopa

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -500,10 +500,17 @@ NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Local<T> val) {
     return v8::String::NewExternal(v8::Isolate::GetCurrent(), resource);
   }
 
+#if NODE_MODULE_VERSION >= 42  // io.js v1.0.0
+  NAN_INLINE v8::Local<v8::String> NanNew(
+      v8::String::ExternalOneByteStringResource *resource) {
+    return v8::String::NewExternal(v8::Isolate::GetCurrent(), resource);
+  }
+#else
   NAN_INLINE v8::Local<v8::String> NanNew(
       v8::String::ExternalAsciiStringResource *resource) {
     return v8::String::NewExternal(v8::Isolate::GetCurrent(), resource);
   }
+#endif
 
 # define NanScope() v8::HandleScope scope(v8::Isolate::GetCurrent())
 # define NanEscapableScope()                                                   \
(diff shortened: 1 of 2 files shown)


bnoordhuis:
@kkoopa @rvagg PTAL, it's #ifdef'ing on the node version now.  The first commit removes the overloads for String::ExternalAsciiStringResource (the commit log explains why).  The second commit brings them back in case you think it's best to keep them.

rvagg:
@kkoopa are you around to cut a 1.5 within the next couple of days? io.js is going live and it would be good to have our bases covered with NAN. If you're not, then I can probably take care of it.

kkoopa:
Sure, I should be around, unless something unexpected happens.

kkoopa:
Has `NODE_MODULE_VERSION` been updated to 13? If so, is it not better to branch on that? I was notified that v8 has moved around some things as well: At some point, `ContextDisposedNotification` has moved from `v8::V8` to the current Isolate.

bnoordhuis:
> Has NODE_MODULE_VERSION been updated to 13?

No, currently just the NODE_VERSION_* macros.  I'll file a PR to bump NODE_MODULE_VERSION to something largish so it won't conflict with joyent/node.

> Has anything else been moved

A number of v8::V8 methods have been moved to v8::Isolate.  I don't know the exact list off the top of my head but it's methods like SetFatalErrorHandler() and such that were already isolate-ified under the hood, just not visibly so in the API.

Another change is that that v8::Value::To*() methods optionally (for now) take an Isolate argument.  And of course there's a slew of new API methods for dealing with generators, promises, symbols, sets, maps, etc.

A minor API addition that is really nice is that v8::Object and everything that descends from it now has a GetIsolate() method.

bnoordhuis:
> I'll file a PR to bump NODE_MODULE_VERSION

Done and landed in https://github.com/iojs/io.js/pull/312.

bnoordhuis:
Okay, decision time.  Do we:

1. Want to keep the ExternalAsciiStringResource stuff around, with the caveat that people will have to update their code some day anyway (only a temporary respite), and

2. Should this PR use NODE_MODULE_VERSION or NODE_VERSION_AT_LEAST?  I have a slight but mostly irrational preference for the latter.

rvagg:
one pro for using `NODE_MODULE_VERSION` is that it sets a better example for addon authors who are browsing through nan.h and learning from it, you shouldn't have to know what version ABI compat chaged, just that it changed.

bnoordhuis:
Okay.  A counterpoint is that version numbers are arguably more readable.  But like I said, it's mostly irrational, I'll update the PR.  What about the first point?

rvagg:
re `ExternalAsciiStringResource`, leave it for now I think so this isn't technically a "breaking" release

bnoordhuis:
Updated, PTAL.

kkoopa:
External strings are quite exotic, so I'm willing to fudge around a bit with it, but if removed we should make it NAN 2.0.

What APIs use external string resource? Would it be possible to make a new class Foo implementing the functionality without changing a bunch of other functions that reference the class?

Main concern, as usual, is to have consistent code and interfaces at the end user level, so the devs don't have to do version sniffing.

bnoordhuis:
> What APIs use external string resource? Would it be possible to make a new class Foo implementing the functionality without changing a bunch of other functions that reference the class?

Overloads of String::NewExternal() and String::MakeExternal().

You can shim it but I don't think you need to bother, external strings aren't that popular.  You use them when you have really big strings that you don't want to copy into the VM but in node land everyone uses buffers (and now typed arrays) for that.

As a data point: the iojs code base has only one place where it uses external strings and there it's just an accident of history.

kkoopa:
I guess this could work as a temporary fix to keep things compiling so we can release NAN 1.5, but I want it gone in NAN 2.0, which should succeed 1.5. The only code in NAN using it is `_NanGetExternalParts`, which is part of the deprecated `NanRawString` chain and should be removed anyway.

However, can you rebase against the onepointfive branch?

bnoordhuis:
I probably need to file a new PR for that.  GH always gets terribly confused when you change branches halfway through a PR.

bnoordhuis:
Continues in #224.

PR #224, opened 2015-01-13
Make compatible with io.js

### LINKED ISSUE #222: ExternalAsciiStringResource is gone in upstream V8

bnoordhuis:

This change makes nan compile again with V8 3.29 and up.

Fixes #222.

R=@kkoopa @rvagg


kkoopa:
LGTM

bajtos:
Any ETA on when this fix will be released?

kkoopa:
4 hours

bajtos:
> 4 hours

Wow, that's awesome! :tada:

PR #228, opened 2015-01-14
Make compatible with io.js

### LINKED ISSUE #222: ExternalAsciiStringResource is gone in upstream V8

bnoordhuis:

This change makes nan compile again with V8 3.29 and up.

Fixes #222.

This is a back-port of commit b003843 from the 1.5 and master branch
and should make modules that depend on nan ~1.4.x work with io.js
without requiring action from their maintainers.

Apologies, I had to create a v1.4 branch (branched off the v1.4.1 tag) to base this PR against.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -500,8 +500,13 @@ NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Local<T> val) {
     return v8::String::NewExternal(v8::Isolate::GetCurrent(), resource);
   }
 
+#if NODE_MODULE_VERSION >= 42  // io.js v1.0.0
+  NAN_INLINE v8::Local<v8::String> NanNew(
+      v8::String::ExternalOneByteStringResource *resource) {
+#else
   NAN_INLINE v8::Local<v8::String> NanNew(
       v8::String::ExternalAsciiStringResource *resource) {
+#endif
     return v8::String::NewExternal(v8::Isolate::GetCurrent(), resource);
   }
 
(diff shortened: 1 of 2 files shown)


rvagg:
big :thumbsup: to me for this

btw @kkoopa if you merge and release this as a patch bump on 1.4.x, do the publish with `--tag backport` (name of tag not so relevant), otherwise, I _think_ that your latest publish will get the `@latest` tag in npm which is what people will get when they `npm install nan`. The alternative is to republish 1.5 again after a 1.4.x publish but that just gets messy.

kkoopa:
Sure, that's a very good idea. Did not think of it. However, I'll be sleeping and working for the next 16 hours, so it might take me a day or two. Maybe you can publish a patch release sooner?

rvagg:
sure, I'll take this one, thanks @kkoopa

rvagg:
gah! I messed up that publish because I used `--tag 1.4`, npm complained that it couldn't be semver but went ahead and published anyway so 1.4.2 was `@latest`, so I've done another patch for 1.4.3 and also 1.5.1 just to make it clean.

kkoopa:
For future reference, what would the correct way have been?

rvagg:
`npm publish --tag backport` probably would have been better; tag name is arbitrary at this stage though because all we're trying to achieve is for it not to get the tag `latest` because `npm install nan` is actually short for `npm install nan@latest` (effectively). We want `~1.4.0` to resolve to `1.4.3` but not for `npm install nan` to install `1.4.3` but the _actual_ latest code, `1.5.0`.

PR #234, opened 2015-01-15
Fix non-inline definition build error with clang++

bnoordhuis:
Fix the following build error when building with clang++:

    In file included from ../cpp/asyncprogressworker.cpp:13:
    In file included from ../../nan.h:63:
    ../../nan_new.h:207:43: error: inline declaration of 'NanNew' follows non-inline definition
    template <typename T> inline v8::Local<T> NanNew(v8::Handle<T> h);
                                              ^
    ../../nan_implementation_12_inl.h:218:1: note: previous definition is here
    NanNew(v8::Handle<T> h) {
    ^
    In file included from ../cpp/asyncprogressworker.cpp:13:
    In file included from ../../nan.h:63:
    ../../nan_new.h:208:43: error: inline declaration of 'NanNew' follows non-inline definition
    template <typename T> inline v8::Local<T> NanNew(v8::Persistent<T> const& p);
                                              ^
    ../../nan_implementation_12_inl.h:224:1: note: previous definition is here
    NanNew(v8::Persistent<T> const& p) {

g++ is more lenient but clang++ is not wrong to complain.

Fixes: https://github.com/rvagg/nan/issues/231

R=@kkoopa

### DIFF
--- a/nan_implementation_12_inl.h
+++ b/nan_implementation_12_inl.h
@@ -214,14 +214,12 @@ Factory<v8::UnboundScript>::New( v8::Local<v8::String> source
 //=== Presistents and Handles ==================================================
 
 template <typename T>
-v8::Local<T>
-NanNew(v8::Handle<T> h) {
+inline v8::Local<T> NanNew(v8::Handle<T> h) {
   return v8::Local<T>::New(v8::Isolate::GetCurrent(), h);
 }
 
 template <typename T>
-v8::Local<T>
-NanNew(v8::Persistent<T> const& p) {
+inline v8::Local<T> NanNew(v8::Persistent<T> const& p) {
   return v8::Local<T>::New(v8::Isolate::GetCurrent(), p);
 }
 


kkoopa:
Looks good at a first glance, tests pass. I'll still ask, is it correct to have 
the inline keyword there in the first place? There have been many discussions 
on where NAN_INLINE is superfluous and not (something about template functions 
implicitly being inline, but only when methods, not as pure functions, and 
only on the second monday of every month, except for leap years). I never 
figured out exactly what's what, and if it is according to the standard or 
implementations or if it is optional or not and whether potential compiler 
bugs exist in regards to this or not.

However, I seem to remember that you had something to say on the subject and I 
will defer to better judgment than mine. Will merge.

bnoordhuis:
@kkoopa Template functions need the inline keyword because of the ODR rule.  The one place where you can skip the inline keyword is in inline method definitions:

```c++
class C {
  inline void f();  // needs 'inline'
  void g() {}  // inline definition
};
```

PR #235, opened 2015-01-15
Fix non-inline definition build error with clang++

bnoordhuis:
Fix the following build error when building with clang++:

    In file included from ../cpp/asyncprogressworker.cpp:13:
    In file included from ../../nan.h:63:
    ../../nan_new.h:207:43: error: inline declaration of 'NanNew' follows non-inline definition
    template <typename T> inline v8::Local<T> NanNew(v8::Handle<T> h);
                                              ^
    ../../nan_implementation_pre_12_inl.h:210:1: note: previous definition is here
    NanNew(v8::Handle<T> h) {
    ^
    In file included from ../cpp/asyncprogressworker.cpp:13:
    In file included from ../../nan.h:63:
    ../../nan_new.h:208:43: error: inline declaration of 'NanNew' follows non-inline definition
    template <typename T> inline v8::Local<T> NanNew(v8::Persistent<T> const& p);
                                              ^
    ../../nan_implementation_pre_12_inl.h:216:1: note: previous definition is here
    NanNew(v8::Persistent<T> const& p) {
    ^
    2 errors generated.

g++ is more lenient but clang++ is not wrong to complain.

Fixes: https://github.com/rvagg/nan/issues/231 (for v0.10 this time.)

R=@kkoopa

Mea culpa, I could swear I tested with v0.10 as well.

### DIFF
--- a/nan_implementation_pre_12_inl.h
+++ b/nan_implementation_pre_12_inl.h
@@ -206,14 +206,12 @@ Factory<v8::StringObject>::New(v8::Handle<v8::String> value) {
 //=== Presistents and Handles ==================================================
 
 template <typename T>
-v8::Local<T>
-NanNew(v8::Handle<T> h) {
+inline v8::Local<T> NanNew(v8::Handle<T> h) {
   return v8::Local<T>::New(h);
 }
 
 template <typename T>
-v8::Local<T>
-NanNew(v8::Persistent<T> const& p) {
+inline v8::Local<T> NanNew(v8::Persistent<T> const& p) {
   return v8::Local<T>::New(p);
 }
 


springmeyer:
:+1: works! https://travis-ci.org/mapnik/node-mapnik/builds/47160299

kkoopa:
I could swear it was tested with that version too... It passed both Travis and AppVeyor.

kkoopa:
Ah, but neither used clang of course. Will merge.

springmeyer:
> Ah, but neither used clang of course.

Not sure if it is related but I'm actually using and upgraded clang-3.5 from https://github.com/mapnik/node-mapnik/blob/master/scripts/build_against_sdk.sh#L30

> Will merge.

Thanks!

PR #237, opened 2015-01-16
Readded missing String constructors.

### LINKED ISSUE #227: NanNew(const char*&, std::size_t&)
i made a module with the async progress worker PR. when trying to upgrade, i get the following compilation error:

```
In member function ‘virtual void TrackWorker::HandleProgressCallback(const char*, std::size_t)’:
error: no matching function for call to ‘NanNew(const char*&, std::size_t&)’
```

My code looks like this:

```cpp
    void HandleProgressCallback(const char *data, size_t size) {
      NanScope();

      Handle<Value> argv[1];

      argv[0] = NanNew(data, size);

      progress->Call(1, argv);
    }
```

i'm assuming this is caused by the `NanNew` rewrite. what's the correct way to do this?

kkoopa:
Added `String` constructors that were inadvertently omitted in the `NanNew` rewrite.
Fixes #227.

R=@agnat

### DIFF
--- a/nan_implementation_12_inl.h
+++ b/nan_implementation_12_inl.h
@@ -150,6 +150,11 @@ Factory<v8::Signature>::New( Factory<v8::Signature>::FTH receiver
 //=== String ===================================================================
 
 Factory<v8::String>::return_t
+Factory<v8::String>::New() {
+  return v8::String::Empty(v8::Isolate::GetCurrent());
+}
+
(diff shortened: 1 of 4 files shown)


rvagg:
:+1: let's get a patch release out for 1.5 soon, preferably with #239 too I think

agnat:
LGTM.

Issue #225, opened 2015-01-14
Changes in v8 between joyent/0.11.14 and io.js/1.0


kkoopa:
The following utility functions have moved from `v8::V8` to `v8::Isolate`

V8 -> Isolate
-----------------------------------------------------
*** moved, but deprecated remains
%%% deprecated and moved before v8 @ 0.11.14, ignore
### removed
-----------------------------------------------------
AddMessageListener ***
RemoveMessageListeners ***
SetCaptureStackTraceForUncaughtExceptions ***

SetCounterFunction
SetCreateHistogramFunction
SetAddHistogramSampleFunction

SetFailedAccessCheckCallbackFunction ***

AddGCPrologueCallback ***
RemoveGCPrologueCallback ***
AddGCEpilogueCallback ***
RemoveGCEpilogueCallback ***

AddMemoryAllocationCallback ***
RemoveMemoryAllocationCallback ***

RunMicrotasks %%%
EnqueueMicrotask %%%
SetAutorunMicrotasks %%%

SetFunctionEntryHook ###
SetJitCodeEventHandler

TerminateExecution ***
IsExecutionTerminating ***
CancelTerminateExecution ***

VisitExternalResources ***
VisitHandlesWithClassIds ***
VisitHandlesForPartialDependence ***

IdleNotification
LowMemoryNotification
ContextDisposedNotification

rvagg:
this looks like work for 1.6

kkoopa:
Only those without special markings are strictly necessary. They are quite easy to address. I'll see what can be done. Those that have deprecated holdouts lying around can still be used for the time being.

ReklatsMasters:
`String::New` was removed. Empty constructor `Object::New()` was removed.
We need compability layer.

kkoopa:
`String::New` was removed a long time ago already, maybe in the version v8 that was in 0.11.13, or even before that. The Object constructor has definitely not been removed.

Issue #226, opened 2015-01-14
atom-shell causing problems


kkoopa:
https://github.com/brianmcd/contextify/pull/145

I hope this is just some temporary fork and that they'll move to io.js instead of running yet another fork. It will be impossible to support three things each claiming to be Node without any reliable way of telling them apart. Why can't v8 announce which version it is?

bnoordhuis:
> Why can't v8 announce which version it is?

I can try sending a patch.  I remember looking into it one time and not quite being able to figure out how their homegrown release script updated version numbers.

kkoopa:
That would be good. However, Node forkers need to backport that patch too, because having it only in the very latest version of v8 won't do much good in the short time.

bnoordhuis:
You can detect that the version macro is not there, that at least tells you something.  But yes, it won't do a great deal of good in the short term.

kkoopa:
What if every fork could add a define to node-version.h

```c++
#define MY_UNIQUE_DEFINITION 1
```
or use a standard name, but unique value,
```c++
#define NODE_FORK_IDENTIFIER 23
```

bnoordhuis:
It could work but it's not ideal because we're really only interested in the V8 version.  Also, it may be an unreliable indicator when linked against a shared libv8.

waTeim:
Here's some workaround possibilities; tl;dr PREPARSER_H in in 3.25 and 3.26 but not in 3.29, lots of changes in v8 3.14 to 3.25 (V8_ARRAY_BUFFER_VIEW_INTERNAL_FIELD_COUNT for instance)  I have not downloaded iojs yet.  

node 11.14 and 11.13 (v8 3.26.33, 3.25.30)

    bizarro% grep \\#define v8*
    v8-debug.h:#define V8_V8_DEBUG_H_
    v8-platform.h:#define V8_V8_PLATFORM_H_
    v8-preparser.h:#define PREPARSER_H
    v8-preparser.h:#define V8EXPORT __declspec(dllexport)
    v8-preparser.h:#define V8EXPORT __declspec(dllimport)
    v8-preparser.h:#define V8EXPORT
    v8-preparser.h:#define V8EXPORT __attribute__ ((visibility("default")))
    v8-preparser.h:#define V8EXPORT
    v8-profiler.h:#define V8_V8_PROFILER_H_
    v8-testing.h:#define V8_V8_TEST_H_
    v8-util.h:#define V8_UTIL_H_
    v8.h:#define V8_H_
[...]

PR #238, opened 2015-01-16
NAN 1.6

kkoopa:
* Added new functions to deal with functions moved from V8 to Isolate. See #225.
* Deprecated `NanNewContextHandle` in favor of `NanNew<Context>`.
* Add support for node::Encode/DecodeBytes/DecodeWrite

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # NAN ChangeLog
 
-**Version 1.5.2: current Node unstable: 0.11.15, Node stable: 0.10.35, io.js: 1.0.3**
+**Version 1.6.0: current Node unstable: 0.11.15, Node stable: 0.10.35, io.js: 1.0.3**
+
+### 1.6.0 Jan 23 2015
+ - Deprecated `NanNewContextHandle` in favor of `NanNew<Context>` 49259af
+ - Support utility functions moved in newer v8 versions (Node 0.11.15, io.js 1.0) a0aa179
+ - Added `NanEncode`, `NanDecodeBytes` and `NanDecodeWrite` 75e6fb9
 
 ### 1.5.2 Jan 23 2015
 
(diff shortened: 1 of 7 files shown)


kkoopa:
This will not work with node webkit, as they have updated v8, but not indicated it in any detectable way as far as I can see. Too bad.

rvagg:
@rogerwang is there someone in particular we should pull in for this? nw.js could lose solid native addon support if we can't find a resolution to this.

kkoopa:
#226 is related for slightly different reasons

rogerwang:
I'll help to fix this. What do I need to do in nw.js? defining NODE_MODULE_VERSION to some value?

rvagg:
@kkoopa do we a good have a recommendation here?

@rogerwang if your ABI is the same as some io.js or node release then match the `NODE_MODULE_VERSION` to that, otherwise make it significantly different. io.js is at 42, atom has jumped to [100](https://github.com/atom/node/commit/83f1b415e95e4d47e70c1cd064fdf153fc65fc47). Another alternative is to introduce a new define we can pick up on that talks to your ABI version.

rogerwang:
It should be the same as io.js. I'll update the io.js in nw.js, so it will be 42. Currently it's some old version (1.0.0).

kkoopa:
Ah, that would of course explain the mismatch. io.js only bumped `NODE_MODULE_VERSION` at the last minute before release. Everything should be alright after updating.

Mithgol:
NW.js v0.12.0-alpha3 was [released](https://groups.google.com/forum/#!topic/nwjs-general/heM9DRTLx5U) ≈8 hours ago.

Its inner Chromium is updated to 41 beta (41.0.2272.32) and io.js is updated to v1.2.0.

Does that mean nan support?

kkoopa:
I guess so.

PR #239, opened 2015-01-17
typecast to uv_async_cb to fix node-webkit compile

springmeyer:
- Fixes mapbox/node-sqlite3#387
 - Unblocks io.js release of node-sqlite3 - mapbox/node-sqlite3#385

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1467,7 +1467,7 @@ class NanCallback {
     uv_async_init(
         uv_default_loop()
       , async
-      , AsyncProgress_
+      , reinterpret_cast<uv_async_cb>(AsyncProgress_)
     );
     async->data = this;
 


rvagg:
oh, not so great that we let this slip in, we obviously don't have test cases for this. I'm :+1: on this change and pushing a patch-release for 1.5. While we wait for additional input @springmeyer can you have a quick look at the tests to see why this isn't flagged anywhere?

bnoordhuis:
I suspect a bug in the NAUV_UVVERSION macro because the NAUV_WORK_CB macro is supposed to take care of the differences in uv_async_cb prototypes.  @brett19?  I believe you originally wrote that?

springmeyer:
Just confirmed that all of the crazy node-sqlite3 build matrix works with this patch: https://travis-ci.org/mapbox/node-sqlite3/builds/47314165

>  @springmeyer can you have a quick look at the tests to see why this isn't flagged anywhere?

Okay, taking a look. As a byproduct see #240 which should help those new to running the tests like me.

springmeyer:
I've come up short on a reduced testcase to replicate/prevent regression. But I've found that it is easy to replicate with the nan directory just by doing:

```
npm install nw-gyp
./node_modules/.bin/nw-gyp rebuild --directory test --target=0.11.5
```

Is it reasonable for the nan .travis.yml to test against node-webkit to be able to catch more subtle breakages like this or do you need something more reduced?

rvagg:
I don't mind using nw-gyp as a devdep if it's going to catch more, I just don't know anything about nw-gyp at all

springmeyer:
Okay, new pull request testing node-webkit at https://github.com/rvagg/nan/pull/241 (idea is we can watch the travis build there to make sure it fails).

kkoopa:
The declaration of AsyncProgress_ comes after this, maybe that's why it's whining? The symbol is undefined at that point.

PR #240, opened 2015-01-17
Add details to readme on how to run tests

springmeyer:
Hopefully I got this right.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,4 +56,4 @@ forcetest:
 	npm test
 
 $(ADDONS): nan.h nan_new.h nan_implementation_pre_12_inl.h nan_implementation_12_inl.h test/binding.gyp $(SOURCES)
-	cd test/ && node-gyp rebuild
+	cd test/ && ../node_modules/.bin/node-gyp rebuild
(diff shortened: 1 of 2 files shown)


kkoopa:
make test works wonders if gnu make is available

rvagg:
@springmeyer I'd put `make test` as an alternative way to run them, I think you can keep what you have there as well

springmeyer:
haha, totally did not see the Makefile. Running `make test` though does not find node-gyp for me in the local `node_modules/.bin`:

```
$ make test
cd test/ && node-gyp rebuild
/bin/sh: node-gyp: command not found
make: *** [test/build/Release/symbols.node] Error 127
```

rvagg:
yeah, it's designed for a certain kind of workflow which assumes a global node-gyp and that may not be ideal for casual contributors .. FYI in .dntrc I do this which is different again:

```
  npm install && \
  node_modules/.bin/node-gyp --nodedir /usr/src/node/ rebuild --directory test && \
  node_modules/.bin/tap --gc test/js/*-test.js
```

kkoopa:
Won't this break make test if I don't have a locally installed node-gyp?

PR #247, opened 2015-01-18
Make explicit downcasts of String lengths.

kkoopa:
msvc correctly warns about this

### DIFF
--- a/nan_implementation_pre_12_inl.h
+++ b/nan_implementation_pre_12_inl.h
@@ -158,19 +158,20 @@ Factory<v8::String>::New(const char * value, int length) {
 
 Factory<v8::String>::return_t
 Factory<v8::String>::New(std::string const& value) {
-  return v8::String::New( value.data(), value.size());
+  assert(value.size() <= INT_MAX && "string too long");
+  return v8::String::New( value.data(), static_cast<int>(value.size()));
 }
 
 inline
 void
 widenString(std::vector<uint16_t> *ws, const uint8_t *s, int l = -1) {
+  size_t len = static_cast<size_t>(l);
   if (l < 0) {
-    size_t foundLength = strlen(reinterpret_cast<const char*>(s));
-    assert(foundLength <= INT_MAX && "string to long");
-    l = foundLength;
+    len = strlen(reinterpret_cast<const char*>(s));
   }
-  ws->resize(l);
-  std::copy(s, s + l, ws->begin());
+  assert(len <= INT_MAX && "string too long");
+  ws->resize(len);
+  std::copy(s, s + len, ws->begin());
 }
 
 Factory<v8::String>::return_t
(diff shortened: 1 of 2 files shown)


nan_implementation_pre_12_inl.h
```diff
@@ -185,7 +186,8 @@ Factory<v8::String>::New(const uint8_t * value, int length) {
   if (wideString.size() == 0) {
     return v8::String::Empty();
   } else {
-    return v8::String::New(&wideString.front(), wideString.size());
+    return v8::String::New(&wideString.front()
+         , static_cast<int>(wideString.size()));
```
bnoordhuis: No assert here?

kkoopa: I think there's an assert in the widenString function called above, so it should not be necessaey to assert again, as the length ought to have been checked already.

kkoopa: But I see now that it is only when no length is specified to `widenString`, so yes, should add an assert here.

kkoopa: Moved the assert in widenString outside the branch instead. Now `wideString.size()` is guaranteed to be `<= INT_MAX` at this point.

nan_implementation_pre_12_inl.h
```diff
@@ -166,9 +167,9 @@ void
 widenString(std::vector<uint16_t> *ws, const uint8_t *s, int l = -1) {
   if (l < 0) {
     size_t foundLength = strlen(reinterpret_cast<const char*>(s));
-    assert(foundLength <= INT_MAX && "string to long");
     l = foundLength;
```
bnoordhuis: There's an implicit downcast here now.

kkoopa: crap, you're right

kkoopa:
Hopefully I got it right this time.

nan.h
```diff
@@ -29,6 +29,9 @@
 #include <climits>
 #if defined(_MSC_VER)
 # pragma warning( disable : 4530 )
+# if _MSC_VER < 1700  // < VS 2012, don't know when the warning stops appearing
+#  pragma warning( disable : 4344 )
```
bnoordhuis: I agree that C4344 is an annoying warning but I don't think disabling it file-wide is a good thing.

nan_implementation_pre_12_inl.h
```diff
@@ -158,18 +158,19 @@ Factory<v8::String>::New(const char * value, int length) {
...
   if (l < 0) {
-    size_t foundLength = strlen(reinterpret_cast<const char*>(s));
-    assert(foundLength <= INT_MAX && "string to long");
-    l = foundLength;
+    len = strlen(reinterpret_cast<const char*>(s));
   }
-  ws->resize(l);
+  assert(len <= INT_MAX && "string too long");
+  ws->resize(len);
   std::copy(s, s + l, ws->begin());
```
bnoordhuis: s + len?

nan.h
kkoopa: Perhaps not. As it cannot be limited, I'll just remove it.

nan_implementation_pre_12_inl.h
kkoopa: facepalm

kkoopa:
Fourth? time's the charm.

PR #243, opened 2015-01-18
Fix #242: Disambiguate function pointer arguments

agnat:
This PR fixes #242. Turns out I initially proposed exactly the wrong fix. Using a function template specialization just doesn't work:

1. You can't come up with default arguments for a full specialization. (Kind of obvious)
2. Less obvious: Full specializations are *not* part of the overload set. See http://www.gotw.ca/publications/mill17.htm

So, we need to do the other thing. Instead of using a specialization, like:

````c++
template <>
typename NanIntern::Factory<v8::FunctionTemplate>::return_t
NanNew<v8::FunctionTemplate>(NanFunctionCallback callback) {
    return NanIntern::Factory<v8::FunctionTemplate>::New(callback);
}
````

we're going to use an overload:

Issue #208, opened 2014-10-27
NAN 2 Discussion


rvagg:
Just a place for putting ideas of things that we might want to break. I don't believe we should invest too much in this and make it a big deal, the more we put in here the longer it'll take and more painful it will be. I think we should just embrace a more rapid release model and not treat a 2.0.0 as a particularly big deal. Regardless, there still needs to be breakage to fix and improve things so put them in here!

Something I'd like us to think about for some point in the future (probably not v2) is making Isolates first-class in NAN so that we're not dealing with a lowest-common-denominator API. Perhaps we introduce a `NanIsolate` that's a placeholder for pre-0.11 but a real Isolate after that. Then we could require it everywhere V8 needs now but just ignore it in older versions. The point of this would be to make way for code that wants to use Isolates in a more sophisticated way, even though it'd be unlikely you could do that with code that supports pre-0.11 code you'd still be able to use NAN for the additional sugar and helpers and we'd set NAN up better for supporting future versions of V8.

kkoopa:
Closing in favor of #334.

PR #248, opened 2015-01-19
Add function and object template factories

agnat:
This popped up while working on #243.

Ready for review.


kkoopa:
What about the 4-arg `FunctionTemplate` with the final length argument?

agnat:
Right... I probably was looking at an old v8 header and must have missed it. Will add it later today...

kkoopa:
If it wasn't always available we can't really add it. Lowest common denominator and all. I just assumed it was.

agnat:
Hm, I thought about that when implementing `Factory<Function>`. (Direct construction of `Function` is only available in newer v8.) I think just offering the LCD is not going to cut it in the long run. A project might target node >=0.10 or io.js. It still wants NaNs future proofing even (or specially) for features that are not part of the LCD (yet). I think a good approach is to wrap new features early and if possible provide graceful degradation on older versions. We already do that for (bound) scripts. If we can't reasonably emulate a feature so be it. It just needs to be documented.

Just my opinion, though.

kkoopa:
Yes, where there is not too much work, backporting and emulating is all very good. Just want the external interface to be identical.

I think that something correctly written in proper NAN should work on every supported version. It should not be that parts only work on certain versions.

For that we should have new major versions of NAN that drop support for older nodes and break the API.

i.e. NAN 1 is for 0.8+, NAN 2 for 0.10+, NAN 3 for 0.12+. But nothing is of course set in stone and I don't know how v8 will evolve in the future, this needs to be adapted to reality. io.js claims to update v8 more often, so maybe we have to follow that. Who knows...

See e.g. #208

kkoopa:
Can't write a test for `NanNew<Function>` without doing it conditionally on version. Perhaps a conditional define would be the solution? Build with -DNAN_BACKWARDS_COMPATIBLE 1337 or something to get it enabled.

agnat:
> On Jan 19, 2015, at 15:51, Benjamin Byholm <notifications@github.com> wrote:
> 
> Can't write a test for NanNew<Function> without doing it conditionally on version.
> 

Just backport the constructor. It just creates a function template and returns the resulting function. I already checked. I can do it tonight...

The length argument requires more doing, though.
> Perhaps a conditional define would be the solution? Build with -DNAN_BACKWARDS_COMPATIBLE 1337 or something to get it enabled.
> 

Opt-in seems like the right thing to do (in general). Either with defines or by including additional/different headers.

kkoopa:
If backporting works, that would be great. From what I have seen, it is usually impossible, as anything that uses v8 "internal" stuff needs to be in the correct classes, and they are all non-extendable. But, if it just does a `FunctionTemplate` and returns the function, acting like a convenience method, that should indeed work!

Length argument might be impossible for above mentioned reasons. I'll take a quick glance at the code and see.

I like opt-in too. As explicit as possible, that should guard against spurious bug reports for "broken" functionality. Still requires good documentation (better organized than what we currently have).

kkoopa:
Length can be done. It just ultimately just does `obj->set_length(length);`, which can be done externally through `FunctionTemplate::SetLength`. So, just construct a function template and so `SetLength` on it before returning.

agnat:
Unfortunately `FunctionTemplate::SetLength()` is a new v8 feature, too. Guess, we have to skip the length argument for now... :-/

kkoopa:
Crap. Maybe it's possible to use
Object::ForceSet(NanNew("length"), NanNew(length), v8::ReadOnly);
or similar?

@bnoordhuis: Any suggestions?

kkoopa:
I guess it is new because it is part of the ES6 standard.
https://people.mozilla.org/~jorendorff/es6-draft.html#sec-functioninitialize

Let status be DefinePropertyOrThrow(F, "length", PropertyDescriptor{[[Value]]: 
len, [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true}).

kkoopa:
The actual v8 implementation seems to define the property through an accessor pair operating on an internal field of sorts at `kLengthOffset`.

kkoopa:
I want to get this merged so I can push the new versions.

Waiting for backport of `Function::New`.

Unsure about whether `FunctionTemplate::New` with `length` argument is doable, can hold off until later.

agnat:
Ack. Will finish it tonight.

agnat:
I found no (good) way to back port `Function.length`. `#define NAN_ENABLE_FUNCTION_LENGTH` enables the additional argument. Not sure I like it, though. Maybe we should keep it under the wraps for a bit and think about it some more.

Ready for review.

kkoopa:
I think the extra length argument is quite messy, but I can't see a better way of doing it either. I agree with keeping it under the wraps until some future point, preferrably not having it clutter up the code for now. If absolutely necessary at some future point, we can add it in -- provided that no fully compatible solution shows up.

agnat:
I removed the length argument for now. I'll add another PR to discuss our options.

kkoopa:
Excellent. I'll merge this, do some testing and push out the new releases then.

test/cpp/nannew.cpp
```diff
@@ -342,9 +376,13 @@ NAN_METHOD(testRegression242) {
...
   NanTap t(args[0]);
 
-  // This line needs to *compile*. Not much to test at runtime.
+  // These lines must *compile*. Not much to test at runtime.
   Local<FunctionTemplate> ft = NanNew<FunctionTemplate>(overloaded);
   (void)ft;  // not unused
+#if (NODE_MODULE_VERSION >= 12)
+  Local<Function> f = NanNew<Function>(overloaded);
+  (void)f;  // not unused
+#endif
```
kkoopa: Is this conditioning necessary any more?

agnat:
Nope. Fixed.

test/cpp/nannew.cpp
agnat: Nope. Fixed. (Sorry, about the spam).

PR #251, opened 2015-01-19
Add support for node::Encode/DecodeBytes/DecodeWrite.

### LINKED ISSUE #250: Add support for node::Encode/DecodeBytes/DecodeWrite
Could you add support for those handy utils?

```cpp
#if NODE_VERSION_AT_LEAST(0,11,12)
#define NanEncode(buf, buflen, encoding) \
  node::Encode(v8::Isolate::GetCurrent(), buf, buflen, encoding)
#define NanDecodeBytes(val, encoding) \
  node::DecodeBytes(v8::Isolate::GetCurrent(), val, encoding)
#define NanDecodeWrite(buf, buflen, val, encoding) \
  node::DecodeWrite(v8::Isolate::GetCurrent(), buf, buflen, val, encoding)
#else
#define NanEncode(buf, buflen, encoding) \
  node::Encode(buf, buflen, encoding)
#define NanDecodeBytes(val, encoding) \
  node::DecodeBytes(val, encoding)
#define NanDecodeWrite(buf, buflen, val, encoding) \
  node::DecodeWrite(buf, buflen, val, encoding)
#endif
```

kkoopa:
Fixes #250.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -155,6 +155,62 @@ NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Local<T> val) {
   return val;
 }
 
+#if NODE_MODULE_VERSION >= 42 // io.js 1.0
+  NAN_INLINE
+  void NanSetCounterFunction(v8::CounterLookupCallback cb) {
+    v8::Isolate::GetCurrent()->SetCounterFunction(cb);
+  }
+
+  NAN_INLINE
+  void NanSetCreateHistogramFunction(v8::CreateHistogramCallback cb) {
+    v8::Isolate::GetCurrent()->SetCreateHistogramFunction(cb);
+  }
+
+  NAN_INLINE
+  void NanSetAddHistogramSampleFunction(v8::AddHistogramSampleCallback cb) {
+    v8::Isolate::GetCurrent()->SetAddHistogramSampleFunction(cb);
+  }
+
+  NAN_INLINE bool NanIdleNotification(int idle_time_in_ms) {
+    return v8::Isolate::GetCurrent()->IdleNotification(idle_time_in_ms);
+  }
+
+  NAN_INLINE void NanLowMemoryNotification() {
+    v8::Isolate::GetCurrent()->LowMemoryNotification();
+  }
+
+  NAN_INLINE void NanContextDisposedNotification() {
+    v8::Isolate::GetCurrent()->ContextDisposedNotification();
+  }
+#else
+  NAN_INLINE
+  void NanSetCounterFunction(v8::CounterLookupCallback cb) {
+    v8::V8::SetCounterFunction(cb);
+  }
+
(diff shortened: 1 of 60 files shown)


nan.h
```diff
@@ -1816,6 +1816,45 @@ namespace Nan {
   enum Encoding {ASCII, UTF8, BASE64, UCS2, BINARY, HEX, BUFFER};
 }
 
+NAN_INLINE v8::Local<v8::Value> NanEncode(const void *buf, size_t len, enum Nan::Encoding encoding = Nan::BINARY) {
```
bnoordhuis: Maybe line-wrap this?

nan.h
```diff
@@ -1816,6 +1816,45 @@ namespace Nan {
...
+#if (NODE_MODULE_VERSION > 0x000B)
+return node::Encode(v8::Isolate::GetCurrent(), buf, len, static_cast<node::encoding>(encoding));
+#else
+# if  (NODE_MODULE_VERSION < 0x000B)
+  if (encoding == Nan::BUFFER) {
+    return node::Encode(buf, len, node::BINARY);
```
bnoordhuis: This should be `return node::Buffer::New(buf, len);`.

nan.h
```diff
@@ -1816,6 +1816,45 @@ namespace Nan {
...
+#endif
+}
+
+NAN_INLINE ssize_t NanDecodeBytes(v8::Handle<v8::Value> val, enum Nan::Encoding encoding = Nan::BINARY) {
+#if (NODE_MODULE_VERSION > 0x000B)
+  return node::DecodeBytes(v8::Isolate::GetCurrent(), val, static_cast<node::encoding>(encoding));
+#else
+# if (NODE_MODULE_VERSION < 0x000B)
+  if (encoding == Nan::BUFFER) {
+    return node::DecodeBytes(val, node::BINARY);
```
bnoordhuis: `DecodeBytes(buf, BUFFER)` calls ToString() on the buffer, which interprets (for better or worse) the contents of the buffer as UTF-8.  It's arguably a bug but it's better to be compatible.

bnoordhuis:
The DecodeBytes() and DecodeWrite() calls that transpose BUFFER to BINARY work but perhaps it's clearer to call the corresponding Buffer methods directly.

kkoopa:
I did that first, but changed to this after I noticed the code paths would become identical as it felt cleaner and does not perform the same sanity checks twice. Otherwise I need to replicate a bunch of error checking (val is not, array, because 'raw' encoding is no longer supported) and other edge cases, which makes it grow more than one might first think.

nan.h
kkoopa: Ah, yes. I forgot. Mostly wanted to see that Travis did not choke.

kkoopa:
Should be fixed now.

PR #253, opened 2015-01-19
Enabled more compiler warnings.

kkoopa:
Wundef and Wshadow produce too much noise from v8 and node. Maybe it is possible to selectively ignore them from these sources? This probably needs conditioning on other compilers (msvc, clang).

RFC

### DIFF
--- a/test/binding.gyp
+++ b/test/binding.gyp
@@ -1,155 +1,95 @@
-{"targets": [
+{
+  "target_defaults":
+    {
+        "cflags" : ["-Wall", "-Wextra"],
+        "include_dirs": ["<!(node -e \"require('..')\")"]
+    },
+  "targets": [
     {
         "target_name" : "symbols"
       , "sources"     : [ "cpp/symbols.cpp" ]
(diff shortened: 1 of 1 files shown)


bnoordhuis:
LGTM

kkoopa:
I don't think msvc accepts Wextra, will try to add a condition for that. Perhaps on os==windows.

bnoordhuis:
'cflags' is only used for Makefile and ninja based builds.  For MSVC, you need to tweak 'msvs_settings'.

mathiask88:
In my project I got this
```
'conditions': [
				['OS=="linux"', {
					'cflags_cc': ['-O3', '-fPIC', '-pedantic', '-fexceptions']
					}
				],
				['OS=="win"', {
					'msvs_settings': {
						'VCCLCompilerTool': {
							'AdditionalOptions': [ '/EHsc' ]
						}
					}
				}]
			]
```

kkoopa:
Ah, excellent.

Issue #212, opened 2014-11-02
NanNew<Boolean>(myglib_bool) fails with nan 1.4


kapouer:
Using glib's gboolean as input for NanNew<Boolean>() fails with

```
../node_modules/nan/nan.h:1080:22: error: could not convert ‘v8::Boolean::New((arg != 0))’ from ‘v8::Handle<v8::Boolean>’ to ‘v8::Local<v8::Boolean>’
```

Easily fixable by casting to bool.
It doesn't look like it's something you did on purpose, hence the report. Is it ?

PR #254, opened 2015-01-19
windows: Circumvent warning C4800

kkoopa:
msvc does not like imlpicit boolean conversion

Is this the right way to fix it?
What if I use an unsigned int?

Would it be better to do this?
```c++
template<>
inline typename NanIntern::Factory<v8::Boolean>::return_t
NanNew<v8::Boolean, A0>(A0 arg0) {
  return NanIntern::Factory<v8::Boolean>::New(arg0 != 0);
}
```
Should the above then be complemented with this?
```c++
template<>
inline typename NanIntern::Factory<v8::Boolean>::return_t
NanNew<v8::Boolean, bool>(A0 arg0) {
  return NanIntern::Factory<v8::Boolean>::New(arg0);
}
```

R=@agnat

### DIFF
--- a/test/cpp/nannew.cpp
+++ b/test/cpp/nannew.cpp
@@ -323,14 +323,20 @@ NAN_METHOD(testRegression212) {
 
   typedef int  gint;
   typedef gint gboolean;
+#if defined(_MSC_VER)
+# pragma warning( disable : 4800 )
+#endif
   t.ok(_( assertType<Boolean>( NanNew<Boolean>(gboolean(23)))));
+#if defined(_MSC_VER)
+# pragma warning( default : 4800 )
+#endif
 
   return_NanUndefined();
 }
 
 /* Compile time regression test for https://github.com/rvagg/nan/issues/242
  * In the presence of overloaded functions NaN should be able to pick the one
- * matching NanFunctionCallback. 
+ * matching NanFunctionCallback.
  */
 void overloaded() {}
 NAN_METHOD(overloaded) {


mathiask88:
Yes this is the right solution. But I dont like this warning because the compiler optimizes the code the same way at compile time as far as I know..

agnat:
> Is this the right way to fix it?

I hope not! Let's see. This is from `testRegression212` (#212), right? If you recall that one, it was not about silently accepting anything as a bool, but it was about not bailing out if it happens to be an int. I'd say we're good on that and the warning is correct. Users of `crazy_bool` just have to roll their own conversion/cast/comparison.

In the test case we should _not_ convert the `gboolean`. It would defeat the purpose of the test. Instead we should disable the warning for `testRegression212`.

agnat:
We should try _really_ hard to avoid specializations (or overloads for that matter) on `NanNew()`. We know where that leads. AFAICT, it is the only way to handle the function pointers. But that should be it ... I hope ;).

kkoopa:
Yes, warning comes from those two tests, gboolean and raw int. Disabling the warning around them sounds fair. I now agree fully on this with both of you.

kkoopa:
This ought to do.

agnat:
LGTM.

mathiask88:
I think you missed a 4800 in news.cpp
```
// #212
NAN_METHOD(NewBoolean2) {
  NanScope();
  NanReturnValue(NanNew<v8::Boolean>(1));
}
```

kkoopa:
You are correct. Will fix.

PR #259, opened 2015-01-21
Support joyent/node 0.11.15

kkoopa:
Scheduled for 1.6.0.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -158,7 +158,8 @@ NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Local<T> val) {
   return val;
 }
 
-#if NODE_MODULE_VERSION >= 42 // io.js 1.0
+/* io.js 1.0  */
+#if NODE_MODULE_VERSION >= 42 || NODE_VERSION_AT_LEAST(0, 11, 15)
   NAN_INLINE
   void NanSetCounterFunction(v8::CounterLookupCallback cb) {
     v8::Isolate::GetCurrent()->SetCounterFunction(cb);


bnoordhuis:
LGTM but I would keep the comment.

kkoopa:
Yes, I readded the comment.

bnoordhuis:
@kkoopa You should back-port this to the v1.4 branch.

bnoordhuis:
Or maybe not... NanSetCounterFunction() is new in v1.5.

kkoopa:
Really? It's quite many new additions.

Yes, I think 1.6 is the right version.

1.5 had the `NanNew` rewrite, will push out one more patch release of that with bugfixes, as soon as #248 has landed.

As soon as that is done, I will push 1.6.0, which will introduce support for joyent/node 0.11.15, hopefully will work with atom-shell, and also have the new Encode/Decode functions.

kkoopa:
Actually, I think 1.4 or 1.5 should work on 0.11.15. The moved utility function wrappers were omitted from 1.5.0 due to time constraints. As long as you don't need them, either version is fine to use.

The wrappers will be in 1.6.0, which is still unreleased. Both joyent/0.11.15 and iojs/1.0.0 require the wrappers if said utility functions are used.

In summary: 1.6 is the only branch that needs this.

drewish:
Will this work with nw.js?

kkoopa:
It **should** work with some version of nw.js, but I don't know for sure.
#285 might be related.

kkoopa:
The problem lies entirely with nw.js, which lies about every version imaginable. This is unmanageable.

rogerwang:
@kkoopa why did you say nw.js lies about the version? Please give more details -- we'd like to fix any potential issues at our side and make things smooth for other projects including Nan.

kkoopa:
Because nw-headers-v0.11.6.tar.gz states that it is node 0.11.13 in node-version.h, when the actual node 0.11.13 came with a different v8 version. There is no way of telling that this is not normal node 0.11.13.

rogerwang:
The discussion on NW.js moved to #285

PR #261, opened 2015-01-22
appVeyor improvements

mathiask88:
Try this one. Should run fine if I didn't miss something.


mathiask88:
And if you want to build not only x86 but also x64 I can add this too. Or VS13 instead of VS10, or both. But I could not get VS10+x64 get to work yet, because appVeyor has 2010 express installed that is based on Windows Platform SDK and the Linker can't find the kernel32.lib which is located in the SDK directory.

PR #265, opened 2015-01-23
fix a bug occurred the compilation error

XadillaX:
```
1.0.3/src/node.h:251:34: note: candidate function not viable: cannot convert argument of
      incomplete type 'const void *' to 'const char *'
NODE_EXTERN v8::Local<v8::Value> Encode(v8::Isolate* isolate,
                                 ^
```

https://github.com/rvagg/nan/issues/264

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1847,7 +1847,7 @@ NAN_INLINE v8::Local<v8::Value> NanEncode(
 #if (NODE_MODULE_VERSION > 0x000B)
   return node::Encode(
       v8::Isolate::GetCurrent()
-    , buf, len
+    , (const char *)buf, len
     , static_cast<node::encoding>(encoding));
 #else
 # if  (NODE_MODULE_VERSION < 0x000B)
--- a/nan.h
+++ b/nan.h
@@ -1844,20 +1844,27 @@ namespace Nan {
 
 NAN_INLINE v8::Local<v8::Value> NanEncode(
     const void *buf, size_t len, enum Nan::Encoding encoding = Nan::BINARY) {
+#if (NODE_MODULE_VERSION >= 42)
+  return node::Encode(
+      v8::Isolate::GetCurrent()
+    , static_cast<const char *>(buf), len
+    , static_cast<node::encoding>(encoding));
+#else
 #if (NODE_MODULE_VERSION > 0x000B)
   return node::Encode(
       v8::Isolate::GetCurrent()
-    , (const char *)buf, len
+    , buf, len
     , static_cast<node::encoding>(encoding));
 #else
-# if  (NODE_MODULE_VERSION < 0x000B)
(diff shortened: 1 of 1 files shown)


nan.h
```diff
@@ -1847,7 +1847,7 @@ NAN_INLINE v8::Local<v8::Value> NanEncode(
 #if (NODE_MODULE_VERSION > 0x000B)
   return node::Encode(
       v8::Isolate::GetCurrent()
-    , buf, len
+    , (const char *)buf, len
```
kkoopa:
```c++
# if (NODE_MODULE_VERSION >= 42)
    , static_cast<const char *>(buf)
# else
    , buf
# endif
```

kkoopa:
@bnoordhuis Why was this signature changed in io.js? It also says not to use UCS2. 0.8 only took char * as well, should NanEncode use `const void *` or `const char *`?

nan.h
XadillaX: @kkoopa  you mean use `static_cast<const char *>(buf)` to instead `(const char *)buf`?

kkoopa: Yes, but the cast is strictly necessary only for io.js.

But it might also be that NanEncode itself should take a const char * instead of a const void *. It would remove the need for casting and version checking, but would make less sense when using UCS2. Haven't decided on the best course of action.

bnoordhuis:
@kkoopa That's because of iojs/io.js@535fec8 and iojs/io.js@56fde66c.

XadillaX:
@bnoordhuis So now and later API will stay in `const char *`?

bnoordhuis:
@XadillaX That's correct.

kkoopa:
Is the bug present in 3.14? Should the UCS2 fix be backported to earlier versions?

bnoordhuis:
> Is the bug present in 3.14?

In principle, yes, although V8's CopyChars() implementation is just different enough that it doesn't seem to trigger gcc's SSE heuristics.  If you want to back-port the change, you have my blessing.

XadillaX:
> AppVeyor build failed.

any suggestion?

kkoopa:
Ignore it, io.js does not work on windows.

mathiask88:
Not yet ;) 1.0.4 will address this error I think.

mscdex:
+1

mathiask88:
+1 on getting this fixed, so that my addon becomes io.js ready :)

XadillaX:
Will this PR be merged? I'm in a hurry.

kkoopa:
More or less. I've been holding off because I want to find a way of accounting for the API change regarding UCS2 in io.js. Will however not be backporting changes to older versions, as that was too much work. Been busy with other responsibilities.

ETA: the coming week

hit9:
I'm hoping this fixed. :)

kkoopa:
Fixed through #273.

Issue #267, opened 2015-01-23
1.6 breaks more shit :) NaNEncode?


jonathanong:
```js
In file included from ../node_src/Shaheen.h:29:0,
                 from ../node_src/Shaheen.cc:2:
../node_modules/nan/nan.h: In function ‘v8::Local<v8::Value> NanEncode(const void*, std::size_t, Nan::Encoding)’:
../node_modules/nan/nan.h:1851:44: error: invalid conversion from ‘const void*’ to ‘const char*’ [-fpermissive]
     , static_cast<node::encoding>(encoding));
                                            ^
In file included from ../node_src/Shaheen.h:23:0,
                 from ../node_src/Shaheen.cc:2:
/home/vagrant/.node-gyp/1.0.3/src/node.h:251:34: error:   initializing argument 2 of ‘v8::Local<v8::Value> node::Encode(v8::Isolate*, const char*, size_t, node::encoding)’ [-fpermissive]
 NODE_EXTERN v8::Local<v8::Value> Encode(v8::Isolate* isolate,
                                  ^
```

thlorenz:
I ran into something similar when trying to make v8-profiler work with latest nan and iojs.
For now I upgraded nan there to `1.5.3` since the latest nan broke for io.js.

The stack trace is below and here is the [related v8-profiler PR](https://github.com/node-inspector/v8-profiler/pull/51).

```
nan 1.6.1

➝  node-gyp rebuild --nodedir ../iojs/ && ../iojs/deps/npm/bin/npm-cli.js test
gyp info it worked if it ends with ok
gyp info using node-gyp@1.0.2
gyp info using node@1.0.0 | darwin | x64
child_process: customFds option is deprecated, use stdio instead.
gyp info spawn python
gyp info spawn args [ '/Users/thlorenz/npm-global/lib/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'make',
gyp info spawn args   '-I',
gyp info spawn args   '/Volumes/d/dev/ns/nsolid/v8-profiler/build/config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/Users/thlorenz/npm-global/lib/node_modules/node-gyp/addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/Volumes/d/dev/ns/nsolid/iojs/common.gypi',
[...]

PR #273, opened 2015-02-05
NanEncode: fix argument type for node::Encode

### LINKED ISSUE #272: ../node_modules/nan/nan.h:1848:10: error: no matching function for call to 'Encode'
Just got this:

```
In file included from ../src/prf.cc:2:
../node_modules/nan/nan.h:1848:10: error: no matching function for call to 'Encode'
  return node::Encode(
         ^~~~~~~~~~~~
```

indutny:
`node::Encode` expects `const char *`, not `const void *`.

Fix #272

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1842,22 +1842,39 @@ namespace Nan {
   enum Encoding {ASCII, UTF8, BASE64, UCS2, BINARY, HEX, BUFFER};
 }
 
+#if !NODE_VERSION_AT_LEAST(0, 10, 0)
+# include "nan_string_bytes.h"  // NOLINT(build/include)
+#endif
+
 NAN_INLINE v8::Local<v8::Value> NanEncode(
     const void *buf, size_t len, enum Nan::Encoding encoding = Nan::BINARY) {
-#if (NODE_MODULE_VERSION > 0x000B)
+#if (NODE_MODULE_VERSION >= 42)
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  node::encoding node_enc = static_cast<node::encoding>(encoding);
+
+  if (encoding == Nan::UCS2) {
+    return node::Encode(
+        isolate
+      , reinterpret_cast<const uint16_t *>(buf)
+      , len / 2);
+  } else {
+    return node::Encode(
+        isolate
+      , reinterpret_cast<const char *>(buf)
+      , len
+      , node_enc);
+  }
+#elif (NODE_MODULE_VERSION > 0x000B)
   return node::Encode(
       v8::Isolate::GetCurrent()
     , buf, len
     , static_cast<node::encoding>(encoding));
 #else
-# if  (NODE_MODULE_VERSION < 0x000B)
-  if (encoding == Nan::BUFFER) {
-    assert(len <= node::Buffer::kMaxLength);
-    return v8::Local<v8::Value>::New(node::Buffer::New(
-        static_cast<char *>(const_cast<void *>(buf)), len)->handle_);
-  }
-# endif
+# if NODE_VERSION_AT_LEAST(0, 10, 0)
   return node::Encode(buf, len, static_cast<node::encoding>(encoding));
+# else
+  return NanIntern::Encode(reinterpret_cast<const char*>(buf), len, encoding);
+# endif
 #endif
 }
 
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ LINT_SOURCES = \
 	examples/async_pi_estimate/sync.cc \
 	examples/async_pi_estimate/sync.h \
 	nan.h \
+	nan_string_bytes.h \
 	nan_implementation_12_inl.h \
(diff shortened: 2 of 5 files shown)


ide:
Fixes #267 too (same issue)

indutny:
cc @rvagg @kkoopa

kkoopa:
It is more complicated than this. #265 has more information. It boils down to needing special case handling of UCS2 encoding while keeping the API consistent.

indutny:
@kkoopa looks better now?

indutny:
Oopps made a mistake. Force-pushing in a bit.

indutny:
@kkoopa Done

kkoopa:
Yes, this is much better. Don't see anything objectionable. Will see what test outcome is.

indutny:
@kkoopa either AppVeyor hang, or has very long queue...

kkoopa:
It usually has a very long queue.

nan.h
```diff
@@ -1845,10 +1845,21 @@ namespace Nan {
...
-    , buf, len
-    , static_cast<node::encoding>(encoding));
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  node::encoding node_enc = static_cast<node::encoding>(encoding);
+
+  if (encoding == Nan::UCS2) {
+    return node::Encode(
+        isolate
+      , reinterpret_cast<const uint16_t *>(buf)
+      , len);
```
kkoopa: This will not work properly on joyent/node 0.11.16, it will encode with the default encoding=node::BINARY. It is currently only io.js that has broken out UCS 2 handling.

kkoopa:
Can you add unit tests as well? With all the implicit arguments, this might actually compile even if it produces the wrong output.

indutny:
Not at the moment, busy with other stuff right now. Maybe tomorrow?

kkoopa:
Sure, it can wait until then.

indutny:
@kkoopa please review ;)

kkoopa:
Tests look good, but don't seem to pass at the moment. Also, you should still condition on NODE_MODULE_VERSION being >= 42 before calling `node::Encode(Isolate *, const uint16_t *, size_t)`. In earlier versions of the 0.11 series (e.g. joyent/node 0.11.16), that will resolve to `node::Encode(Isolate *, const uint16_t *, size_t, node::encoding = BINARY)` because the three-argument node::Encode does not exist.

kkoopa:
Tests look good, but don't seem to pass at the moment. Also, you should still condition on NODE_MODULE_VERSION being >= 42 before calling `node::Encode(Isolate *, const uint16_t *, size_t)`. In earlier versions of the 0.11 series (e.g. joyent/node 0.11.16), that will resolve to `node::Encode(Isolate *, const uint16_t *, size_t, node::encoding = BINARY)` because the three-argument node::Encode does not exist.

indutny:
Gosh, node.js 0.8 does not support HEX encoding. @kkoopa how much do we care about it? Should I polyfill it?

kkoopa:
Preferrably, yes. The API should be consistent across versions.

indutny:
@kkoopa there is another thing. It does not support any encoding except binary :) I guess I'll just put the node-v0.10 encoding function in.

Globegitter:
Is this PR supposed to supersede https://github.com/rvagg/nan/pull/265 or are these achieving slightly different things?
I am trying to build `node-sass` on iojs and getting it to work with https://github.com/rvagg/nan/pull/265 but not with this PR.

kkoopa:
It does not? How can I have missed that in the first place? How annoying.

indutny:
@kkoopa I'm porting string_bytes.h to nan.

kkoopa:
Alright, I guess it had to happen sooner or later. I've been putting it off for 
too long. Parts might have to be in a class, as you probably want some of it 
to be static methods. Switches and such don't inline too well. The good thing 
is that existing code can be cleaned up if it does not have to contain 
workarounds for missing parts of the string conversion routines.

indutny:
@kkoopa I hope this commit will do the job :)

indutny:
@kkoopa yay, travis is happy with it.

kkoopa:
Wonderful. I assume you have tested on io.js too, so that's all good. One final thing, now that you backported string_bytes.h, it might not be necessary to have the special BUFFER encoding handling
in `# if  (NODE_MODULE_VERSION < 0x000B)`. Removing it would clean up and simplify the code a little.

indutny:
@kkoopa does 0x000B means node.js v0.10? and < 0x000B means 0.8?

kkoopa:
Yes, that should be it. Had that special case because 0.8 did not support BUFFER encoding, but it was handled almost same way as BINARY.

indutny:
Ok, force pushed!

kkoopa:
Seems to work, indeed. Many thanks.

indutny:
You are welcome! Please release ;)

reqshark:
fixes `ws`

PR #274, opened 2015-02-05
Test on io.js on Travis

kkoopa:
Testing io.js on Travis


kkoopa:
@rvagg What's up with pangyp and node 0.8? https://travis-ci.org/rvagg/nan/jobs/49679768

Issue #268, opened 2015-02-01
Allow explicit context for NanCallback->Call()


mscdex:
Currently `NanCallback->Call()` always uses the global scope for `this` when executing the callback. It would be great to have `Call()` overloaded to accept a context, like `NanMakeCallback()`.

kkoopa:
Makes sense, this should be a straightforward addition.

PR #275, opened 2015-02-07
Made NanCallback::Call accept optional target.

kkoopa:
Addresses #268

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1402,15 +1402,49 @@ class NanCallback {
     return NanNew(handle)->Get(kCallbackIndex)->IsUndefined();
   }
 
-  v8::Handle<v8::Value> Call(int argc, v8::Handle<v8::Value> argv[]) const {
+  NAN_INLINE v8::Handle<v8::Value>
+  Call(v8::Handle<v8::Object> target
+     , int argc
+     , v8::Handle<v8::Value> argv[]) const {
+#if (NODE_MODULE_VERSION > 0x000B)  // 0.11.12+
+    v8::Isolate *isolate = v8::Isolate::GetCurrent();
+    return Call_(isolate, target, argc, argv);
+#else
+    return Call_(target, argc, argv);
+#endif
+  }
+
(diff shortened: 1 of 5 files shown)


rvagg:
LGTM

nan.h
```diff
@@ -1402,7 +1402,7 @@ class NanCallback {
     return NanNew(handle)->Get(kCallbackIndex)->IsUndefined();
   }
 
-  v8::Handle<v8::Value> Call(int argc, v8::Handle<v8::Value> argv[]) const {
+  v8::Handle<v8::Value> Call(int argc, v8::Handle<v8::Value> argv[], v8::Handle<v8::Object> target=NanGetCurrentContext()->Global()) const {
```
bnoordhuis: It's not wrong but the arguments are in the wrong order when compared to v8::Function::Call() or node::MakeCallback().  Why don't you overload Call() like this?

```c++
v8::Handle<v8::Value> Call(v8::Handle<v8::Object> recv, int argc, v8::Handle<v8::Value> argv[]) const {
  // ...
}

v8::Handle<v8::Value> Call(int argc, v8::Handle<v8::Value> argv[]) const {
  return Call(NanGetCurrentContext()->Global(), argc, argv);
}
```

(There's something of a lost performance opportunity there because NanGetCurrentContext() always retrieves the current isolate from thread-local storage.)

Aside: newer versions of V8 let you call a function with a non-object as the receiver.  I don't believe node::MakeCallback() currently reflects that, though.

kkoopa: Yes, I too thought that the arguments might be better of in the other order.

Since 0.10, `node::MakeCallback` does take `v8::Value` if it can be cast to `v8::Object`. Do you mean v8 now accepts any v8::Value? If it was not possible in the past, we probably cannot implement it anyway.

bnoordhuis: Yes, you can now pass null or undefined, for ES6 strict mode conformance.

kkoopa:
I have revised the implementation.

bnoordhuis:
LGTM

PR #278, opened 2015-02-09
Test NanNew<Context> better.

kkoopa:
@agnat This fails. What is the correct fix? An overload?

```
../../nan_new.h: In instantiation of ‘typename NanIntern::Factory<T>::return_t NanNew(A0, A1) [with T = v8::Context; A0 = long int; A1 = v8::Handle<v8::ObjectTemplate>; typename NanIntern::Factory<T>::return_t = v8::Local<v8::Context>]’:
../cpp/nannew.cpp:110:8:   required from here
../../nan_new.h:215:47: error: invalid conversion from ‘long int’ to ‘v8::ExtensionConfiguration*’ [-fpermissive]
   return NanIntern::Factory<T>::New(arg0, arg1);
```

### DIFF
--- a/test/cpp/nannew.cpp
+++ b/test/cpp/nannew.cpp
@@ -103,10 +103,13 @@ NAN_METHOD(testContext) {
   NanScope();
   NanTap t(args[0]);
 
-  t.plan(4);
+  t.plan(5);
   t.ok(_( assertType<Context>( NanNew<Context>())));
   ExtensionConfiguration extensions(0, NULL);
   t.ok(_( assertType<Context>( NanNew<Context>(&extensions))));
+  t.ok(_( assertType<Context>(
+          NanNew<Context>(reinterpret_cast<ExtensionConfiguration *>(NULL)
+          , Handle<ObjectTemplate>()))));
   t.ok(_( assertType<Context>(
           NanNew<Context>(&extensions, Handle<ObjectTemplate>()))));
   t.ok(_( assertType<Context>(


agnat:
Is this urgent? If you don't mind I'd like to push it on the stack for a couple of days...

kkoopa:
Not urgent, can be worked around with a typecast.

kkoopa:
This seems to be a seriously annoying problem. More so than expected. It won't accept `NULL`, because it is an integer, not a pointer. Making a specialization that takes an integer compiles, but then it warns about using passing `NULL` to non-pointer argument. Why can't the schizophrenic compiler make up its mind? Who thought it was a good idea to define a pointer as an integer in the first place, then to have the audacity to whine about it? `NULL` is a macro defined as `0L`, the compiler has no business whining if I then use it in place of an integer. `nullptr_t` does not work either, because it is verboten.

agnat:
This `NULL` business is a language defect.

Ok, probably to strong, but... here is a bit on the subject: http://stackoverflow.com/questions/7016861/why-are-null-pointers-defined-differently-in-c-and-c

C++11 users can use `nullptr`. I think the most sensible thing is to introduce a new signature without the extension config argument:

````
NanNew<Context>(Handle<ObjectTemplate>(), ...);
````

... and document that. The whole thing is pretty YAGNI anyway...

kkoopa:
It is more than a language defect. It is utterly retarded. `NULL` is, always has been and should always be `(void *) 0`, nothing else. How I loathe C++. If they wanted to fuck things up they should have made a new language and called it D, not Wrong C With Classes.

The new signature might be sensible, but I really dislike making an ugly exception. Better leave things as is and document that C++ is a shitty language instead. Have to use a typecast if `nullptr` is not available.

agnat:
You ain't gonna need it, so we won't do it... fine with me. ;)

agnat:
BTW, there is a language called D. Beautiful close-to-the-metal meta programming stuff, Just how we like it.

(I think _all_ single letter names are already taken :-P)

kkoopa:
Yes, I know there is a D language although I have never used it or seen more than its name, but I believe it is much newer than C++, so C++ should have called itself D and dropped all pretense of trying to be C-compatible when it is not. It could have saved so much suffering and made for a much cleaner, saner design. Digraphs and trigraphs also don't belong anywhere. Give me complex arithmetic instead. Operator overloading could also have been left out, it does not belong anywhere. But now I notice that I am ranting.

I'll change the test code to use a typecast and update the documentation.

agnat:
Hehe... not sure I'm following. You _don't_ want operator overloading but ask for complex arithmetic?

> But now I notice that I am ranting.

No shit, Sherlock.

:-P

kkoopa:
> Hehe... not sure I'm following. You don't want operator overloading but ask for complex arithmetic?
Exactly. Give it to me out of the box, set it in stone and never let it be changed again.

I fixed the test to use a typecast now.

PR #279, opened 2015-02-10
attempt to fix `NanNew<Integer>( sizeof(char) )`

TooTallNate:
I'm not sure if this is really the proper fix or not, but in previous versions of nan I have been able to do:

``` c++
size_t b = 1;
NanReturnValue(NanNew<Integer>(b));
```

But on 1.6.2 this no longer compiles :(

This seems to make it compile again, though I'm getting a compiler warning that I don't quite understand:

```
In file included from ../cpp/nannew.cpp:14:
In file included from ../../nan.h:80:
../../nan_new.h:109:26: warning: inline function 'NanIntern::IntegerFactory<v8::Integer>::New' is not defined [-Wundefined-inline]
  static inline return_t New(size_t value);
                         ^
../../nan_new.h:211:33: note: used here
  return NanIntern::Factory<T>::New(arg0);
                                ^
../../nan_new.h:109:26: warning: inline function 'NanIntern::IntegerFactory<v8::Int32>::New' is not defined [-Wundefined-inline]
  static inline return_t New(size_t value);
                         ^
../../nan_new.h:211:33: note: used here
  return NanIntern::Factory<T>::New(arg0);
                                ^
../../nan_new.h:122:26: warning: inline function 'NanIntern::Factory<v8::Uint32>::New' is not defined [-Wundefined-inline]
  static inline return_t New(size_t value);
                         ^
../../nan_new.h:211:33: note: used here
  return NanIntern::Factory<T>::New(arg0);
                                ^
3 warnings generated.
```

Hoping that we can get `size_t` support back so that I don't need to add a bunch of `static_cast<int32_t>(b)` casts to my modules. Cheers!

### DIFF
--- a/nan_new.h
+++ b/nan_new.h
@@ -106,6 +106,7 @@ struct Factory<v8::NumberObject> : FactoryBase<v8::NumberObject> {
 template <typename T>
 struct IntegerFactory : FactoryBase<T> {
   typedef typename FactoryBase<T>::return_t return_t;
+  static inline return_t New(size_t value);
   static inline return_t New(int32_t value);
   static inline return_t New(uint32_t value);
 };
@@ -118,6 +119,7 @@ struct Factory<v8::Int32> : IntegerFactory<v8::Int32> {};
 
 template <>
 struct Factory<v8::Uint32> : FactoryBase<v8::Uint32> {
+  static inline return_t New(size_t value);
   static inline return_t New(int32_t value);
   static inline return_t New(uint32_t value);
 };
--- a/test/cpp/nannew.cpp
+++ b/test/cpp/nannew.cpp
@@ -182,13 +182,16 @@ NAN_METHOD(testNumber) {
 
   t.ok(_( NanNew<Int32>(5)->Value() == 5 ));
   t.ok(_( NanNew<Int32>(-42)->Value() == -42 ));
+  t.ok(_( NanNew<Int32>(sizeof(uint8_t))->Value() == 1 ));
   t.ok(_( assertType<Int32>( NanNew<Int32>(23) )));
 
   t.ok(_( NanNew<Uint32>(5u)->Value() == 5u ));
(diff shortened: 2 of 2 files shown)


mathiask88:
https://github.com/rvagg/nan/issues/270

TooTallNate:
Well clearly my patch doesn't work correctly anyways.

This _did used to_ work though, in like v1.2.0 of nan, so there was a breaking change somewhere along the way.

agnat:
Thing is: Older versions of nan allowed for pretty wild implicit conversions. We fixed that in 1.5. Unfortunately some of these unintended conversions are actually in use... like `size_t`. So, yes. You'll probably need to add some sort of range checking and a cast. Sorry 'bout that.

(Your patch doesn't work because it only _declares_ two new functions but doesn't _define_ them. There is no function body. That's also what the warning is about. If you're curious the implementations are [here](https://github.com/rvagg/nan/blob/master/nan_implementation_12_inl.h) and [here](https://github.com/rvagg/nan/blob/master/nan_implementation_pre_12_inl.h))

kkoopa:
What he said. This is the correct way, even if it may be more inconvenient. If typing out the casting is a problem, I'd suggest defining a macro or helper function in your own project that does the casting.

TooTallNate:
I had a feeling that was gonna be the case. It sounds like 1.5.0 should have maybe been 2.0.0 :wink:

In my case the `size_t` are a result of a `sizeof()` call, so it's safe for me to assume that they're within Number range. I'll add casts as necessary.

Thanks for the explanation.

agnat:
>  It sounds like 1.5.0 should have maybe been 2.0.0

Hehe, yeah. Maybe... Generic template functions and version numbers are natural enemies. :wink:

> In my case the size_t are a result of a sizeof() call, so it's safe for me to assume that they're within Number range.

`sizeof(char[1<<55])`? :grin:

PR #280, opened 2015-02-10
Disconnect NanScope and NanReturnValue

kkoopa:
Got rid of a bunch of `NanScope`s, this could improve performance. To get old behavior, use `NanEscapableScope` and combine `NanEscapeScope` with `NanReturnValue`, but this should not be necessary on v8-exposed methods (NAN_METHOD, etc).

I guess this should be held off until NAN 2, as it would break backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,8 +161,6 @@ using v8::Number;
 
 // Simple synchronous access to the `Estimate()` function
 NAN_METHOD(CalculateSync) {
-  NanScope();
-
   // expect a number as the first argument
   int points = args[0]->Uint32Value();
   double est = Estimate(points);
@@ -228,8 +226,6 @@ class PiWorker : public NanAsyncWorker {
 
 // Asynchronous access to the `Estimate()` function
 NAN_METHOD(CalculateAsync) {
-  NanScope();
-
   int points = args[0]->Uint32Value();
   NanCallback *callback = new NanCallback(args[1].As<Function>());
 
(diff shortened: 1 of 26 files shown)


rvagg:
IMO our next release should be NAN2 and we should remove all 0.11.x code while we're at it.

kkoopa:
Sounds like a plan. Time to break out the sledgehammer.

reqshark:
but isn't a lot of the 0.11.x stuff used for 0.12?

kkoopa:
Yes, that will not be removed. Just a few things that had been changed late in the 0.11 cycle.

kkoopa:
Any thoughts on this? Will it lead to unforeseen problems somehow?

bnoordhuis:
NODWIB, I mean LGTM.

kkoopa:
Will the contract from the side of node / io.js always be that there is a surrounding `HandleScope`? Essentially, is this guaranteed to be proper / correct?

bnoordhuis:
The implicit HandleScope has been there since 2009 or 2010 and I can't think of a reason why the V8 people (or io.js or node.js for that matter) would want to ax it.

kkoopa:
Haha, one can't thing of a reason why they would want to ax many things, until they suddenly do it. But okay, I think we can go with this then. It saves typing `NanScope();` at the start of every method, even if you don't create any handles at all.

bnoordhuis:
On second thought, there may an issue with this change.  I suspect that many people write code like this:

```c++
NAN_METHOD(F) {
  NanScope();
  // ...
  NanReturnValue(value);
}
```

Without a `scope.Close()`, that's not going to work.

kkoopa:
This is a breaking change, yes, under consideration for 2.0. You cannot have the `NanScope()` there anymore. If you want the old behavior, you do `NanEscapableScope()` paired with `NanEscapeScope()`, e.g. `NanReturnValue(NanEscapeScope(foo));`.

The benefit is that you don't need scopes in external methods, which usually constitute the bulk of a binding.

kkoopa:
Making new PR targetting next.

PR #299, opened 2015-03-27
feat(SaveToPersistent): Allow saving v8::Value to persistent

mbroadst:
NanAsyncWorker previously required saving a v8::Objec to persistent
which did not allow any other value types to be stored. This change
opts for the more generic v8::Value, allowing you to store all
value types in the persistent store.

**NOTE:** I have no idea what kind of requirements you guys have for BC (well okay not BC in this case, but whatever-not-changing-private-api-is-called), this is obviously a breaking change and could be implemented with new accessors/mutators instead. Thoughts?

### DIFF
--- a/test/cpp/bufferworkerpersistent.cpp
+++ b/test/cpp/bufferworkerpersistent.cpp
@@ -35,9 +35,8 @@ class BufferWorker : public NanAsyncWorker {
   void HandleOKCallback () {
     NanScope();
 
-    v8::Local<v8::Object> handle = GetFromPersistent("buffer");
-    v8::Local<v8::Value> argv[] = { handle };
-    callback->Call(1, argv);
+    v8::Local<v8::Value> handle = GetFromPersistent("buffer");
+    callback->Call(1, &handle);
   }
 
  private:
--- a/nan.h
+++ b/nan.h
@@ -1511,15 +1511,15 @@ class NanCallback {
   }
 
   NAN_INLINE void SaveToPersistent(
-      const char *key, const v8::Local<v8::Object> &obj) {
+      const char *key, const v8::Local<v8::Value> &value) {
     v8::Local<v8::Object> handle = NanNew(persistentHandle);
-    handle->Set(NanNew<v8::String>(key), obj);
+    handle->Set(NanNew<v8::String>(key), value);
   }
 
-  v8::Local<v8::Object> GetFromPersistent(const char *key) const {
+  v8::Local<v8::Value> GetFromPersistent(const char *key) const {
     NanEscapableScope();
     v8::Local<v8::Object> handle = NanNew(persistentHandle);
-    return NanEscapeScope(handle->Get(NanNew(key)).As<v8::Object>());
+    return NanEscapeScope(handle->Get(NanNew(key)));
   }
 
   virtual void Execute() = 0;


mbroadst:
As a follow up question: the reason I wanted to do this was because I wanted to make a generic worker with variable return value. It's still not working because I think I'm getting scoping wrong trying to do e.g.:

```
void MyWorker::Execute() {
    // do some stuff
    saveToPersistent("result", NanNew("test value"));
}
```

crashes every time, any ideas?

mbroadst:
Oh this is going to be the same problem I've been having, in that I can't actually save v8 values anywhere from within Execute huh?

test/cpp/bufferworkerpersistent.cpp
```diff
@@ -35,8 +35,8 @@ class BufferWorker : public NanAsyncWorker {
   void HandleOKCallback () {
     NanScope();
 
-    v8::Local<v8::Object> handle = GetFromPersistent("buffer");
-    v8::Local<v8::Value> argv[] = { handle };
+    v8::Local<v8::Value> handle = GetFromPersistent("buffer");
+    v8::Local<v8::Value> argv[] = { handle.As<v8::Object>() };
     callback->Call(1, argv);
```
bnoordhuis: You can rewrite this whole block as:

```c++
v8::Local<v8::Value> handle = GetFromPersistent("buffer");
callback->Call(1, &handle);
```

bnoordhuis:
> Oh this is going to be the same problem I've been having, in that I can't actually save v8 values anywhere from within Execute huh?

Correct. :-)

The PR LGTM but it's a breaking change in that it will stop some code from compiling.  Over to @kkoopa.

kkoopa:
Yeah, can probably pull this in for NAN 2, as there will be other breaking changes there.

mbroadst:
@bnoordhuis yeah there's no good way to make this a non-breaking change unfortunately since just overloading SaveToPersistent with Value leads to ambiguity.. so I guess we'll see what @kkoopa says.

wrt the other question, can you point to any good examples of how I might go about solving my problem? I'm currently faced with binding a pretty massive C library in node and I'm desperately trying to reduce the code dupe for most of these accessor types of methods (GetCapabilities, GetHostname, GetWahtever, GetFoo) which all need to be async workers so as not to block the event loop. The problem is that they return variable types (Strings, Numbers, Arrays, Objects), and so I was trying to be tricky here and just convert the data in the Execute method and save as a v8::Value - no dice..

kkoopa:
Extend `NanAsyncWorker` with a template class.

mbroadst:
@kkoopa yes that's what I was trying to do initially, and that's not too bad with primitive types (int, bool, char*, etc), overriding the Execute() method to store in a local value for a template and then converting it to javascript in HandleOKCallback. However, in the case where I've got an array or a map it becomes somewhat more difficult. 

I feel like this is a bit of a pain point for nan, in that I have to temporarily store that data somewhere to wait to get back into the original calling context in order to covert it to the javascript type. I guess I'm asking if there is any example of being able to build these javascript types right at the point that I obtain the data? I feel that would reduce the complexity of these types of tasks (which I imagine many people run into)

kkoopa:
Merged in ffce18d0005ac28ba9b6a775727289daac844a0d.

PR #302, opened 2015-04-03
Feature request: Adding comparison operators to NanCallback

bbuecherl:
This simplifies comparison of `NanCallback`s constructed with the same `v8::Local<v8::Function>`-object (see the tests in `test/cpp/nancallback.cpp` for use-cases). What do you think about it?

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1393,6 +1393,17 @@ class NanCallback {
     NanDisposePersistent(handle);
   }
 
+  bool operator==(const NanCallback &other) const {
+    NanScope();
+    v8::Local<v8::Value> a = NanNew(handle)->Get(kCallbackIndex);
+    v8::Local<v8::Value> b = NanNew(other.handle)->Get(kCallbackIndex);
+    return a->StrictEquals(b);
+  }
+
+  bool operator!=(const NanCallback &other) const {
+    return !this->operator==(other);
+  }
+
   NAN_INLINE void SetFunction(const v8::Handle<v8::Function> &fn) {
     NanScope();
     NanNew(handle)->Set(kCallbackIndex, fn);
--- a/test/cpp/nancallback.cpp
+++ b/test/cpp/nancallback.cpp
@@ -24,6 +24,16 @@ NAN_METHOD(SpecificContext) {
   NanReturnUndefined();
 }
 
+NAN_METHOD(CompareCallbacks) {
+  NanScope();
+
+  NanCallback cb1(args[0].As<v8::Function>());
+  NanCallback cb2(args[1].As<v8::Function>());
+  NanCallback cb3(args[2].As<v8::Function>());
+
+  NanReturnValue(NanNew<v8::Boolean>(cb1 == cb2 && cb1 != cb3));
+}
+
 void Init (v8::Handle<v8::Object> target) {
   target->Set(
       NanNew<v8::String>("globalContext")
@@ -32,6 +42,10 @@ void Init (v8::Handle<v8::Object> target) {
   target->Set(
       NanNew<v8::String>("specificContext")
     , NanNew<v8::FunctionTemplate>(SpecificContext)->GetFunction()
+  );
+  target->Set(
+      NanNew<v8::String>("compareCallbacks")
+    , NanNew<v8::FunctionTemplate>(CompareCallbacks)->GetFunction()
   );
 }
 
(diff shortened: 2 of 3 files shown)


kkoopa:
I guess it's alright.

nan.h
```diff
@@ -1393,6 +1393,15 @@ class NanCallback {
     NanDisposePersistent(handle);
   }
 
+  friend bool operator==(NanCallback &cb1, NanCallback &cb2) {
+    if(cb1.IsEmpty() && cb2.IsEmpty()) return TRUE;
```
kkoopa: return true;

test/cpp/nancallback.cpp
```diff
@@ -24,6 +24,16 @@ NAN_METHOD(SpecificContext) {
...
 }
 
+NAN_METHOD(CompareCallbacks) {
+  NanScope();
+
+  NanCallback cb1(args[0].As<v8::Function>());
+  NanCallback cb2(args[1].As<v8::Function>());
+  NanCallback cb3(args[3].As<v8::Function>());
+
+  NanReturnValue(cb1 == cb2 && cb1 != cb3);
```
kkoopa: NanReturnValue(NanNew(cb1 == cb2 && cb1 != cb3));

nan.h
```diff
@@ -1393,6 +1393,15 @@ class NanCallback {
     NanDisposePersistent(handle);
   }
 
+  friend bool operator==(NanCallback &cb1, NanCallback &cb2) {
+    if(cb1.IsEmpty() && cb2.IsEmpty()) return 1;
```
bnoordhuis: Style issues: space after if and true instead of 1.  Using `friend` for a member method is superfluous.  There's also a logic error in the IsEmpty() checks, it should be:

```c++
if (cb1.IsEmpty() || cb2.IsEmpty()) {
  return cb1.IsEmpty() && cb2.IsEmpty();
}
return cb1.GetFunction() == cb2.GetFunction();
```

Or maybe just rewrite it as:

```c++
// More efficient.
Local<Value> a = NanNew(cb1.handle)->Get(kCallbackIndex);
Local<Value> b = NanNew(cb2.handle)->Get(kCallbackIndex);
return a->StrictEquals(b);
```

test/cpp/nancallback.cpp
```diff
@@ -24,6 +24,16 @@ NAN_METHOD(SpecificContext) {
   NanReturnUndefined();
 }
 
+NAN_METHOD(CompareCallbacks) {
+  NanScope();
+
+  NanCallback cb1(args[0].As<v8::Function>());
+  NanCallback cb2(args[1].As<v8::Function>());
+  NanCallback cb3(args[3].As<v8::Function>());
```
bnoordhuis: Off by one?

nan.h
bnoordhuis: Also... are you sure these methods are actually called?  It's a member function so the prototype should be:

```c++
bool operator==(const NanCallback& other) const { /* ... */ }
```

bbuecherl:
sorry, forgive my newbiness, I've fixed code style and effiency, should be fine now?

nan.h
```diff
@@ -1393,6 +1393,19 @@ class NanCallback {
     NanDisposePersistent(handle);
   }
 
+  bool operator==(const NanCallback &other) const {
+    v8::Local<v8::Value> a = NanNew(handle)->Get(kCallbackIndex);
+    v8::Local<v8::Value> b = NanNew(other.handle)->Get(kCallbackIndex);
+    if (a->IsUndefined() || b->IsUndefined()) {
+      return a->IsUndefined() && b->IsUndefined();
+    }
+    return a->StrictEquals(b);
```
bnoordhuis: Now that you're using StrictEquals() you don't need the IsUndefined() checks any more.

bnoordhuis:
LGTM with a comment.

kkoopa:
If you would finally squash into one commit, I'll be happy to merge.

nan.h
```diff
@@ -1393,6 +1393,16 @@ class NanCallback {
     NanDisposePersistent(handle);
   }
 
+  bool operator==(const NanCallback &other) const {
+    v8::Local<v8::Value> a = NanNew(handle)->Get(kCallbackIndex);
+    v8::Local<v8::Value> b = NanNew(other.handle)->Get(kCallbackIndex);
+    return a->StrictEquals(b);
```
kkoopa: Hold on, won't this leak handles? Should perhaps use a NanScope or make it inline?

bnoordhuis: Yes, needs a NanScope.  I'd add it and land it.

bbuecherl:
fixed the `NanScope()#` and squashed commits in #305

PR #304, opened 2015-04-03
Replaced NanObjectWrapHandle macro with function

kkoopa:
`node::ObjectWrap::handle()` should get a const qualifier


bnoordhuis:
LGTM.  In hindsight, I probably should have made `node::ObjectWrap::handle()` const.

kkoopa:
It's never too late to redeem yourself.

Issue #306, opened 2015-04-07
Backport thread local storage (TLS)


gagern:
libuv/libuv@5d2434bf71e47802841bad218d521fa254d1ca2d from joyent/libuv#904 added support for thread local storage. Would it be in scope for Nan to backport this in order to offer TLS support for node modules even when compiling against older versions of node?

This question was raised in https://github.com/polotek/libxmljs/pull/300#commitcomment-10607079 discussing gagern/libxmljs@55a686129f11fe4f9db63c7e3d0268b4cef5c45e.

If you are interested in this, I can provide a pull request. Or you can paste the relevant code yourself, once you have decided where it should go. It might make sense to declare the functions as `inline` or `static` to avoid having the same function in multiple source files, and so that dead code elimination can drop them if they are unused.

kkoopa:
Sure. It would at least be appropriate for NAUV (NAN for UV), which does not exist as a separate project yet, so it could go with the other uv stuff of NAN for now. Please provide a PR, it might be suitable to insert the methods after [line 101](https://github.com/rvagg/nan/blob/master/nan.h#L101) and make their names match the NAN scheme.

PR #307, opened 2015-04-07
Ticket #306: Backport thread local storage

gagern:
This makes thread local storage available to older versions of libuv as well.  It is essentially a backport of libuv/libuv@5d2434bf71e47802841bad218d521fa254d1ca2d, written by Ben Noordhuis.  But the names were changed, and the functions were turned into inline functions to avoid having the same non-inline function in multiple translation units.  The test is a lot more node-based, even if the things to test more or less follow those tested in libuv.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -100,6 +100,94 @@ typedef v8::String::ExternalOneByteStringResource
     void func(uv_async_t *async)
 #endif
 
+#if NAUV_UVVERSION >= 0x000b0b
+
+typedef uv_key_t nauv_key_t;
+
+inline int nauv_key_create(nauv_key_t *key) {
+  return uv_key_create(key);
+}
+
+inline void nauv_key_delete(nauv_key_t *key) {
+  uv_key_delete(key);
+}
+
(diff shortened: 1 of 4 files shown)


kkoopa:
@bnoordhuis You seem qualified to check it. It seems alright to me. Nice with passing tests too.

nan.h
```diff
@@ -100,6 +100,94 @@ typedef v8::String::ExternalOneByteStringResource
...
+}
+
+inline void nauv_key_set(nauv_key_t* key, void* value) {
+  if (pthread_setspecific(*key, value))
+    abort();
+}
+
+#else
+
+#include <Windows.h>
```
bnoordhuis: Can you write this as `#include <windows.h>`?  It doesn't matter on Windows but people trying to cross-compile on a case-sensitive file system will have trouble.

bnoordhuis:
LGTM with a suggestion.

kkoopa:
Thanks. It will be out in 1.8.0, but I don't know when that is.

Issue #289, opened 2015-02-27
Heads up: large incompatible V8 API changes coming up


bnoordhuis:
https://code.google.com/p/v8/issues/detail?id=3929

https://groups.google.com/forum/#!topic/v8-users/gQVpp1HmbqM - outlines the changes.

Some of the changes already landed upstream and will land in io.js in two or three months.

kkoopa:
Oh my. Again?

kkoopa:
I wonder if these maybe monads can be backported to the old API somehow. It would probably be the nicer option.

kkoopa:
kkoopa/nan@c66082dee22375c65cbd2563a4457cf7fb7eeac5

kkoopa:
Made a small list of major changes so far in v8 4.3, it is not exhaustive. All of these have been adapted to return `Maybe`s.
```c++
Script::Compile
Script::Run
Script::CompileUnbound

Value::Converters
Value::Equals

String Factory Methods

Object::Mostly Everything

Array::CloneElementAt

Function::NewInstance
Function::Call

Date::New

RegExp::New

FunctionTemplate::GetFunction

ObjectTemplate::NewInstance

TryCatch::StackTrace
```
and then IdleNotification was renamed and changed its argument `bool IdleNotification(int idle_time_in_ms) --> bool IdleNotificationDeadline(double deadline_in_seconds)`

kkoopa:
Having looked at the necessary changes as of latest 4.3, it does not make sense to backport the `Maybe` types. It is easier to just unwrap the necessary return values for new versions that use maybe types internally. No v8 functions seem to accept `MaybeLocal<T>` or `Maybe<T>` as arguments, so the whole system seems quite broken. You always end up wanting to unwrap.

domenic:
> You always end up wanting to unwrap.

Sure, but you want to unwrap and decide for yourself how to handle the None case, right? Instead of #310, which just calls `.ToLocalChecked()` always.

There are a number of issues opened on Chromium where incorrect `.ToLocalChecked()` changes cause crashes. It seems better if nan does not automatically encourage this buggy style.

kkoopa:
I thought of emulating old behavior, not doing a check in the conversion, essentially getting rid of the new API altogether by making it conform to the old. But, if this is too restrictive, the other alternative is backporting the new API, as I originally did https://github.com/kkoopa/nan/commit/c66082dee22375c65cbd2563a4457cf7fb7eeac5.

[...]

PR #317, opened 2015-04-21
Backported Maybe types from V8 v4.3

kkoopa:
This is the alternative solution to #289.
Currently forces the empty `String` constructor to return `NanMaybeLocal<String>`.

Now `NanNew<T>` will return a `NanMaybeLocal<T>` for some `T` (`NanBoundScript`, `NanUnboundScript`, `Date`, `RegExp` and `String`). For the convenience overload `NanNew(...)`, this thus only applies to the `String` types. This seems a little clumsy. Should the convenience method do `NanMaybeLocal::ToLocalChecked()`?


kkoopa:
Closing for new PR targeted at next.

PR #321, opened 2015-04-25
Use <cstdlib> for abort()

saper:
In https://github.com/sass/node-sass/issues/892
we have a report that node-sass compilation
may fail due to undefined abort().

I think we should not rely on node or uv to
include <stdlib.h> for us.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/abort.html

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -27,6 +27,7 @@
 #include <node_object_wrap.h>
 #include <cstring>
 #include <climits>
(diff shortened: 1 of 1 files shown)


kkoopa:
Thanks. Offending code is a new addition, it compiled fine in every remotely sane environment tested, so it was not caught until now.

PR #323, opened 2015-04-29
Added more error types and made them all accept Handle<String> as argument

### LINKED ISSUE #322: NanThrowError(Handle<Value>)
```c++
NAN_INLINE void NanThrowError(v8::Handle<v8::Value> error) {
    NanScope();
    v8::Isolate::GetCurrent()->ThrowException(error);
  }
```
This should probably be changed to
```c++
NAN_INLINE void NanThrowError(v8::Handle<v8::Value> error) {
    NanScope();
    v8::Isolate::GetCurrent()->ThrowException(v8::Exception::Error(error));
  }
```
and the same goes for the old version.

kkoopa:
Fixes #322.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ LINT_SOURCES = \
 	test/cpp/asyncprogressworker.cpp \
 	test/cpp/asyncworkererror.cpp \
 	test/cpp/bufferworkerpersistent.cpp \
+	test/cpp/error.cpp \
 	test/cpp/gc.cpp \
 	test/cpp/morenews.cpp \
 	test/cpp/multifile1.cpp \
--- a/nan.h
+++ b/nan.h
@@ -644,16 +644,16 @@ return args.GetReturnValue().Set(Nan::imp::NanEnsureHandleOrPersistent(value))
       info_->persistent.SetWeak(info_, &_NanWeakCallbackDispatcher<T, P>);
   }
 
-template<typename T, typename P>
-NAN_INLINE _NanWeakCallbackInfo<T, P>* NanMakeWeakPersistent(
-    v8::Handle<T> handle
-  , P* parameter
-  , typename _NanWeakCallbackInfo<T, P>::Callback callback) {
-    _NanWeakCallbackInfo<T, P> *cbinfo =
-     new _NanWeakCallbackInfo<T, P>(handle, parameter, callback);
-    cbinfo->persistent.SetWeak(cbinfo, &_NanWeakCallbackDispatcher<T, P>);
-    return cbinfo;
-}
+  template<typename T, typename P>
+  NAN_INLINE _NanWeakCallbackInfo<T, P>* NanMakeWeakPersistent(
+      v8::Handle<T> handle
+    , P* parameter
+    , typename _NanWeakCallbackInfo<T, P>::Callback callback) {
+      _NanWeakCallbackInfo<T, P> *cbinfo =
+       new _NanWeakCallbackInfo<T, P>(handle, parameter, callback);
+      cbinfo->persistent.SetWeak(cbinfo, &_NanWeakCallbackDispatcher<T, P>);
+      return cbinfo;
+  }
 
 # define NAN_WEAK_CALLBACK(name)                                               \
     template<typename T, typename P>                                           \
--- a/nan.h
+++ b/nan.h
@@ -659,58 +659,64 @@ NAN_INLINE _NanWeakCallbackInfo<T, P>* NanMakeWeakPersistent(
     template<typename T, typename P>                                           \
     static void name(const _NanWeakCallbackData<T, P> &data)
 
-# define _NAN_ERROR(fun, errmsg) fun(NanNew<v8::String>(errmsg))
-
-# define _NAN_THROW_ERROR(fun, errmsg)                                         \
-    do {                                                                       \
+# define X(NAME)                                                               \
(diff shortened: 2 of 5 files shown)


kkoopa:
My only concern here is with the Scopes. I tried to remove some where unnecessary and add them where needed.

kkoopa:
@bnoordhuis Do you have time to look this over?

Makefile
```diff
@@ -23,6 +23,7 @@ LINT_SOURCES = \
         test/cpp/asyncprogressworker.cpp \
         test/cpp/asyncworkererror.cpp \
         test/cpp/bufferworkerpersistent.cpp \
+	test/cpp/error.cpp \
```
bnoordhuis: Indentation seems a bit off here.  Maybe that's because it's a tab instead of spaces like on the surrounding lines?

kkoopa: Could be. I normalized everything to tabs in 80cf724cf898ac5568b52d44b06c0d5c0ebf0764, but this commit is older. I'll rebase with master or something.

nan.h
```diff
@@ -637,18 +645,76 @@ NAN_INLINE _NanWeakCallbackInfo<T, P>* NanMakeWeakPersistent(
...
+  NAN_INLINE void NanThrowReferenceError(const char* errmsg) {
+    _NAN_THROW_ERROR(v8::Exception::ReferenceError, errmsg);
+  }
+
+  NAN_INLINE void NanThrowReferenceError(v8::Handle<v8::String> errmsg) {
+    NanScope();
+    v8::Isolate::GetCurrent()->ThrowException(
+        v8::Exception::ReferenceError(errmsg));
+  }
+
```
bnoordhuis: I would probably use an x-macro for this.  There is a lot of duplication with only minor changes in functionality.  It makes it somewhat hard to eyeball what the differences are and if they are correct.

An x-macro would probably also let you get rid of the _NAN_THROW_ERROR macro.

nan.h
```diff
@@ -1169,24 +1242,80 @@ NAN_INLINE _NanWeakCallbackInfo<T, P>* NanMakeWeakPersistent(
...
     _NAN_THROW_ERROR(v8::Exception::RangeError, errmsg);
   }
 
+  NAN_INLINE
+  v8::Local<v8::Value> NanThrowRangeError(v8::Handle<v8::String> errmsg) {
+    NanEscapableScope();
+    return NanEscapeScope(v8::Local<v8::Value>::New(
+        v8::ThrowException(v8::Exception::RangeError(errmsg))));
+  }
+
```
bnoordhuis: Do you need the NanScopes / NanEscapableScopes in the NanThrow methods?

kkoopa: Dunno. Unnecessary or missing scopes are the main thing I'm concerned with. Here I added it because `Exception::RangeError` creates a `Local<Handle>`, which is returned by `ThrowException` and wrapped in a `Local<Value>` which is returned to the caller. Thus calling this function involves at least two handle creations, only one of which should bubble up to the surrounding `HandleScope`.

bnoordhuis: Okay.  It's not wrong to use a HandleScope here, it's just that the overhead is pretty high compared to the meat of the function.  That probably doesn't matter so much here, though.

kkoopa:
Updated. I decided to deprecate the sugary error routines with the error code property, so that they can be removed for 2.0. I don't think it's in wide use, and it is trivial to reimplement for anyone that wants it.

kkoopa:
CI says it's good.

nan.h
```diff
@@ -644,73 +644,83 @@ return args.GetReturnValue().Set(Nan::imp::NanEnsureHandleOrPersistent(value))
...
 
 # define NAN_WEAK_CALLBACK(name)                                               \
     template<typename T, typename P>                                           \
     static void name(const _NanWeakCallbackData<T, P> &data)
 
-# define _NAN_ERROR(fun, errmsg) fun(NanNew<v8::String>(errmsg))
-
-# define _NAN_THROW_ERROR(fun, errmsg)                                         \
-    do {                                                                       \
+# define ERROR_TYPES                                                           \
```
bnoordhuis: Suggestion: name this NAN_ERROR_TYPES just in case a type/value/macro ERROR_TYPES exists elsewhere.  Or scrap it and call X(Error), X(RangeError), etc. directly below.

nan.h
```diff
@@ -1191,68 +1201,71 @@ NAN_INLINE _NanWeakCallbackInfo<T, P>* NanMakeWeakPersistent(
...
+  namespace Nan { namespace imp {
+    NAN_INLINE v8::Local<v8::Value> E(const char *msg, const int errorNumber) {
+      NanEscapableScope();
+      v8::Local<v8::Value> err = v8::Exception::Error(NanNew<v8::String>(msg));
+      v8::Local<v8::Object> obj = err.As<v8::Object>();
+      obj->Set(NanNew<v8::String>("code"), NanNew<v8::Int32>(errorNumber));
+      return NanEscapeScope(err);
+    }
+  }  // end of namespace imp
+  }  // end of namespace Nan
```
bnoordhuis: This duplicates the E function above, doesn't it?

kkoopa: They ought to be in different ifdefs, one for 0.10 and one for 0.12.

kkoopa:
Got rid of the `ERROR_TYPES` macro.

kkoopa:
Is this good to merge?

kkoopa:
Merged as 4265f06c880f9c166716c45b0aed05a90452f49d

PR #329, opened 2015-05-01
Renamed identifiers with leading underscores

### LINKED ISSUE #219: NAN uses identifiers with leading underscores
Leading underscores in names is a minefield. Better use trailing underscores to tag internal symbols. NAN currently uses names starting with `_N` and those are definitely reserved:

http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

Also, consider using the `NanInternal` namespace.

kkoopa:
Fixes #219 

Left some deprecated identifers which will be removed anyway.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -252,19 +252,21 @@ NAN_INLINE uint32_t NanUInt32OptionValue(
 template<typename T>
 v8::Local<T> NanNew(v8::Handle<T>);
 
-template<typename T>
-NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Handle<T> val) {
-  return NanNew(val);
-}
+namespace NanIntern {
+  template<typename T>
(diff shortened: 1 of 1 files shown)


kkoopa:
RFC

rvagg:
I don't mind, if you're using enough of these outside of NAN to see it as a general pattern. I've used a few of the macros myself before so can see the sense in this.

Is `NanIntern` an idiomatic naming for an internal C++ namespace?

kkoopa:
I will update the PR to reflect that `NanIntern` was renamed to `Nan::imp` in b4529bb78ed63d5aec9ecee97b1d46eb49fc3773. The main objective here was to first and foremost get rid of all identifiers with intial underscores, as that violates the standard. I have not reflected much on what to actually expose and not yet. Some of these can probably be moved to the internal namespace.

PR #334, opened 2015-05-07
Next

kkoopa:
Start collecting pull requests for NAN 2 here.


bnoordhuis:
LGTM although the CI seems (partially) unhappy.

Out of curiosity, why was NanNewContextHandle deprecated?

kkoopa:
AppVeyor hasn't pulled in build images for io.js 2 yet, they have it scheduled 
for today, so it should start working by itself soon enough.

NanNewContextHandle was replaced with NanNew<Context>.

kkoopa:
Only things missing now are:
* the new `Buffer` implementation
* sticking everything (but macros) in namespace `Nan`
* replacement of some `NanNew<String>` routines that should not be

kkoopa:
Don't know why crapveyor lists a failure although nothing failed. Can't reproduce.

kkoopa:
Probably fixed it now. It was due to a sort of broken test (more like broken io.js). Calling `gc()` is supposed to be synchronous, but it does not appear to be so really. Made a polling loop to circumvent it.

kkoopa:
A whole lot of documentation will need to be written and rewritten. Should we make use of Doxygen or some other similar alternative?

It would be nice to have proper documentation that did not have too much prerequisite API knowledge. In the past, this was less of a problem, as NAN was mostly used to port existing addons, but now almost all of them use NAN and the docs are more important for new developments.

Should NAN be documented the same way a core module would be documented, to fit in with other node docs?

P.S. I don't like writing documentation.

agnat:
I'm not a big fan of doxygen. It is either in the way or annoyingly uninformative. Sometimes both. I'd rather help with writing proper documentation. IMO it is the only way to describe things like `NanNew<>()` for the average reader.

P.S. I don't either ... but then who does?

kkoopa:
I'm not that big a fan of Doxygen myself either. I like reading its output, but don't like seeing it in the code or writing it. The reason I thought about it is because many classes are almost directly derived from V8. Please do help out with writing proper documentation, I have so many things to do that documentation gets rather low priority, even though I know that documentation is important for the end-user.

kkoopa:
Should we do some general restructuring? Grouping things into namespaces, perhaps into classes. I'd be tempted to stick the `Buffer` routines in a namespace or class. What about splitting into more files? 

@agnat As you did an excellent job on the `NanNew` rewrite, I wonder if you have anything in mind regarding these points?

agnat:
Hm, I don't feel the need to get the buffer stuff out of the way. It's only a hand full of functions. If anything I'd go with a namespace. Don't invent classes just to encapsulate a few functions.

I would however drop the Nan-prefix on everything in namespace Nan. I'd also go with an all lowercase `namespace nan`...

Regarding single vs. multiple files, I personally like to spread things out a bit. But that's just me...

kkoopa:
All lowercase namespace does not work. It is used by something existing and leads to conflicts. I'd drop the prefix if it weren't too much work.

trevnorris:
`NanCopyBuffer` is still using `Buffer::New`. I've updated the next branch to be `Buffer::Copy`.

kkoopa:
And `Buffer::Use` was renamed, yes. I have not gotten around to updating this yet. Been working on (another) migration tool. Will probably fix it this weekend.

trevnorris:
@kkoopa cool, figured. just wanted to make sure we didn't cross wires.

kkoopa:
Yes, it is highly appreciated.

kkoopa:
There, I've removed the Nan prefix from all names and updated the buffer routines for the latest changes.

Aside from missing documentation, general testing, cleanup and final reviewing, this ought to be near a final releasable state.

I will probably remove the conversion script from here and release that as a separate thing. I've been working on a better conversion tool based on libclang. The analysis part works rather well, aside from occasional segfaults on token processing, most likely due to a Buffer getting wrongly garbage collected somewhere. Unfortunately, the rewriting part still has some unsolved problems.

kkoopa:
Merged. 2.0.0 Is out.

@rvagg Nothing should block io.js 3.0 anymore.

rvagg:
Woohoo! Thanks @kkoopa, I'm working on release notes for v3 now. Should be out soon.

PR #335, opened 2015-05-08
Deprecate more cruft before 2.0

kkoopa:
`NanGetPointerSafe` and `NanSetPointerSafe` were only used in deprecated string conversion routines. Probably has no external users.

The OptionValue routines are just sugar with a narrow (?) use case.

/cc @rvagg: Objections?


rvagg:
no objections, I've only used them in the various leveldown* packages and put them in here in the assumption that they'd be more generally useful but they turn out to be a bit too specific to the pattern I was using and would be fine to see them removed (actually, we'll what kind of a fuss we create by deprecating them, that'll be interesting)

gigadude:
I'd just started using them... maybe move them into a nanutil.h (and round out the set of C++ types while you're at it)?

ralphtheninja:
Would it be possible to release other c-modules as well and use them just like nan?

ralphtheninja:
@kkoopa When will 2.0 be released you think?

kkoopa:
After the new `Buffer` implementation is done in io.js.

PR #337, opened 2015-05-08
Renamed identifiers with leading underscores

### LINKED ISSUE #219: NAN uses identifiers with leading underscores

kkoopa:
Retargeted against next
Replaces #329
Fixes #219

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -324,71 +324,53 @@ namespace Nan { namespace imp {
   }
 #endif
 
-#if (NODE_MODULE_VERSION > NODE_0_10_MODULE_VERSION)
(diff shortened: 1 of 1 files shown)


kkoopa:
Merged as c8e293efe762445c3e68d7a3baacb2edd1ba6fbd

PR #338, opened 2015-05-09
Workerpersistent

kkoopa:


rvagg:
lgtm, but why?

kkoopa:
because `SaveToPersistent(const char *key, const v8::Local<v8::Value> &value)` converts `key` to a `String` before setting. If you already have a `String`, you would have to first convert it to a `char *` only to have it converted back later.

rvagg:
ah, of course, then :+1:!

PR #339, opened 2015-05-09
Disconnect NanScope and NanReturnValue

kkoopa:
Continued from #280

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,8 +176,6 @@ using v8::Number;
 
 // Simple synchronous access to the `Estimate()` function
 NAN_METHOD(CalculateSync) {
-  NanScope();
-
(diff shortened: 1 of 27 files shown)


rvagg:
lgtm

I'm behind on the `NanEnsureHandleOrPersistent` change but why isn't it just `NanEnsureHandle`?

kkoopa:
It's an internal name. Persistent is not descended from Handle since 0.12, this ensures Persistent or Handle, because args.GetReturnValue.Set() accepts a Persistent or a Handle.

kkoopa:
Merged as bc99c11477fd808cc64d2ff854bbfd2b54410de2

PR #340, opened 2015-05-09
Replaced NanObjectWrapHandle with class NanObjectWrap

kkoopa:


kkoopa:
Merged as 9f59d2a4be0c9e985509670f6e9f67aec7e37a85

PR #341, opened 2015-05-09
Nomacro

kkoopa:
Replaced a bunch of macros with classes. The only ones left are `NanReturnValue` et al. Can't figure out a good way of removing them. I considered wrapping all exported functions similarly to the weak callback and adding a `NanExport` function, but that leaves the problem of resolving the right function when doing new `FunctionTemplate` or `Function`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,8 +176,6 @@ using v8::Number;
 
 // Simple synchronous access to the `Estimate()` function
 NAN_METHOD(CalculateSync) {
-  NanScope();
-
   // expect a number as the first argument
   int points = args[0]->Uint32Value();
   double est = Estimate(points);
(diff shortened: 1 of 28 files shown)


kkoopa:
I've come up with a plan that might work.

rvagg:
The only thing I'm concerned about here is having different addons compile against different versions of NAN, we're running a greater risk of ABI incompatibilities when moving away from macro's aren't we? Are we going to end up needing a `NAN_MODULE_VERSION` or of our own to do some checking and erroring at load time?

kkoopa:
NAN is a header only library, so ABI changes would not be a problem.

kkoopa:
Merged as e931e216c4e59ffddc462e1d4f11f247e53a83b2.

PR #344, opened 2015-05-10
Nonanreturn

kkoopa:
Remove `NanReturnValue`.


nan.h
```diff
@@ -90,12 +90,223 @@
...
+  inline void Set(bool value) { value_ = v8::Boolean::New(value); }
+  inline void Set(double i) { value_ = v8::Number::New(i); }
+  inline void Set(int32_t i) { value_ = v8::Int32::New(i); }
+  inline void Set(uint32_t i) { value_ = v8::Uint32::New(i); }
+
+  // Fast JS primitive setters
+  inline void SetNull() { value_ = v8::Null(); }
+  inline void SetUndefined() { value_ = v8::Undefined(); }
+  inline void SetEmptyString() { value_ = v8::String::Empty(); }
+};
```
bnoordhuis: I see at least one problem with this approach: it's going to be unreliable with nested HandleScopes.

```c++
NAN_METHOD(F) {
  NanReturnValue return_value;  // For the sake of argument
  {
    HandleScope handle_scope;
    Local<Object> value = Object::New();
    return_value.Set(value);
    // HandleScope goes out of scope, zapping |value|
  }  
}
```

You would have to use a v8::Persistent to avoid that.

kkoopa: Yes, that would be necessary. I tried with `Persistent` earlier when trying to track down the problem at hand, but it was not the cause. Couldn't compile a debug build of Node 0.10.38 either, linking mksnapshot failed for unknown reason.

Another problem is deleting the function pointer wrapper class after running the callback body. Is v8 somehow taking control of the pointer I set with `SetPointerInInternalField`? Tried with `v8::External` earlier, but that was definitely managed by v8 and needed weak callbacks, which could be quite detrimental to performance.

kkoopa:
Figured out what that problem was. Used a `Persistent` and added some indirection to make it work.
The problem now is that I cannot free the callback wrappers. If I uncomment [this](https://github.com/iojs/nan/pull/344/files#diff-558c37be49a47656e3ff41806c24bd33R258), the process segfaults.

ralphtheninja:
Why does it segfault?

kkoopa:
It tries to dereference the freed pointer on a second invocation... Ah, yes, that's because they are only created when creating the function template. They might not need freeing then.

rvagg:
@kkoopa why are we doing this?

kkoopa:
Gets rid of ugly hacks based on macros. It is confusing that you have to use `NanReturnValue` et al in exposed methods, but cannot use them in internal functions. This would enable a clean, consistent experience.

kkoopa:
There, now a poof of concept works on 0.10. I'll add the missing accessors and 0.12 support later.

kkoopa:
Missing accessors added. Now it just lacks 0.12 support and polish.

kkoopa:
Ready for initial review.

kkoopa:
It's easier to see the relevant changes [here](https://github.com/kkoopa/nan/compare/nomacro...kkoopa:nonanreturn).

rvagg:
I'm going to have to decline reviewing this one, it's _huge_ and I can't really fit it all in my head right now

kkoopa:
Heh, there is plenty of repetition with the same pattern applied to each callback type (function, getter, setter, deleter, etc.). It's more or less the same idea as with the weak callback.

kkoopa:
@trevnorris Would you be able to review this too? It's quite large, but much of it is repetition of the same pattern. It wraps the callbacks (Function, Getter, Setter, NamedPropertyGetter, etc), allowing for new callback signatures, similarly to the weak callback wrapper. It's a backport of the new style callback API.

kkoopa:
Merged as 6ee88d2d78c99080384d2233fe2b5cfd8eb0a66b

trevnorris:
@kkoopa sorry about not responding earlier. Finally getting through my backlog. Still want a review, or don't worry about it since it's already merged?

kkoopa:
Sure, but do the review on 6ee88d2d78c99080384d2233fe2b5cfd8eb0a66b. There were some additional changes when I merged all three open PRs.

kkoopa:
I moved some stuff between commits to make things easier. The commit to review is d0bbd6e046b48ddadbd6ee52c4c3f732cdfcf69d.

PR #345, opened 2015-05-13
Removed String converters

kkoopa:
`NanUtf8String` and `NanUcs2String` just duplicate v8's equivalent classes. `NanAsciiString` is not really an ASCII string, true ASCII is a subset of UTF-8, so that can be used instead.


kkoopa:
Merged in efced26a146e0ceb9cb78026dec4c5b3145dfcc8.

springmeyer:
@kkoopa - are you sure this is uneeded? I'm trying to port https://github.com/mapnik/node-mapnik/blob/a40d9ce8c6c44f201c5319d84c34d77f79072a9a/src/mapnik_palette.cpp#L45-L46 to Nan 2.0.5 and after switching to `Nan::Utf8String` I'm seeing test failures. It may be on my end, but I thought I'd let you know just in case you have ideas. I wonder if v8's `string->WriteAscii` behaves differently than `string->WriteUtf8`.

kkoopa:
They behave very differently. The old so called Ascii methods dealt in raw bytes with no actual encoding. This was rectified. The difference here is that you are not passing a valid ASCII string. Since ASCII is a subset of UTF-8, a valid ASCII string is valid UTF-8. Use a Buffer instead.

springmeyer:
@kkoopa - Thank you! That fixed it :)

PR #346, opened 2015-05-13
Fix gc text

kkoopa:
Debug builds would assert that it is not allowed to allocate v8 handles in gc callbacks.


bnoordhuis:
LGTM but I suspect you mean s/text/test/.

kkoopa:
Yup

kkoopa:
Merged as 2bd7bed3cba98264577ff82d0c86acf18365789f.

PR #348, opened 2015-05-13
Persistent

kkoopa:
Made a new `NanPersistent` class.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -89,6 +89,12 @@
 #define ATOM_0_21_MODULE_VERSION 41
 #define IOJS_1_0_MODULE_VERSION  42
 #define IOJS_1_1_MODULE_VERSION  43
+#define IOJS_2_0_MODULE_VERSION  44
+
+#define TYPE_CHECK(T, S)                                                       \
+    while (false) {                                                            \
+      *(static_cast<T *volatile *>(0)) = static_cast<S*>(0);                   \
+    }
 
 #if (NODE_MODULE_VERSION < NODE_0_12_MODULE_VERSION)
 typedef v8::InvocationCallback NanFunctionCallback;
@@ -106,6 +112,28 @@ typedef v8::String::ExternalAsciiStringResource
 #else
 typedef v8::String::ExternalOneByteStringResource
     NanExternalOneByteStringResource;
+#endif
+
+#if NODE_MODULE_VERSION > NODE_0_10_MODULE_VERSION
+template<typename T>
+class NanNonCopyablePersistentTraits :
+    public v8::NonCopyablePersistentTraits<T> {};
+template<typename T>
+class NanCopyablePersistentTraits :
+    public v8::CopyablePersistentTraits<T> {};
+
+template<typename T>
+class NanPersistentBase :
+    public v8::PersistentBase<T> {};
+
+template<typename T, typename M = v8::NonCopyablePersistentTraits<T> >
+class NanPersistent;
+#else
+template<typename T> class NanNonCopyablePersistentTraits;
+template<typename T> class NanPersistentBase;
+template<typename T, typename P> class NanWeakCallbackData;
+template<typename T, typename M = NanNonCopyablePersistentTraits<T> >
+class NanPersistent;
 #endif
 
 #include "nan_new.h"  // NOLINT(build/include)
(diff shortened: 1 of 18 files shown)


kkoopa:
@bnoordhuis Could you review this, please? Or suggest someone else who can do it.

kkoopa:
Aaaand the weak callback will change again in 4.3, so that will require more rewriting again. I'm getting fed up with this.

bnoordhuis:
Maybe @trevnorris is willing to review?  I'm a bit short on time at the moment.

trevnorris:
I'm fine to do a review, but it's been a while since I've looked at the nan source.

kkoopa:
No need to look at old source (much of it is getting rewritten anyway). This is more or less standalone, it's a backport of the new Persistent handles to old v8, with a unified API, plus a redone `PersistentBase<T>::SetWeak` which accepts a different weak callback signature.

nan.h
```diff
@@ -1527,26 +1755,28 @@ class NanObjectWrap : public node::ObjectWrap {
...
+
+ public:
   typedef void (*Callback)(
       NanWeakCallbackData & data  // NOLINT(runtime/references)
       );
-  NanWeakCallbackData(v8::Handle<T> handle, P* param, Callback cb)
-    : parameter(param), callback(cb) {
-    NanAssignPersistent(persistent, handle);
-    Revive();
+  NanWeakCallbackData(v8::Persistent<T> *handle, P* param, Callback cb)
```
trevnorris: there a preference on placement of `*`?

kkoopa: Yes, should be `T *foo`. V8 does it the other way around, I've copied much of the class definitions from v8.h. Probably missed many of these.

trevnorris:
Changes look good, all tests pass. LGTM.

nan.h
```diff
@@ -1527,26 +1755,28 @@ class NanObjectWrap : public node::ObjectWrap {
...
+    : persistent(), parameter(param), callback(cb) {
+    std::memcpy(&persistent, handle, sizeof(persistent));
```
kkoopa: This is to copy a (new-style) `v8::Persistent` without duplicating the storage cell, aka old-style copying. Need the actual `Persistent` handle to be able to access it from the weak callback, as I cannot reconstruct a Persistent handle in any way. Can't use a reference or pointer to the original hanlde, as it will go out of scope and be deallocated (not touching the underlying global storage cell) until the weak callback is called.

Is this safe? Should be. Is there any better way?

trevnorris: theoretically I believe so, but investigating more.

@bnoordhuis have any words of wisdom here?

kkoopa: To clarify, the situation is this:
```c++
{
  Isolate * isolate; /*...*/
  Local<Value> val; /*...*/
  Persistent<Value> p(isolate, val);
  int * i = new int(0);
  p.SetWeak(isolate, i, weak);
} // p goes out of scope and is destroyed, but its storage cell remains

void weak(...) {
  // dispose p
}
```
This would normally require having p globally accessible (directly or indirectly), to be able to reference it from the weak callback. I embed a copy of p in the data pointer supplied to the weak callback, so I may dispose it through that. Note that I do not want a copy made of the storage cell p refers to, just of the handle.

trevnorris: To verify you need to store the `v8::Persistent<T> *handle` that's passed to `NanWeakCallbackData()`? IIRC the pointer to a `v8::Persistent` will never change in its lifetime. So couldn't you do something like:
```cpp
NanWeakCallbackData(v8::Persistent<T>* handle, P* pram, Callback cb)
    : persistent(&handle),
      parameter(param),
      callback(cb) {
}
```
So then this would change to use a pointer:
```cpp
 private:  // data members
  v8::Persistent<T>* persistent;
```
You can then continue to do operations like the following:
```cpp
  bool IsNearDeath() const { return (*persistent).IsNearDeath(); }
```
Though I could also be completely missing something.

kkoopa: The Persistent<T>, the handle, is on the stack. It points to a T*, which is persistently stored. The handle is freed when it goes out of scope, but the T* remains and needs disposing later, in the weak callback. I can't dispose of the T* without a Persistent<T>, but I cannot construct a Persistent<T> from a T*, so the only way is by doing a bit copy of the original Persistent<T> to a heap-allocated struct.

trevnorris: > The Persistent<T>, the handle, is on the stack.

Oh duh.

> I cannot construct a Persistent<T> from a T*

Sorry. Not following this.

Back to the safety thing. I've done everything I could to it and seems to work fine. But you can thank @bnoordhuis for making me paranoid about doing this type of thing whenever a template class is involved.

bnoordhuis: Excuses le mot but memcpy'ing the persistent handle seems like a terrible idea, never mind that memcpy'ing non-trivial classes is nearly always a bad idea in general.  It's actively subverting how persistent handles are supposed to work and it could break on a whim.

kkoopa: But it is a trivial class. All it contains is a T *value_;

bnoordhuis: That's not what 'trivial' means in this context. :-)  See here: http://en.cppreference.com/w/cpp/concept/TriviallyCopyable

kkoopa: Meh, it just replicates the private Persistent constructor from T*. http://stackoverflow.com/a/3022238

bnoordhuis: The pertinent word that is missing in your last sentence is 'currently'. :-)  It may work not but even a trivial V8 upgrade could break it.

kkoopa: It could be adapted if that happens, but it has remained stable for a loong time. It's more likely to go away due to them redoing the weak handles again. They're already getting rid of them in favor of phantom persistents.

kkoopa:
Fixed placement of `*` and `&`.

nan.h
kkoopa: Will you lose sleep because of its hackish nature or is it acceptable for lack of better choices?
Maybe the V8 team could change V8 to dispose the persistent for you in the 4.5 WeakCallbackInfo-style weak callbacks? Having this copy of the Persistent handle around is only needed to dispose of it in the weak callback. Nothing else. If V8 took care of that little thing, there would be no need to copy the handle in the future.

kkoopa:
And added a comment statind that the `memcpy` thing assumes that `v8::Persistent<T>` exhibits bitwise copy semantics.

kkoopa:
Redid weak callback.

kkoopa:
Merged as 4a2ae84f84465b2a47aacbdb02c5c0b09b76dd52

PR #353, opened 2015-05-18
Fix incorrect, but working test

### LINKED ISSUE #352: [news test] NanNew<RegExp>("foo", kNone) is not /foo/g ?
Hi,
the "news" test fails here (nodejs 0.10.33 with old v8 3.14) and just returns /foo/ not /foo/g.
https://github.com/nodejs/nan/blob/master/test/cpp/news.cpp#L87
https://github.com/nodejs/nan/blob/master/test/js/news-test.js#L54

is this failing for you too ?

kkoopa:
Fixes #352?

### DIFF
--- a/test/js/news-test.js
+++ b/test/js/news-test.js
@@ -51,7 +51,7 @@ test('news', function (t) {
   t.equal(bindings.newLatin1String(), 'strïng');
   t.equal(bindings.newUcs2String(), 'strïng');
   t.equal(bindings.newStdString(), 'strïng');
-  t.deepEquals(bindings.newRegExp(), /foo/g);
+  t.equal(bindings.newRegExp().toString(), '/foo/');
   t.deepEquals(bindings.newStringObject(), new String("foo"));
   t.deepEquals(bindings.newNumberObject(), new Number(0.5));
   t.deepEquals(bindings.newBooleanObject(), new Boolean(true));
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
   "devDependencies": {
     "bindings": "~1.2.1",
     "node-gyp": "~1.0.2",
-    "pangyp": "~2.0.1",
-    "tap": "~0.7.1",
+    "pangyp": "~2.2.0",
+    "tap": "~1.0.7",
     "xtend": "~4.0.0"
   },
   "license": "MIT"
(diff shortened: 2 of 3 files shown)


bnoordhuis:
Not quite.  I think the issue is this:

```diff
diff --git a/test/js/news-test.js b/test/js/news-test.js
index 0de6e45..2a9c836 100644
--- a/test/js/news-test.js
+++ b/test/js/news-test.js
@@ -51,7 +51,7 @@ test('news', function (t) {
   t.equal(bindings.newLatin1String(), 'str<C3><AF>ng');
   t.equal(bindings.newUcs2String(), 'str<C3><AF>ng');
   t.equal(bindings.newStdString(), 'str<C3><AF>ng');
-  t.deepEquals(bindings.newRegExp(), /foo/g);
+  t.equals(bindings.newRegExp().toString(), '/foo/');
   t.deepEquals(bindings.newStringObject(), new String("foo"));
   t.deepEquals(bindings.newNumberObject(), new Number(0.5));
   t.deepEquals(bindings.newBooleanObject(), new Boolean(true));
```

kkoopa:
Aren't they the same, sort of? Depending on which is seen as authoritative, the C++, part or the js part.

kkoopa:
Ah, changing from deepEquals to equals too.... Why?

bnoordhuis:
Sorry, I should have mentioned that: I think the bug is that t.deepEquals doesn't seem to value-compare regular expressions.  Changing it to t.equal doesn't really matter, that was just a consistency thing.

By the way, I updated the diff; apparently t.equal and t.equals are both legal.

kkoopa:
How could the test ever fail then? Maybe with newer tap... Gonna check.

kkoopa:
Nope, but updating tap caught something else. I have made the changes.

test/js/news-test.js
```diff
@@ -51,7 +51,7 @@ test('news', function (t) {
   t.equal(bindings.newLatin1String(), 'strïng');
   t.equal(bindings.newUcs2String(), 'strïng');
   t.equal(bindings.newStdString(), 'strïng');
-  t.deepEquals(bindings.newRegExp(), /foo/g);
+  t.equal(bindings.newRegExp(), /foo/);
```
bnoordhuis: Does that actually work?  t.equal is identity equality, isn't it?

kkoopa: https://github.com/isaacs/node-tap#equalfound-wanted-message-extra

kkoopa:
There, Travis seems to like this.

bnoordhuis:
LGTM

kkoopa:
Thanks.

PR #355, opened 2015-05-23
Introduce NanNewOneByteString

kkoopa:
This has been on the list for a long time. Don't really care much for it myself, I think it becomes ugly, but if we are to support lesser compilers, then so be it.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -639,6 +639,12 @@ class NanEscapableScope {
 
 #if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                      \
   (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
+  NAN_INLINE NanMaybeLocal<v8::String>
+  NanNewOneByteString(const uint8_t * value, int length = -1) {
+    return v8::String::NewFromOneByte(v8::Isolate::GetCurrent(), value,
+          v8::NewStringType::kNormal, length);
+  }
+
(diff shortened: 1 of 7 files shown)


nan.h
```diff
@@ -909,6 +924,32 @@ class NanEscapableScope {
...
+  }
+  }  // end of namespace imp
+  }  // end of namespace Nan
+
+  NAN_INLINE NanMaybeLocal<v8::String>
+  NanNewOneByteString(const uint8_t * value, int length = -1) {
+    std::vector<uint16_t> wideString;
+    Nan::imp::widenString(&wideString, value, length);
+    if (wideString.size() == 0) {
+      return Factory<v8::String>::return_t(v8::String::Empty());
```
bnoordhuis: Just a note: you don't have to return String::Empty() yourself, V8 already does that.

nan.h
```diff
@@ -909,6 +924,32 @@ class NanEscapableScope {
     return function_template->HasInstance(value);
   }
 
+  namespace Nan { namespace imp {
+  NAN_INLINE void
+  widenString(std::vector<uint16_t> *ws, const uint8_t *s, int l = -1) {
```
bnoordhuis: Does `l` need a default value?

bnoordhuis:
LGTM

kkoopa:
Fixed. CI is not working currently because of 2600bd138d3cd1c1c01414ca663d44a2ebd9d053 and the fact that the new tap version has not been published to npm (it contains a showstopper bug fix for windows and the tag is in the git repository). I won't downgrade tap again, will just wait for the new version to get published.

PR #358, opened 2015-05-24
fix tap version

yorkie:
Fix the failing at [AppVeyor](https://ci.appveyor.com/project/RodVagg/nan/build/job/sa4i4frsxsqt4c5c), because the current tap version is [1.1.0](https://www.npmjs.com/package/tap), but we were using 1.1.2.

Related commit: 2600bd138d3cd1c1c01414ca663d44a2ebd9d053
R=@kkoopa

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "bindings": "~1.2.1",
     "node-gyp": "~1.0.2",
     "pangyp": "~2.2.0",
-    "tap": "~1.1.2",
+    "tap": "~1.1.0",
     "xtend": "~4.0.0"
   },
   "license": "MIT"


kkoopa:
Waiting for isaacs to publish 1.1.2 to npm. I mistakenly thought it was published already, as it was tagged in github. It contains fixes that made tap 1 unusable on Windows. 1.1.0 will not work on Windows. I decided not to go back down to 0.7, but wait for npm to get updated instead. Presumably everything will start working tomorrow.

kkoopa:
Oh yeah, 45d75f8d3853c9050383c54923c7b17012dc767c needs to be reapplied for tap 1 as well.

yorkie:
Got it, closing

PR #359, opened 2015-05-24
Rename Buffer Functions

kkoopa:
These fit the pattern better. Apparently there were never any tests of buffer functionality.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -285,6 +285,8 @@ template<typename P> class NanWeakCallbackInfo;
 #endif
 
 namespace Nan { namespace imp {
+  static const size_t kMaxLength = 0x3fffffff;
+
   template<typename T>
   NAN_INLINE
   v8::Local<T>
@@ -613,9 +615,14 @@ class NanEscapableScope {
   ) {
     // arbitrary buffer lengths requires
     // NODE_MODULE_VERSION >= IOJS_3_0_MODULE_VERSION
-    assert(length <= 0x3fffffff && "too large buffer");
+    assert(length <= Nan::imp::kMaxLength && "too large buffer");
+#if NODE_MODULE_VERSION > IOJS_2_0_MODULE_VERSION
+    return node::Buffer::New(
+        v8::Isolate::GetCurrent(), data, length, callback, hint);
+#else
     return NanMaybeLocal<v8::Object>(node::Buffer::New(
         v8::Isolate::GetCurrent(), data, length, callback, hint));
+#endif
   }
 
   NAN_INLINE NanMaybeLocal<v8::Object> NanCopyBuffer(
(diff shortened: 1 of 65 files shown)


bnoordhuis:
The CI doesn't seem particularly thrilled.  Is this a breaking change?

Apropos the `s/NanBufferUse/NanBufferNew/` change: `Buffer::Use()` and `Buffer::New()` do different things.  The former takes ownership of the pointer, the latter makes a copy (unless you use the `Buffer::New()` overload that takes a callback.)

kkoopa:
CI is broken because of a yet unpublished tap update.

There's a comment in the buffer source stating that it really should be called Buffer::New, figured it might as well be.

kkoopa:
@trevnorris Re: https://github.com/nodejs/io.js/blob/master/src/node_buffer.h#L63 Is this still the case?

kkoopa:
Actually, if we go with #356, we could introduce namespace Nan::Buffer to mirror node::Buffer, minus the isolates, or why not use a class?

trevnorris:
@kkoopa Neither of those are actually going to happen. Would cause too much breakage.

kkoopa:
Yes, but I'm willing to do that in Nan. Should it be done?

trevnorris:
It's just a name switch to make the API make more sense. So completely aesthetic.

kkoopa:
I'll take that as a 'yes'.

kkoopa:
Rebased

bnoordhuis:
LGTM.  I'm kind of sad there are no tests for the NanBuffer functions.

kkoopa:
I'll add some tests tomorrow and merge when CI is happy. It actually seems to error out on 0.8 right now: https://travis-ci.org/nodejs/nan/jobs/64168203

trevnorris:
@kkoopa a minor API change is that `node::Buffer::New()` used to abort if the length passed was `> 0x3fffffff`. Once Buffers are backed by Uint8Array you'll be able to pass an arbitrary size. Thing is, if `malloc()` returns `nullptr` then `Buffer::New()` will return an empty object. Indicating that the memory couldn't be allocated.

don't know if this will affect nan, but just in case wanted to give you a heads up.

kkoopa:
As we even support node 0.8, that limit will stay in NAN on all platforms to make it consistent. I guess that assert should be added to the NAN wrapper then.

trevnorris:
Sad that the new functionality will be missing, but understand where you're coming from.

kkoopa:
If someone wants at it, they can always add alternate code paths with ifdefs themselves.

kkoopa:
Added tests,

trevnorris:
@kkoopa Upon further implementation I realized that having the length be a `uint32_t` isn't sufficient. It will need to change to `size_t`. I'm unsure whether to change the existing API or create a new one that accepts `size_t`. Either way, wanted to let you know.

kkoopa:
You can safely change the existing API to use `size_t`, since it is bigger than `uint32_t` and has the same sign. Upcasting is done automatically. NAN will keep `uint32_t` for backwards compatibility. I'll add `isEmpty` checks.

kkoopa:
Where's [this test failure](https://travis-ci.org/nodejs/nan/jobs/64180954) in Buffer coming from?

kkoopa:
I guess it's because `Buffer::Use` tries to free the supplied buffer, which does not work, as it is static.

kkoopa:
@trevnorris, if there is a need to do an `IsEmpty` check, shouldn't `Buffer` be rewritten to return a `MaybeLocal<Object>`?

nan.h
```diff
@@ -605,31 +605,45 @@ class NanEscapableScope {
...
 
-  NAN_INLINE v8::Local<v8::Object> NanBufferUse(
+  NAN_INLINE v8::Local<v8::Object> NanNewBuffer(
       char* data
     , uint32_t size
   ) {
+    // arbitrary buffer lengths requires
+    // NODE_MODULE_VERSION >= IOJS_3_0_MODULE_VERSION
+    assert(size <= 0x3fffffff && "too large buffer");
     return node::Buffer::Use(v8::Isolate::GetCurrent(), data, size);
```
kkoopa: I'm assuming this one cannot return an empty handle, as it should not be calling malloc.

nan.h
```diff
@@ -1008,10 +1031,13 @@ class NanEscapableScope {
     delete[] data;
```
kkoopa: Is this correct? I though smalloc used `malloc`, which means that this should be `free(data)`.

trevnorris:
Yes. `Buffer::Use()` assumes it has control of the memory that's passed to it.

I'm not sure about returning a `MaybeLocal<>`. Going to get some feedback about that.

nan.h
trevnorris: You're correct.

kkoopa:
`Buffer` was updated to take `Isolate *` arguments when that change hit V8, this would be the natural thing to do for following in that line, especially since nobody will expect that a `Local` can be empty any more since V8 4.3, as all those APIs return a `MaybeLocal` in those cases. Not returning `MaybeLocal` from `Buffer` would thus be error-prone and unnatural.

trevnorris:
By "not sure" I meant that I wasn't sure what the other maintainers thought about the idea of changing the return value. After brief discussion it does seem returning `MaybeLocal` is the smart thing to do.

kkoopa:
Merged.

Issue #362, opened 2015-05-27
Can't set `target` for `NanCallback->Call`


xzyfer:
https://github.com/nodejs/nan/pull/275 introduced an optional `target` parameter for `NanCallback->Call`. This parameter (I believe) allows you to set the context for the callback i.e. this.

However not matter what I do the `this` value in my JavaScript callback is always the default `Global` context. I would expect the following to make foo available on this in my JavaScript callback.

```cpp
Local<Object> target = NanNew<Object>();
target->Set(NanNew<String>("foo"), NanNew<String>("bar"));
myNanCallback->Call(target, argv_v8.size(), &argv_v8[0]);
```

Is this a bug, or am I misunderstanding `target`?

bnoordhuis:
I looked at the code and I don't see anything that is obviously wrong.  I added a regression test in #363 and that's passing for me as well (with io.js at least.)

kkoopa:
Instead of passing `&argv_v8[0]`, you could just do `argv_v8`, but these are equivalent statements. Don't know why it does not work for you, tests pass here. Can you make a reproducible failing test?

xzyfer:
Looking into this now. For what it's worth the change you suggested causes an error

```
../src/callback_bridge.h:139:37: error: no matching member function for call to 'Call'
    NanNew<Value>(bridge->callback->Call(context, argv_v8.size(), argv_v8));
                  ~~~~~~~~~~~~~~~~~~^~~~
```

Using iojs v2.0.0.

kkoopa:
I was under the assumption that argv_v8 was an array of `Handle<Value>`.

xzyfer:
In our case `argv_v8` is a `std::vector<Handle<Value>>`.

xzyfer:
I'll try to boil down to a reduced test case. It's currently deep inside node-sass.

kkoopa:
@xzyfer What's the status on this?

xzyfer:
This is still an issue. I'm trying create a test case. 
I looked at the regression test you added. As far as I can tell you've simply asserting the code is valid, not that it produces the intended result. Am I missing something?

mathiask88:
I did a small test on this with iojs 2.1.0/NAN 1.8.4
```c++
#include <node.h>
#include <nan.h>

namespace test {

  NAN_METHOD(test) {
    NanCallback cb(args[0].As<v8::Function>());
    v8::Handle<v8::Value> argv[1];

[...]

PR #363, opened 2015-05-27
Add regression test for Call with custom receiver

bnoordhuis:
Refs #362.

R=@kkoopa

It passes for me with io.js, FWIW.


kkoopa:
LGTM

PR #364, opened 2015-05-27
Added version conversion script.

### LINKED ISSUE #360: Create conversion script for v1.8 to v2.0
Much of the breaking changes in 2.0 should be rather easy to fix automatically. The trickiest would be the introduction of `Maybe` types. These could be converted straight away by always doing `ToLocalChecked` or `FromJust`, but I don't think it is such a good idea to let people skip through fixing that properly. Maybe just detect them and print a message in the conversion report, that this needs to be converted manually, same could be added as a comment in the processed source code.

Simply using regexes without actually parsing the C++ code should work well enough in most cases, and since everybody use git, it will be easy to go through the diff and apply chunks in patch mode, ignoring erroneous replacements.

The other way would be to use libclang to actually parse the C++ code and do conversion there. This could be made robustly, but I don't think it is worth the effort compared to the former option.

TL;DR: 2to3.py

kkoopa:
Fixes #360. It would be nice if someone could take over this and make it an actual program.

### DIFF
--- a/tools/1to2.js
+++ b/tools/1to2.js
@@ -159,6 +159,12 @@ function replace(file, s) {
     return [p1, 'NanObjectWrap', p2].join('');
   });
 
+  s = s.replace(/(?:v8\:\:)?TryCatch/g, function() { return 'NanTryCatch'; });
+
+  s = s.replace(/(\w+)->StackTrace\(\)/g, function (match) {
+    return match + '.ToLocalChecked()';
+  });
+
   s = s.replace(/(\w+)->Equals\((\w+)/, function (match, p1, p2) {
     return ['NanEquals(', p1, ', ', p2].join('');
   });
@@ -284,8 +290,6 @@ function replace(file, s) {
   s = s.replace(/NODE_SET_METHOD/g, 'NanSetMethod');
 
   s = s.replace(/NODE_SET_PROTOTYPE_METHOD/g, 'NanSetPrototypeMethod');
-
-  s = s.replace(/(?:v8\:\:)?TryCatch/g, function() { return 'NanTryCatch'; });
 
   s = s.replace(/NanAsciiString/g, 'NanUtf8String');
 
--- a/tools/1to2.js
+++ b/tools/1to2.js
@@ -47,15 +47,19 @@ function replace(file, s) {
     return [p1, 'info', p2].join('');
   });
 
-  s = s.replace('NanScope()', 'NanScope scope');
+  s = s.replace(/NanScope\(\)/g, 'NanScope scope');
 
-  s = s.replace('NanEscapableScope()', 'NanEscapableScope scope');
+  s = s.replace(/NanEscapableScope\(\)/g, 'NanEscapableScope scope');
 
-  s = s.replace('NanReturnValue', 'info.GetReturnValue().Set');
+  s = s.replace(/NanReturnValue/g, 'info.GetReturnValue().Set');
 
-  s = s.replace('NanReturnUndefined()', 'return');
+  s = s.replace(/NanReturnUndefined\(\)/g, 'return');
 
-  s = s.replace('NanReturnNull', 'info.GetReturnValue().SetNull');
+  s = s.replace(/NanReturnNull/g, 'info.GetReturnValue().SetNull');
+
+  s = s.replace(/NanDisposePersistent\s*\(\s*(\w+)/g, function(match, p1) {
+    return p1 + '.Reset(';
+  });
 
   s = s.replace(/(?:v8\:\:)?Persistent/g, function() { return 'NanPersistent'; });
 
@@ -69,17 +73,17 @@ function replace(file, s) {
     return p1 + '.handle(';
   });
 
+  s = s.replace(/NODE_SET_METHOD/g, 'NanSetMethod');
+
+  s = s.replace(/NODE_SET_PROTOTYPE_METHOD/g, 'NanSetPrototypeMethod');
+
   s = s.replace(/(?:v8\:\:)?TryCatch/g, function() { return 'NanTryCatch'; });
 
-  s = s.replace('NanAsciiString', 'NanUtf8String');
+  s = s.replace(/NanAsciiString/g, 'NanUtf8String');
 
-  s = s.replace('NanUcs2String', 'v8::String::Value');
+  s = s.replace(/NanUcs2String/g, 'v8::String::Value');
 
-  s = s.replace(/NanDisposePersistent\s*\(\s*(\w+)/g, function(match, p1) {
-    return p1 + '.Dispose(';
-  });
-
-  s = s.replace('_NAN_', 'NAN_');
+  s = s.replace(/_NAN_/g, 'NAN_');
 
   for (i = 0; i < length; i++) {
     s = s.replace(removedregex[i], function (match, p1, p2) {
--- a/tools/1to2.js
+++ b/tools/1to2.js
@@ -36,30 +36,65 @@ var commander = require('commander'),
       'NAN_INDEX_DELETER',
       'NAN_INDEX_QUERY'],
     callbackregex = [],
+    toconverters = [
+      'Boolean',
+      'Number',
+      'String',
+      'Object',
+      'Integer',
+      'Uint32',
+      'Int32'],
+    toconvertersregex = [],
+    tovalues = [
+      ['bool', 'Boolean'],
+      ['double', 'Number'],
+      ['int64_t', 'Integer'],
+      ['uint32_t', 'Uint32'],
+      ['int32_t', 'Int32']],
+    tovaluesregex = [],
     length,
     i;
 
-commander
-  .version(JSON.parse(fs.readFileSync('package.json', 'utf8')).version)
-  .usage('[options] <file ...>')
-  .parse(process.argv);
+fs.readFile('package.json', 'utf8', function (err, data) {
+  if (err) {
+    throw err;
+  }
 
-if (!process.argv.slice(2).length) {
-  commander.outputHelp();
-}
+  commander
+      .version(JSON.parse(data).version)
+      .usage('[options] <file ...>')
+      .parse(process.argv);
+
+  if (!process.argv.slice(2).length) {
+    commander.outputHelp();
+  }
+});
+
 
 function replace(file, s) {
   var i, length;
 
   for (i = 0, length = removedregex.length; i < length; i++) {
     s = s.replace(removedregex[i], function (match) {
-    return '// ERROR: Rewrite using Buffer\n' +  match;
+      return '// ERROR: Rewrite using Buffer\n' +  match;
     });
   }
 
   for (i = 0, length = callbackregex.length; i < length; i++) {
     s = s.replace(callbackregex[i], function (match, p1) {
-    return p1;
+      return p1;
+    });
+  }
+
+  for (i = 0, length = toconverters.length; i < length; i++) {
+    s = s.replace(toconvertersregex[i], function (match, p1) {
+      return ['NanTo<', toconverters[i], '>(',  p1].join('');
+    });
+  }
+
+  for (i = 0, length = tovaluesregex.length; i < length; i++) {
+    s = s.replace(tovaluesregex[i], function (match, p1) {
+      return ['NanTo<', tovalues[i][0], '>(',  p1].join('');
     });
   }
 
@@ -158,8 +193,18 @@ for (i = 0, length = removed.length; i < length; i++) {
 }
 
 for (i = 0, length = callbacks.length; i < length; i++) {
-  callbackregex[i] =
-      new RegExp('(' + callbacks[i] + '\\(.*\\)\\s*\\{)\\s*NanScope\\(\\)\\s*;', 'g');
+  callbackregex[i] = new RegExp(
+      '(' + callbacks[i] + '\\(.*\\)\\s*\\{)\\s*NanScope\\(\\)\\s*;', 'g');
+}
+
+for (i = 0, length = toconverters.length; i < length; i++) {
+  toconvertersregex[i] =
+      new RegExp('(\\S+)->To' + toconverters[i] + '\\(', 'g');
+}
+
+for (i = 0, length = tovalues.length; i < length; i++) {
+  tovaluesregex[i] =
+      new RegExp('(\\S+)->' + tovalues[i][1] + 'Value\\(', 'g');
 }
 
 for (i = 2, length = process.argv.length; i < length; i++) {
--- a/tools/1to2.js
+++ b/tools/1to2.js
@@ -7,41 +7,51 @@
  ********************************************************************/
 
 var events = require('events'),
-    emitter = new events.EventEmitter(),
     fs = require('fs'),
     glob = require('glob'),
-    i, length, removedregex = [];
-
-const removed = [
-  'GetIndexedPropertiesExternalArrayDataLength',
-  'GetIndexedPropertiesExternalArrayData',
-  'GetIndexedPropertiesExternalArrayDataType',
-  'GetIndexedPropertiesPixelData',
-  'GetIndexedPropertiesPixelDataLength',
-  'HasIndexedPropertiesInExternalArrayData',
-  'HasIndexedPropertiesInPixelData',
-  'SetIndexedPropertiesToExternalArrayData',
-  'SetIndexedPropertiesToPixelData'];
-
-length = removed.length;
-
-for (i = 2; i < process.argv.length; i++) {
-  glob(process.argv[i], function (err, matches) {
-    if (err) {
-      throw err;
-    }
-    matches.forEach(function (currentValue) {
-      emitter.emit('file', currentValue);
-    });
-  });
-}
-
-for (i = 0; i < length; i++) {
-  removedregex[i] = new RegExp('^(.*)\\W' + removed[i] + '\\W(.*)', 'm');
-}
+    removed = [
+      'GetIndexedPropertiesExternalArrayDataLength',
+      'GetIndexedPropertiesExternalArrayData',
+      'GetIndexedPropertiesExternalArrayDataType',
+      'GetIndexedPropertiesPixelData',
+      'GetIndexedPropertiesPixelDataLength',
+      'HasIndexedPropertiesInExternalArrayData',
+      'HasIndexedPropertiesInPixelData',
+      'SetIndexedPropertiesToExternalArrayData',
+      'SetIndexedPropertiesToPixelData'],
+    removedregex = [],
+    callbacks = [
+      'NAN_METHOD',
+      'NAN_GETTER',
+      'NAN_SETTER',
+      'NAN_PROPERTY_GETTER',
+      'NAN_PROPERTY_SETTER',
+      'NAN_PROPERTY_ENUMERATOR',
+      'NAN_PROPERTY_DELETER',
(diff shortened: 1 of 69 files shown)


tools/1to2.js
```diff
@@ -0,0 +1,107 @@
...
+    return [p1, 'info', p2].join('');
+  });
+
+  s = s.replace('NanScope()', 'NanScope scope');
+
+  s = s.replace('NanEscapableScope()', 'NanEscapableScope scope');
+
+  s = s.replace('NanReturnValue', 'info.GetReturnValue().Set');
+
+  s = s.replace('NanReturnUndefined', 'return');
```
mathiask88: Did you miss the parentheses here?

kkoopa: Yes.

tools/1to2.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+for (i = 0; i < length; i++) {
+  removedregex[i] = new RegExp('^(.*)\\W' + removed[i] + '\\W(.*)', 'm');
+}
+
+function replace(file, s) {
+  var i;
+
+  s = s.replace(/(\W)args(\W)/, function (match, p1, p2) {
```
mathiask88: What about global flags for the RegExps?

kkoopa: Probably good.

tools/1to2.js
```diff
@@ -0,0 +1,107 @@
...
+      throw err;
+    }
+    matches.forEach(function (currentValue) {
+      emitter.emit('file', currentValue);
+    });
+  });
+}
+
+for (i = 0; i < length; i++) {
+  removedregex[i] = new RegExp('^(.*)\\W' + removed[i] + '\\W(.*)', 'm');
```
mathiask88: Still one left ;)

kkoopa: Left this one intentionally, it's in multiline mode already, no need to put the same warning many times on a line.

mathiask88:
What do you think about a replacement for `NODE_SET_(PROTOTYPE_)METHOD` with `NanSet(Prototype)Method`?

tools/1to2.js
```diff
@@ -0,0 +1,107 @@
...
+
+  s = s.replace('NanEscapableScope()', 'NanEscapableScope scope');
+
+  s = s.replace('NanReturnValue', 'info.GetReturnValue().Set');
+
+  s = s.replace('NanReturnUndefined()', 'return');
+
+  s = s.replace('NanReturnNull', 'info.GetReturnValue().SetNull');
+
+  s = s.replace(/(?:v8\:\:)?Persistent/g, function() { return 'NanPersistent'; });
```
mathiask88: This one is not correct. `NanDisposePersistent` will end up as `NanDisposeNanPersistent`

tools/1to2.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function replace(file, s) {
+  var i;
+
+  s = s.replace(/(\W)args(\W)/g, function (match, p1, p2) {
+    return [p1, 'info', p2].join('');
+  });
+
+  s = s.replace('NanScope()', 'NanScope scope');
```
mathiask88: I think it is better to write this as regexp `/.../g´ also, because this will only replace the first occurrence.

kkoopa: All of the string replacements would be better of as regular expressions, with whitespace hardening and whatnot.

kkoopa:
Yes, it should do the SET_PROTOYPE_METHOD replacement. I had forgotten about that one. For that matter, io.js should update those functions to make them work.

mathiask88:
I added some regexps, what do you think about that? And I changed the `Dispose` to `Reset`. If I convert my addon with the gist file there are only some Maybes left. When does NanNew return Maybe and when Local? Seems a bit random to me :S

tools/1to2.js
kkoopa: It merely needed moving so it runs after the `NanDisposePersistent` replacement.

kkoopa:
from the top of my head, String, Regex and Date return maybes. NanEmptyString() gives a Local<String> directly.

kkoopa:
New regexes look mostly good, but I am not too keen on doing `toLocalChecked` automatically. Possibly behind a flag (would need to add command line options parsing then too).

mathiask88:
If I'm honest I think I didn't understand the maybe types at all. If I want to create a local string from a const string like `NanNew<v8::String>("V3")` how can this raise an exception so that a Maybe type is returned instead of a Local?

kkoopa:
Only way I could think of is if the input char array is not valid for the encoding, or more likely if it exceeds maximum allowable size. In practice it should never fail, but there is no other way.

mathiask88:
That was why I added the auto toLocalCheck() regexp for const strings, but it is so ugly to read because lines will get very long very fast `!data_obj->Get(NanNew<v8::String>("Area").ToLocalChecked())->IsInt32()`. Is it too much overhead that Nan checks const strings and returns Local instead of Maybe?

kkoopa:
It wouldn't be too much overhead, but it cannot determine when it's a string literal and when it's a pointer to const char data.

mathiask88:
What do you think of something like this 
```javascript
  s = s.replace(/(NAN_METHOD\(.*\)[\s\t\n\r]*\{)[\s\n\r\t]+NanScope\(\);/g, function(match, p1){
    return p1;
  })
```

And in my gist there were also regexes for `NanAssignPersistent` and `NanEscapeScope`, and changes for the `Persistent` one so that it does not match any occurrence inside a word.

BTW what do you mean by "make it an actual program"? Works like a charm already ;)

kkoopa:
The algorithm is horribly innefficient. There are no options, no help, no warnings, etc. Just a crude script cobbled together in half an hour. And I don't really do JavaScript.

Shouldn't all ws be included in \s already?

mathiask88:
Yes you are right `\s*` is enough. I don't think a conversion script needs to be that efficient. +1 for options help and warnings. What is the ETA for this?

kkoopa:
NAN 2.0 release (which will happen once io.js 3.0 is released and [our documentation has been upgraded](https://github.com/nodejs/nan/pull/334#issuecomment-105324117)). As described in #360, the idea of this is not to be a run-and-publish panacea, but to ease the burden of a person doing the porting.

It is true that it does not have to be efficient, but it hurts my sense of perfectionism and causes me to lose sleep at night.

I really hate writing documentation, so any help is welcome there. The less documentation I have to write, the faster it will be done. I know it would be preferrable for whoever wrote the code to also document it, but I don't have time or motivation to write all the documentation in addition to all the code.

I like the regex removing unnecessary `NanScope`s, actually thought of doing something like that myself, but wasn't sure if it would be reliable enough, e.g. say you had some useless logging statements or variable declararations before the `NanScope` (which is valid, but uncommon). I'll add it and update it to also handle `NAN_GETTER`, `NAN_SETTER` and all the other callbacks.

`NAN_WEAK_CALLBACK` could also be automatically replaced.

mathiask88:
Ok, I will try to do some help here. I understand your perfectionism because I am a perfectionist as well, however programming is not my profession but rather a hobby. Let's try to save your sleep :)

Got some errors in the current regexps
```javascript
  s = s.replace(/(?:node\:\:)?ObjectWrap/g, function() { return 'NanObjectWrap'; });

  s = s.replace(/NanObjectWrapHandle\s*\((\w+)/g, function (match, p1) {
    return p1 + '.handle(';
  });
```
`ObjectWrap` is the same like `Persistent` it replaces every occurrence in a word e.g `NanObjectWrapHandle -> NanNanObjectWrap` and `NanObjectWrapHandle` takes a pointer so it should be `return p1 + '->handle(';` I think.

I think if we use external dependencies anyway, we could use the `commander` npm package to get option parsing and help into the script.

kkoopa:
Thanks. I've updated the script.

kkoopa:
And added the automated `ToLocalChecked` conversion you propsed, as well as for `RegExp` and `Date`. Note that single quotes do not denote strings in C++.

Still missing are all the converters (`NanTo<T>`) from 72e200e7e3a6fafc956fbf7790878aa320003001 as well as all its added helpers.

kkoopa:
Added that missing functionality. Can't come up with anything more right now.

kkoopa:
Merged

mathiask88:
Oh, cool. There are still bugs in some regexes. I will try to correct them.

kkoopa:
Sure, send a PR against the next branch.

kkoopa:
I rewrote it and fixed some bugs.

mathiask88:
Nice, I like it, but applying the master regex only once will introduce some bugs.
Example:
`Database* database = node::ObjectWrap::Unwrap<Database>(args[0]->ToObject());`
For this line the regex will match the `ObjectWrap` and the `ToObject`, but not the `args`.

kkoopa:
I'll see if that can be fixed by making one of them less greedy, otherwise it cannot be easily solved. There are insurmountable problems in using regular expressions for this, so at some point you have to give up.

kkoopa:
Nope, that can't fix it. It can be remedied by doing n passes, but that's not really nice.

mathiask88:
Yep, already tried this (https://gist.github.com/mathiask88/066557dd334f86531c37). But the regexes need to be stricter, that they only match the non-converted string. But as you said this is not really nice. The version from yesterday did the job better but slower.

kkoopa:
I fixed this with two passes in the latest update.

mathiask88:
Yes, my solution iterates until everything is done. But as I said the regexes need to be stricter, because now you got things like `NanNanTryCatch try_catch;`. And there are still some bad regexes like
```
    if (!array->Get(i)->IsObject())
    if NanGet((!array, i)->IsObject())
```

I will try to find more, so that the script runs as good as possible.

kkoopa:
javascript's regex engine lacks backrefs, but you can reverse the string to work around that, e.g. 'TryCatch', not preceded by 'Nan'  would be 'hctaCyrT(?!naN)'

kkoopa:
But I found a better way that should work in this case.

kkoopa:
I started [playing with libclang](https://github.com/kkoopa/node-libclang/blob/6ecce3ede3f61dcc792e666b09ca55d3c556474c/analyzer.js) a bit. This is what a proper conversion tool would require. I think it's probably too much work for little gain. This toy example proposes correct substitutions for '->Int32Value' (and could easily do all those other methods from value with no arguments that fit the pattern).

kkoopa:
[Added a bit more](https://github.com/kkoopa/node-libclang/blob/27142dc759733e1dd9f7a8fdc5009ebbd31df2d1/analyzer.js).
```sh
target->Set(NanNew("Context"), t->GetFunction())
NanSet(target, NanNew("Context"), t->GetFunction())
-----------------------------------------------------------------------------------------------
t->GetFunction()
NanGetFunction(t)
-----------------------------------------------------------------------------------------------
args[0]->ToInteger()
NanTo<v8::Integer>(args[0])
-----------------------------------------------------------------------------------------------
NanReturnValue(args.This())
args[0]->Int32Value()
NanTo<int32_t>(args[0])
-----------------------------------------------------------------------------------------------
args[1]->Int32Value()
NanTo<int32_t>(args[1])
-----------------------------------------------------------------------------------------------
args[0].As<v8::Object>()->Get(0)->Int32Value()
NanTo<int32_t>(args[0].As<v8::Object>()->Get(0))
-----------------------------------------------------------------------------------------------
args[0].As<v8::Object>()->Get(0)
NanGet(args[0].As<v8::Object>(), 0)
-----------------------------------------------------------------------------------------------
args[0]->Int32Value()
NanTo<int32_t>(args[0])
-----------------------------------------------------------------------------------------------
NanReturnValue(NanNew<Integer>(value))
t->InstanceTemplate()->SetAccessor(
      NanNew("state"), Socket::GetState)
NanSetAccessor(t->InstanceTemplate(),
      NanNew("state"), Socket::GetState)
-----------------------------------------------------------------------------------------------
t->InstanceTemplate()->SetAccessor(
      NanNew("pending"), GetPending, SetPending)
NanSetAccessor(t->InstanceTemplate(),
      NanNew("pending"), GetPending, SetPending)
-----------------------------------------------------------------------------------------------
target->Set(NanNew("SocketBinding"), t->GetFunction())
NanSet(target, NanNew("SocketBinding"), t->GetFunction())
-----------------------------------------------------------------------------------------------
t->GetFunction()
NanGetFunction(t)
-----------------------------------------------------------------------------------------------
args[0]->ToObject()
NanTo<v8::Object>(args[0])
-----------------------------------------------------------------------------------------------
args[1]->ToInteger()
NanTo<v8::Integer>(args[1])
-----------------------------------------------------------------------------------------------
NanReturnValue(args.This())
NanObjectWrapHandle(this)->Get(NanNew(callback_symbol))
NanGet(NanObjectWrapHandle(this), NanNew(callback_symbol))
-----------------------------------------------------------------------------------------------
NanObjectWrapHandle(this)->Get(NanNew(monitor_symbol))
NanGet(NanObjectWrapHandle(this), NanNew(monitor_symbol))
-----------------------------------------------------------------------------------------------
NanReturnValue(NanNew<Integer>(socket->state_))
NanReturnValue(NanNew<Integer>(socket->pending_))
value->Int32Value()
NanTo<int32_t>(value)
-----------------------------------------------------------------------------------------------
wrappedValue->ToInteger()
NanTo<v8::Integer>(wrappedValue)
-----------------------------------------------------------------------------------------------
wrappedValue->ToObject()
NanTo<v8::Object>(wrappedValue)
-----------------------------------------------------------------------------------------------
args[0]->ToInteger()
NanTo<v8::Integer>(args[0])
-----------------------------------------------------------------------------------------------
NanReturnValue(socket->GetSockOpt<int>(option))
NanReturnValue(socket->GetSockOpt<uint32_t>(option))
NanReturnValue(socket->GetSockOpt<int64_t>(option))
NanReturnValue(socket->GetSockOpt<uint64_t>(option))
NanReturnValue(socket->GetSockOpt<char*>(option))
args[0]->ToInteger()
NanTo<v8::Integer>(args[0])
-----------------------------------------------------------------------------------------------
NanReturnValue(socket->SetSockOpt<int>(option, args[1]))
NanReturnValue(socket->SetSockOpt<uint32_t>(option, args[1]))
NanReturnValue(socket->SetSockOpt<int64_t>(option, args[1]))
NanReturnValue(socket->SetSockOpt<uint64_t>(option, args[1]))
NanReturnValue(socket->SetSockOpt<char*>(option, args[1]))
args[0]->ToInteger()
NanTo<v8::Integer>(args[0])
-----------------------------------------------------------------------------------------------
args[0]->ToInteger()
NanTo<v8::Integer>(args[0])
-----------------------------------------------------------------------------------------------
NanReturnValue(msg.GetBuffer())
args[1]->ToInteger()
NanTo<v8::Integer>(args[1])
-----------------------------------------------------------------------------------------------
NanReturnValue(NanNew<String>(version_info))
obj->Set(NanNew<String>("public"), NanNew<String>(public_key))
NanSet(obj, NanNew<String>("public"), NanNew<String>(public_key))
-----------------------------------------------------------------------------------------------
obj->Set(NanNew<String>("secret"), NanNew<String>(secret_key))
NanSet(obj, NanNew<String>("secret"), NanNew<String>(secret_key))
-----------------------------------------------------------------------------------------------
NanReturnValue(obj)
NODE_DEFINE_CONSTANT(target, ZMQ_CAN_DISCONNECT)
NODE_DEFINE_CONSTANT(target, ZMQ_CAN_UNBIND)
NODE_DEFINE_CONSTANT(target, ZMQ_CAN_MONITOR)
NODE_DEFINE_CONSTANT(target, ZMQ_CAN_SET_CTX)
NODE_DEFINE_CONSTANT(target, ZMQ_PUB)
NODE_DEFINE_CONSTANT(target, ZMQ_SUB)
NODE_DEFINE_CONSTANT(target, ZMQ_XPUB)
NODE_DEFINE_CONSTANT(target, ZMQ_XSUB)
NODE_DEFINE_CONSTANT(target, ZMQ_REQ)
NODE_DEFINE_CONSTANT(target, ZMQ_XREQ)
NODE_DEFINE_CONSTANT(target, ZMQ_REP)
NODE_DEFINE_CONSTANT(target, ZMQ_XREP)
NODE_DEFINE_CONSTANT(target, ZMQ_DEALER)
NODE_DEFINE_CONSTANT(target, ZMQ_ROUTER)
NODE_DEFINE_CONSTANT(target, ZMQ_PUSH)
NODE_DEFINE_CONSTANT(target, ZMQ_PULL)
NODE_DEFINE_CONSTANT(target, ZMQ_PAIR)
NODE_DEFINE_CONSTANT(target, ZMQ_STREAM)
NODE_DEFINE_CONSTANT(target, ZMQ_POLLIN)
NODE_DEFINE_CONSTANT(target, ZMQ_POLLOUT)
NODE_DEFINE_CONSTANT(target, ZMQ_POLLERR)
NODE_DEFINE_CONSTANT(target, ZMQ_SNDMORE)
NODE_DEFINE_CONSTANT(target, STATE_READY)
NODE_DEFINE_CONSTANT(target, STATE_BUSY)
NODE_DEFINE_CONSTANT(target, STATE_CLOSED)
wrappedValue->ToInteger()
NanTo<v8::Integer>(wrappedValue)
-----------------------------------------------------------------------------------------------
wrappedValue->ToInteger()
NanTo<v8::Integer>(wrappedValue)
-----------------------------------------------------------------------------------------------
wrappedValue->ToInteger()
NanTo<v8::Integer>(wrappedValue)
-----------------------------------------------------------------------------------------------
wrappedValue->ToInteger()
NanTo<v8::Integer>(wrappedValue)
-----------------------------------------------------------------------------------------------
```

mathiask88:
Cool, no false positives anymore. But this needs also some iterations per file to replace nested patterns?

kkoopa:
Not necessarily, but it will be very demanding and slow still.

mathiask88:
Yes still slow but I think good results are more important. No one will be peeved if it takes 0.5s more. And we can still split the jobs to all CPU cores. I got 12 cores(incl HT) and they are soo bored all the time ;)

mathiask88:
Some bad news related to windows user. Installing clang takes 280MB, iojs/node version must be x86 because prebuild clang is only x86 and the function `clang_visitChildrenWithBlock`(https://github.com/kkoopa/node-libclang/blob/master/lib/dynamic_clang.js#L1072) does not exist in the windows dll.

kkoopa:
Maybe that function can be avoided, if not, then it's too bad for Windows. They can always install a proper OS for development.

mathiask88:
Yes the function can be commented out in node-libclang. And yes they can do, but I think we should also stick to fringe groups ;) If we stick to the clang solution, the dependency chain is quiet big. What do you think of including everything needed into a atom-shell ready to run GUI app. Nothing complicated, just a window with folder/file selection and a "Go"-Button. But we can include all the right dependencies especially for windows libclang.dll, patched node-libclang and x86 iojs.

kkoopa:
If you feel like writing one, sure. What would the GUI do? I only intended to make it refactor a set of files.

mathiask88:
Not more than refactor a set of files. But we can bundle all dependencies ready to go, have an easy file selection and maybe some help text. I will look into it.

kkoopa:
Good. I'll set up a separate project for the refactoring tool when I get around to it. Does not make sense to bundle with NAN.

mathiask88:
Do we want the user to select a binding.gyp to convert a project or individual files or both?

kkoopa:
Dunno. binding.gyp would probably be best, but may be much work.

Issue #365, opened 2015-05-28
Release proposal: 1.9


kkoopa:
Let's get this out before io.js 3.0 is out, so the 1-series can be laid to rest. There will be no more features added to NAN 1 ever.

Only sort of open issue is https://github.com/nodejs/nan/issues/362, but I believe it might be erroneous somehow since all our tests pass and the code is rather straightforward. If it is real it can be fixed in a patch release.

rvagg:
:thumbsup: what's the delta between 1.8.4 and this?

kkoopa:
Insignificant, IIRC. A couple deprecations, some small additions. The deprecations could even be undone. So much will change that a migratory path through deprecations won't work. There would be hardly any non-deprecated functionality left.

Basically just to tie up a few loose ends and say goodbye to NAN 1.

kkoopa:
To clarify, here are the (significant) commits:

Added `NanFatalException` 81d4a2c0e04fab2ec4a48bdd776b722077de94be

Introduce and use macros for disallowing assign, copy and move 3357130aa844ec795471743c025d0dd399bbf855

Prohibit moving non-copyable classes fba4a299d18ce2f81f76f49ae41e3e43ab875c21

Disallow assign, copy and move in `_NanWeakCallbackInfo` c96d9b206103f4122059c99530ecac2872d2a56c

Disallow assign, copy and move in `NanCallback` 4b808cbfba50322202e8f976fa6d494e00f81714

Disallow assign, copy and move in `NanAsyncWorker` 1f55c595d817ee9aa204fb6327fbc7c875b5ef9e

Allow `NanReturnValue(Persistent<T>)` 915f1553281ec46a47f5782e074ad5217eafe68d
In NAN 2.0, the equivalent will only work for `Global`s (previously known as `UniquePersistent`s) due to upcoming changes in V8 4.5.

Deprecate `NanGetPointerSafe` and `NanSetPointerSafe` 3e60290a31d7d650b9f46a846da5d42be98b7cb8

Deprecate `NanBooleanOptionValue` and `NanUInt32OptionValue` 0ad254b48668728f9cd738c377ae609bf2808f02

Added more error types and made them all accept `Handle<String>` as arguments 4265f06c880f9c166716c45b0aed05a90452f49d
Needs doc update to list new error types.

Additionally, these could be backported from 'next':

[...]

Issue #284, opened 2015-02-18
MakeCallback is dangerous, not only or best way of calling callback; provide direct method


metamatt:
NodeJS lets you invoke a "callback" (really, invoke JS code from C++ code) 2 ways:

* calling the callback directly; if you have a `Handle<Function> foo` you can just `foo->Call()`
* calling the callback via `MakeCallback` (in src/node.cc)

The former ("direct") case just calls from C++ into JS synchronously, and returns back into C++ when the JS code returns.

The latter (`MakeCallback`) case calls the JS callback function, but then also ends the current tick, calling `process._tickCallback()` and any *other* callbacks registered via `nextTick()`. This form is intended only for use by C++ code that was dispatched from the libuv event loop, and not for C++ code that was invoked from JS code. Because in that case (JS code "A" running, calls into C++ code in a node addon, the addon's C++ code calls MakeCallback, MakeCallback dispatches the nextTick callbacks) the JS code marked "A" has been preempted, and state changed by the nextTick callbacks can be corrupted. See https://github.com/TooTallNate/node-weak/issues/35 for an example of how dangerous this is.

Use cases for node addons (that would want to use `nan`) wanting to invoke JS code from C++ code that was invoked directly from JS (not from a libuv event callback) include

* node-weak (again, https://github.com/TooTallNate/node-weak/issues/35 is a practical example of this)
* if you want to use V8 interceptors to provide Proxy-like abilities for a dynamic object to trap accesses and call JS handler routines

For these cases, use of `MakeCallback` is not only incorrect but extremely dangerous. The right answer is to call the callback directly. But `nan` doesn't expose a way of doing this; `NanCallback::Call` is actually wrapping `MakeCallback`. So is the right answer to avoid `NanCallback` altogether? I don't think so; it serves a very real need to deal more easily with V8's memory management; using `Persistent<Function>` got harder in newer V8 (node 0.11/0.12) and so it's much nicer to use `NanCallback` instead of raw `Handle<Function>`. So I think `NanCallback` should provide separate methods for calling a callback directly and calling a callback via `MakeCallback`, and make the difference really obvious so people don't accidentally do the latter. (Note in my [patch](https://github.com/TooTallNate/node-weak/pull/36) for the [node-weak instance of this](https://github.com/TooTallNate/node-weak/issues/35), I do use `NanCallback` for memory management, but then have to go to some lengths to extract the real `Handle<Function>` to call directly.)

[...]

PR #370, opened 2015-05-30
Add dereference and function call operators to NanCallback

kkoopa:
As a way of addressing #284, I added a dereference operator to `NanCallback` as a shorthand for `GetFunction`. I also added a function call operator as a shorthand for `Call`.

Should be included in #365.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1601,6 +1601,22 @@ class NanCallback {
     return !this->operator==(other);
   }
 
+  NAN_INLINE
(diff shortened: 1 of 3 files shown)


bnoordhuis:
LGTM, I suppose, although I question how useful that is.

kkoopa:
It's more natural. You would expect `NanCallback` to be a functor. Makes it easily usable in more code.

PR #372, opened 2015-06-08
Add NanErrnoException

sam-github:
PR is modelled on 81d4a2c0e04, which had neither docs nor tests, so maybe its too small to need them?

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -856,6 +856,15 @@ return args.GetReturnValue().Set(Nan::imp::NanEnsureHandleOrPersistent(value))
     node::FatalException(v8::Isolate::GetCurrent(), try_catch);
   }
 
+  NAN_INLINE v8::Local<v8::Value> NanErrnoException(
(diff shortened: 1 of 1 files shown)


kkoopa:
Thanks. Looks good. Docs were going to be added at release (essentially just saying that it is equivalent to the node function). I can add that documentation myself.

Tests are nice, but not strictly necessary for such a simple construct as a straight wrapper, although good to have in case of regressions. Here, the test is mostly that it still compiles on all platforms, which it seems to do.

sam-github:
Great, thanks @kkoopa, any chance of getting this into 1.9? I'm trying to get node-fs-ext to not spew warnings during installation, they make people nervous.

kkoopa:
Sure, I'll include it.

kkoopa:
Committed in dd87d9e9280cdf27488377b7d6a8cc1d6bd1a34d

PR #383, opened 2015-07-21
Bump dependency on node-gyp to 2.0.2 to pick up https:// downloading of node.js source

reedloden:
`node-gyp` had been using http://nodejs.org for downloading node.js source. This was recently changed in https://github.com/TooTallNate/node-gyp/pull/656 to use https:// instead. Bump dependency to pick up that change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   ],
   "devDependencies": {
     "bindings": "~1.2.1",
-    "node-gyp": "~1.0.2",
+    "node-gyp": "~2.0.2",
     "pangyp": "~2.2.0",
     "tap": "~0.7.1",
     "xtend": "~4.0.0"


reedloden:
Need this changed in `nan` so I can ask other modules to then bump their dependency on `nan`. The fun ol' dependency fixing game.

rvagg:
lgtm, but this is a devDependency so shouldn't be impacting downstream dependencies of NAN

reedloden:
Specifically, the following modules depend on `nan`, and by that association, they get `node-gyp` to use for themselves during building: `contextify`, `node-sass`, and `fsevents` (using these as examples, as they are ones specifically causing problems for me).

.. or am I missing something, and they get `node-gyp` from somewhere else? I just don't see how (I'm not a node guy, so maybe I'm missing something obvious).

rvagg:
node-gyp comes from your installation of npm normally, it's a dependency of npm. `npm install npm@latest -g` to upgrade and get the latest

reedloden:
Oh, `npm` itself depends on `node-gyp`. Perhaps that's where it comes from...

https://github.com/npm/npm/blob/be379977d83c7cc5d961d25c7b883b1ade081a65/package.json#L64 -- looks like somebody just bumped the rev of `node-gyp` there, too.

reedloden:
Ah, yes. Indeed. Ok, thanks. Might be still a good idea to update `nan`'s copy, but seems I can actually fix my real problem by just ensuring `npm` is updated everywhere.

PR #392, opened 2015-08-02
Reimplement ObjectWrap from scratch to prevent memory leaks.

### LINKED ISSUE #389: Is nan v2.0.0 leaking memory?
I'm in the process of migrating the [epoll](https://github.com/fivdi/epoll) module from nan v1.8.4 to nan v2.0.0. Everything is looking ok except for memory usage. There's an epoll test called [brute-force-leak-check.js](https://github.com/fivdi/epoll/blob/master/test/brute-force-leak-check.js). This test is used to create millions of objects and while it's running the memory usage is observed to see if everything looks ok. With `Node.js v0.10.40` and `nan v1.8.4` memory usage is extremely stable and everything looks good. With `Node.js v0.10.40` and `nan v2.0.0` memory usage keeps increasing and eventually the test will terminate with:

```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)
```

I can't figure out where the problem is and wrote the following test to see if the problem is in the `epoll` code or perhaps somewhere else:

```
const testRoot = require('path').resolve(__dirname, '..')
  , bindings = require('bindings')({ module_root: testRoot, bindings: 'objectwraphandle' });

var i, obj;

for (i = 1; i <= 1e9; i += 1) {
  obj = new bindings.MyObject(i);
  if (i % 1e6 === 0) {
    console.log(i);
  }
}
```

This test is similar to the `epoll` `brute-force-leak-check` but creates millions of `nan` `MyObject` instances instead ([objectwraphandle.cpp](https://github.com/nodejs/nan/blob/master/test/cpp/objectwraphandle.cpp)).

If this test is placed in the `nan` `test\js` folder and run from there, the same issue can be seen. With `Node.js v0.10.40` and `nan v1.8.4` memory usage is extremely stable and everything looks good. With `Node.js v0.10.40` and `nan v2.0.0` memory usage keeps increasing and eventually the test will terminate after throwing an instance of `std::bad_alloc`.

Unfortunately, it may take a long time for `std::bad_alloc` to be thrown, even on a machine with only 2GB of RAM, but the constant increase in memory usage is immediately visible. This is the followed by swapping to the hard-disk after a relatively short period of time.
 
Additional information:
```
$ node -v
v0.10.40
$ uname -a
Linux linuxbox 3.13.0-58-generic #97-Ubuntu SMP Wed Jul 8 02:56:50 UTC 2015 i686 i686 i686 GNU/Linux
$ 
```

Is the test a little too stupid/synthetic or is there an issue somewhere?

kkoopa:
Part of #389.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ LINT_SOURCES = \
 	nan_maybe_43_inl.h \
 	nan_maybe_pre_43_inl.h \
 	nan_new.h \
+	nan_object_wrap.h \
 	nan_persistent_12_inl.h \
 	nan_persistent_pre_12_inl.h \
 	nan_string_bytes.h \
--- a/doc/node_misc.md
+++ b/doc/node_misc.md
@@ -34,15 +34,49 @@ v8::Local<v8::Value> Nan::MakeCallback(v8::Local<v8::Object> target,
 <a name="api_nan_object_wrap"></a>
 ### Nan::ObjectWrap()
 
-A simple extension of `node::ObjectWrap` that adds some API not present in older versions of Node. Should be preferred over `node::ObjectWrap` in all cases for consistency.
+A reimplementation of `node::ObjectWrap` that adds some API not present in older versions of Node. Should be preferred over `node::ObjectWrap` in all cases for consistency.
+
+See the Node documentation on [Wrapping C++ Objects](https://nodejs.org/api/addons.html#addons_wrapping_c_objects) for more details.
 
 Definition:
 
 ```c++
-class ObjectWrap : public node::ObjectWrap {
+class ObjectWrap {
  public:
-  v8::Local<v8::Object> handle(); // for older Node where only handle_ is provided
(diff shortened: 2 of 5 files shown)


kkoopa:
I can't figure out why returnvalue keeps failing consistently on 0.10.40 at AppVeyor.

mathiask88:
@kkoopa Can't reproduce (Win10/VS15). Some strange appveyor behavior.

kkoopa:
I can't reproduce it myself either.

kkoopa:
@bnoordhuis Can you review this? Something is still leaking somewhere, but I can't see why it would be this.

bnoordhuis:
I'll (try to) take a look today at this and #389.

kkoopa:
I'll include this in 2.0.1, as it resolves part of the problem. Just have to update the documentation of ObjectWrap.

PR #397, opened 2015-08-05
add overload for Nan::NewInstance that takes argc/argv

demmer:
This adds another overloaded variant of `Nan::NewInstance` that supports passing argc/argv through to the invoked Function.

### DIFF
--- a/doc/maybe_types.md
+++ b/doc/maybe_types.md
@@ -159,6 +159,7 @@ Signature:
 
 ```c++
 Nan::MaybeLocal<v8::Object> Nan::NewInstance(v8::Local<v8::Function> h);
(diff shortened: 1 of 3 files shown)


kkoopa:
Oh, didn't notice this was left out. Good catch. Can you update the signatures to take `v8::Local<v8::Value>` instead of `v8::Handle<v8::Value>`? We've moved away from `Handle` since it is being phased out and will be completely removed from V8 in latter versions.

demmer:
Sure -- done.

kkoopa:
Should this count as a bugfix or a feature addition? I'm willing to go with the former and include this is 2.0.1. It was intended to be included.

rvagg:
I don't think it matters _too much_ but it's technically a feature addition and a 2.1.0 shouldn't be scary, just bump and run

kkoopa:
Merged 29450ed1670051613ffea9dd3f980f8685feceb4

Issue #396, opened 2015-08-05
nan 2.0.0 breaks in node@0.10.26


Raynos:
```
In file included from ../src/sse4_crc32.cpp:18:
../node_modules/nan/nan.h:1178:39: error: no member named 'REPLACE_INVALID_UTF8' in 'v8::String'
                          v8::String::REPLACE_INVALID_UTF8;
                          ~~~~~~~~~~~~^
1 error generated.
make: *** [Release/obj.target/sse4_crc32/src/sse4_crc32.o] Error 1
```

The sse4_crc32 module updated to nan@2 and broke my library in node 0.10.26

Does nan want to support node 0.10.26 ?

kkoopa:
Nope, you have to use 0.10.29 or later.

rvagg:
@kkoopa how difficult would this compatibility be? I know there are a few companies stuck on older versions and the security patches aren't relevant for their use cases.

Raynos:
At least for uber we really should use 0.10.32. We have no reason to not upgrade.

The bigger blocker is the memory leak which makes 2.0.0 unusable ;)

kkoopa:
Maintaining something this outdated is not worth the effort. It's currently 0.8.27+, 0.10.29+, 0.12.0+, 1.0.0+, 2.0.0+ and 3.0.0+, which is more than good enough. Everything takes precedence over supporting outdated minor versions of an outdated branch.

bnoordhuis:
It's easy to check the `NODE_{MAJOR,MINOR,PATCH}_VERSION` macros but I'm ambivalent on whether it's a good idea.

kkoopa:
Don't want to maintain more exceptions than necessary. This would also entail duplicating the functionality provided by https://gist.githubusercontent.com/tjfontaine/f869f373a8e9416809ba/raw/e3eb85201413a79d12ce24a7cb4b02edf0abc1a5/v0.10-invalid-utf8.patch

http://blog.nodejs.org/2014/06/16/openssl-and-breaking-utf-8-change/

The current version of Node 0.10 is 0.10.40, 0.10.29 was released over a year ago. I don't think requiring it at this point is aksing too much. If someone wants to run older versions of 0.10, they can apply the above patch themselves and build from source.

bnoordhuis:
See #399.

bnoordhuis:
Fixed by 60d6687.

[...]

Issue #394, opened 2015-08-04
nan 2.0.0 fails to build against node 0.10.x


rmoetwil:
nan 2.0.0 fails to build against node 0.10.x

Reason for this is the usage of v8::String::REPLACE_INVALID_UTF8 in nan.h.
REPLACE_INVALID_UTF8 is not supported in node 0.10

E.g. when building sleep 3.0.0

```
> sleep@3.0.0 install ...node_modules/sleep
> node-gyp rebuild

make: Entering directory `...node_modules/sleep/build'
  CXX(target) Release/obj.target/node_sleep/sleep.o
In file included from ../sleep.cc:2:0:
../node_modules/nan/nan.h: In constructor ‘Nan::Utf8String::Utf8String(v8::Local<v8::Value>)’:
../node_modules/nan/nan.h:1178:27: error: ‘REPLACE_INVALID_UTF8’ is not a member of ‘v8::String’
                           v8::String::REPLACE_INVALID_UTF8;
```

Apart from upgrading to node 0.12.x any ideas on how to resolve this?

kkoopa:
Use a supported version of Node 0.10 then. The current one is at 0.10.40. `REPLACE_INVALID_UTF8` is available since 0.10.29.

rmoetwil:
Ok. I was misled by the documentation, see https://v8docs.nodesource.com/node-0.10/d2/db3/classv8_1_1_string.html, which doesn't show the option.

But indeed I see the REPLACE_INVALID_UTF8 patch was applied in 0.10.29 and onwards.

kkoopa:
@rvagg why u no account for floating patches in the generated docs?

kkoopa:
As a followup: This should no longer be a problem.

Calpicow:
Using node v0.10.36, and running into this problem when trying to build bcrypt. This should be nan v2.0.5.

```
make: Entering directory `/tmp/node_modules/bcrypt/build'
  CXX(target) Release/obj.target/bcrypt_lib/src/blowfish.o
  CXX(target) Release/obj.target/bcrypt_lib/src/bcrypt.o
  CXX(target) Release/obj.target/bcrypt_lib/src/bcrypt_node.o
In file included from ../src/bcrypt_node.cc:1:
../node_modules/nan/nan.h:318: error: ‘REPLACE_INVALID_UTF8’ is not a member of ‘v8::String’
make: *** [Release/obj.target/bcrypt_lib/src/bcrypt_node.o] Error 1
[...]

PR #399, opened 2015-08-05
Add workaround for missing REPLACE_INVALID_UTF8

bnoordhuis:
`v8::String::REPLACE_INVALID_UTF8` was added in node.js v0.10.29.
Check the version macros to see if it's available.

Fixes: https://github.com/nodejs/nan/issues/394
Fixes: https://github.com/nodejs/nan/issues/396

R=@kkoopa, /cc @rvagg @Raynos

I think we should address this, I'm getting complaints from people about broken builds separate from the filed issues.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -309,6 +309,16 @@ template<typename P> class WeakCallbackInfo;
 
 namespace imp {
   static const size_t kMaxLength = 0x3fffffff;
+  // v8::String::REPLACE_INVALID_UTF8 was introduced
+  // in node.js v0.10.29 and v0.8.27.
+#if NODE_MAJOR_VERSION > 0 || \
+    NODE_MINOR_VERSION > 10 || \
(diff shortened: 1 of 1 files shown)


kkoopa:
What with < 0.8.27 then? And what are everybody doing on 0.10.26?
-0

bnoordhuis:
Didn't realize this was also released in v0.8.  Fixed up, PTAL.

bnoordhuis:
> And what are everybody doing on 0.10.26?

I can attest from first-hand experience that upgrades in large companies are often painful and prolonged.  Big businesses don't move fast (to their detriment but hey.)

kkoopa:
Code looks ok. The proper thing to do would be add 0.8.26 and 0.10.28 to CI so this does not regress, but that will slow down runs even more. I think the risk of regression is fairly small.

PR #404, opened 2015-08-06
Don't use clang++ / g++ syntax extension.

bnoordhuis:
Replace `wbcd->invoke` with `WeakCallbackInfo<P>::invoke` when passing
it as a callback.  The former is not proper syntax although sometimes
(but not always) accepted by compilers.

Should fix the following build error that people have been reporting:

    nan_weak.h:363:42: error: cannot create a non-constant pointer to
    member function
        v8::PersistentBase<T>::SetWeak(wcbd, wcbd->invoke);

Fixes: https://github.com/strongloop/fsevents/issues/80

R=@kkoopa

### DIFF
--- a/nan_weak.h
+++ b/nan_weak.h
@@ -254,7 +254,10 @@ NAN_INLINE void Persistent<T, M>::SetWeak(
         reinterpret_cast<Persistent<v8::Value>*>(this)
       , callback
       , parameter);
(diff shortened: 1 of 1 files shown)


kkoopa:
Looks good.

bnoordhuis:
Landed, thanks for the review.  Can you do a patch release for this?

pedrofranceschi:
+1 This is breaking production. Please fix.

Issue #410, opened 2015-08-11
Template function that used Local handler not working because of MaybeLocal


JCMais:
I'm having some trouble converting the following template to NAN 2

```cpp
template <typename> struct ResultCanBeNull : std::false_type {};
template <> struct ResultCanBeNull<char*> : std::true_type {};

template<typename TResultType, typename Tv8MappingType>
v8::Local<v8::Value> Curl::GetInfoTmpl( const Curl *obj, int infoId )
{
    NanEscapableScope();

    TResultType result;

    CURLINFO info = (CURLINFO) infoId;
    CURLcode code = curl_easy_getinfo( obj->curl, info, &result );

    v8::Local<v8::Value> retVal = NanUndefined();

    if ( code != CURLE_OK ) {

        Curl::ThrowError( "curl_easy_getinfo failed!", curl_easy_strerror( code ) );

    } else {

        //null pointer
        if ( ResultCanBeNull<TResultType>::value && !result ) {

            retVal = NanNew<v8::String>( "" ); //line:1102

        } else {

            retVal = NanNew<Tv8MappingType>( result ); //line:1106
        }

    }

    return NanEscapeScope( retVal );
}
```

Tv8MappingType can be of type v8::Number or v8::String, as Nan::New\<v8::String\> now returns a MaybeLocal handler, the code is not compiling.

```
..\src\Curl.cc(1102): error C2679: binary '=': no operator found which takes a right-hand operand of type 'Nan::MaybeLocal<v8::String>' (
or there is no acceptable conversion)
..\src\Curl.cc(1106): error C2679: binary '=': no operator found which takes a right-hand operand of type 'Nan::MaybeLocal<v8::String>' (
or there is no acceptable conversion)
```

Changing retVal to MaybeLocal is not going to work, because Nan::New\<v8::Number\> returns a Local handler directly.

Any help would be greatly appreciated. 

Original code at https://github.com/JCMais/node-libcurl/blob/develop/src/Curl.cc#L1078

kkoopa:
For the first one, use `EmptyString` https://github.com/nodejs/nan/blob/master/doc/new.md#api_nan_empty_string.

For the second, you need to convert the `MaybeLocal` to a `Local`, using either `MaybeLocal::ToLocal` or `MaybeLocal::ToLocalChecked`, if you are certain that `String` creation cannot fail.

JCMais:
[...]

PR #411, opened 2015-08-12
[Maybe types] Add MakeMaybe(...)

agnat:
Convert `Locals<T>` to `MaybeLocals<T>`. When called with a `MaybeLocal<T>` it is the identity function.

Inspired by #410.

### DIFF
--- a/doc/maybe_types.md
+++ b/doc/maybe_types.md
@@ -36,6 +36,7 @@ The `Nan::MaybeLocal` and `Nan::Maybe` types are monads that encapsulate `v8::Lo
   - <a href="#api_nan_get_start_column"><b><code>Nan::GetStartColumn()</code></b></a>
   - <a href="#api_nan_get_end_column"><b><code>Nan::GetEndColumn()</code></b></a>
   - <a href="#api_nan_clone_element_at"><b><code>Nan::CloneElementAt()</code></b></a>
+  - <a href="#api_nan_make_maybe"><b><code>Nan::MakeMaybe()</code></b></a>
 
 <a name="api_nan_maybe_local"></a>
 ### Nan::MaybeLocal
@@ -477,4 +478,23 @@ Signature:
 
 ```c++
 Nan::MaybeLocal<v8::Object> Nan::CloneElementAt(v8::Local<v8::Array> array, uint32_t index);
+```
+
+<a name="api_nan_make_maybe"></a>
+### Nan::MakeMaybe()
+
+Wraps a `v8::Local<>` in a `Nan::MaybeLocal<>`. When called with a `Nan::MaybeLocal<>` it just returns its argument. This is useful in generic template code that builds on NAN.
+
+Synopsis:
+
+```c++
+  MaybeLocal<v8::Number> someNumber = MakeMaybe(New<v8::Number>(3.141592654));
+  MaybeLocal<v8::String> someString = MakeMaybe(New<v8::String>("probably"));
+```
+
+Signature:
+
+```c++
+template <typename T, template <typename> class MaybeMaybe>
+Nan::MaybeLocal<T> Nan::MakeMaybe(MaybeMaybe<T> v);
 ```
--- a/nan.h
+++ b/nan.h
@@ -2157,6 +2157,32 @@ struct Tap {
 
 #undef TYPE_CHECK
 
+//=== Generic Maybefication ===================================================
+
+namespace imp {
+
+template <typename T> struct Maybefier;
+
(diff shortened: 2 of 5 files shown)


agnat:
Ready for review.

doc/maybe_types.md
```diff
@@ -478,3 +479,22 @@ Signature:
 ```c++
 Nan::MaybeLocal<v8::Object> Nan::CloneElementAt(v8::Local<v8::Array> array, uint32_t index);
 ```
+
+<a name="api_nan_make_maybe"></a>
+### Nan::MakeMaybe()
+
+Converts a `v8::Local<>` into a `Nan::MaybeLocal<>`. When called with a `Nan::MaybeLocal<>` it just returns it's argument. This is useful in generic template code that builds on NAN.
```
kkoopa: remove the apostrophe

nan.h
```diff
@@ -2157,6 +2157,32 @@ struct Tap {
...
+
+template <typename T> struct Maybefier<MaybeLocal<T> > {
+  static MaybeLocal<T> convert(MaybeLocal<T> v) {
+    return v;
+  }
+};
+
+}  // end of namespace imp
+
+template <typename T, template <typename> class CouldbeMaybe>
```
kkoopa: Is this valid in C++03? Any docs on it? Never seen the construct before and it is somewhat puzzling.

kkoopa:
Looks good, apart from the documentation nit. However, wouldn't `EnsureMaybe` or such be a more fitting name? `MakeMaybe` is good, but is it the best possible name?

doc/maybe_types.md
kkoopa: Wraps a `v8::Local<>` in a `Nan::MaybeLocal<>` is better than 'converts'.

nan.h
bnoordhuis: It's a called a template template parameter: http://stackoverflow.com/questions/213761/what-are-some-uses-of-template-template-parameters-in-c

It's valid C++03 and I'm reasonably sure it's even valid C++98.

agnat: What @bnoordhuis says.

IIRC, the STL container allocators rely on it pretty heavily...

agnat:
The best possible name still is `maybefy(...)`... definitely. ;)

But I already had my fun with naming the implementation detail `Maybefier`. So I went with `MakeMaybe(...)` because it has precedent in both NAN and C++/STL... think `make_pair(...)` &c. However, I'm ok with `EnsureMaybe(...)` too. So, since I'm a `+0` it's up to you guys. Just tell me what it should be. ;)

kkoopa:
`MakeMaybe` it is then. Go ahead and flatten and I'll merge once 2.1 gets near.

agnat:
Actually, `EnsureMaybe(...)` is pretty funny too... lol

So, `MakeMaybe(...)` ... definitely.

kkoopa:
Maybe `CouldbeMaybe` could be `MaybeMaybe`?

agnat:
Exactly.

kkoopa:
How much wood would a woodchuck chuck if a woodchuck would chuck wood?

nan.h
agnat: Ok, it's official. There is no good free text on template template parameters.

However, _The Text_ on the subject is [C++ Templates - The Complete Guide](http://www.josuttis.com/tmplbook/), a book every C++ programmer should have.

agnat:
What's up with AppVeyor? It's just doing its windows thing, being unstable and all...? Or should I be worried?

kkoopa:
Nothing to worry about. It is always flaky.

doc/maybe_types.md
```diff
@@ -478,3 +479,22 @@ Signature:
...
+```c++
+  MaybeLocal<v8::Number> someNumber = MakeMaybe(New<v8::Number>(3.141592654));
+  MaybeLocal<v8::String> someString = MakeMaybe(New<v8::String>("probably"));
+```
+
+Signature:
+
+```c++
+template <typename T, template <typename> class CouldbeMaybe>
+Nan::MaybeLocal<T> Nan::MakeMaybe(CouldbeMaybe<T> v);
```
kkoopa: This should say it takes a `MaybeMaybe<T>`.

agnat:
Done.

saper:
`CallMeMaybe` :)

By the way those AppVeyor failures are probably https://github.com/nodejs/nodejs.org/issues/167

kkoopa:
Merged as 48d7b53d9702b0c7a060e69ea10fea8fb48d814d

PR #412, opened 2015-08-12
Make better use of maybe versions of APIs

kkoopa:


bnoordhuis:
LGTM if it works.  AppVeyor is red but it looks like some kind of transient error.

kkoopa:
Committed as bfba85b1072d14c3a1a42ca49e0d43c7a437d638

PR #416, opened 2015-08-16
Fix shadowing issues

kkoopa:


bnoordhuis:
Commit log could go into  more detail about what shadows what.  I infer it's the parameter and the method?

bnoordhuis:
LGTM though.

kkoopa:
Committed as 0a9072d0701d2f42aa2592430af676dc189af784

PR #418, opened 2015-08-17
Add test for getting wrapped value to ObjectWrap test suite

kkoopa:
Fixes warning about unused member variable


rvagg:
Nice, +1

PR #421, opened 2015-08-22
new: allow utf16 string with length

LinusU:
Patches Nan::New so that it can be initialized with pointer to wides string and length.

Without this the following code failed:

```c++
Nan::New(pointer, length)
```

But the following code worked:

```c++
Nan::New<v8::String>(pointer, length)
```

After this patch they both work. It also adds tests and documentation for this feature.


kkoopa:
Merged as 66ac6e65c8ab9394ef588adfc59131b3b9d8347b

LinusU:
Nice :+1:

PR #425, opened 2015-08-24
Remove unused static std::map instances.

bnoordhuis:
They were added in commit 12948c68 ("Remove NanReturn* macros") but
have never been in use as far as I can tell.

R=@kkoopa


kkoopa:
LGTM. They are covered by https://github.com/nodejs/nan/blob/master/nan_callbacks.h#L103-L114.

PR #429, opened 2015-08-27
Workaround for crash in GetWrapper

saper:
Due to some static intialization
issues, do not use static map to
store function wrappers: always
generate a new one.

This is a workaround for a crash
where two node modules compiled
against libstdc++ and llvm libc++
meet to use GetWrapper() static
wrapper cache.

Workaround for: https://github.com/strongloop/fsevents/issues/82

### DIFF
--- a/nan_callbacks.h
+++ b/nan_callbacks.h
@@ -102,15 +102,7 @@ static const int kIndexPropertyFieldCount =      6;
 
 template<typename T, typename P>
 P *GetWrapper(T needle) {
-  static std::map<T, P*> haystack;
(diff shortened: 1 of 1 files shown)


kkoopa:
No, this introduces a memory leak.

saper:
@kkoopa do you have any idea why we started getting crashes with nan 2.0 and not earlier? My current theory is that this is due to C++ libraries mismatch, but nan was using `GetWrapper()` much earlier I think?

kkoopa:
No, it is new for 2.0

PR #431, opened 2015-08-27
Work around duplicate linking bug in clang

kkoopa:
Perhaps this makes clang behave better?
cc @lovell @saper

### DIFF
--- a/nan_callbacks.h
+++ b/nan_callbacks.h
@@ -54,29 +54,6 @@ typedef void(*IndexQueryCallback)(
 namespace imp {
 typedef v8::Local<v8::AccessorSignature> Sig;
 
-#define X(NAME)                                                                \
-struct NAME ## Wrapper {                                                       \
-  explicit NAME ## Wrapper(NAME ## Callback callback_) :                       \
-      callback(callback_) {}                                                   \
-  NAME ## Callback callback;                                                   \
-};
-
-X(Function)
-X(Getter)
-X(Setter)
-X(PropertyGetter)
-X(PropertySetter)
-X(PropertyEnumerator)
-X(PropertyDeleter)
-X(PropertyQuery)
-X(IndexGetter)
-X(IndexSetter)
-X(IndexEnumerator)
-X(IndexDeleter)
-X(IndexQuery)
-
-#undef X
-
 static const int kDataIndex =                    0;
 
 static const int kFunctionIndex =                1;
@@ -100,18 +77,6 @@ static const int kIndexPropertyDeleterIndex =    4;
 static const int kIndexPropertyQueryIndex =      5;
 static const int kIndexPropertyFieldCount =      6;
 
-template<typename T, typename P>
-P *GetWrapper(T needle) {
-  static std::map<T, P*> haystack;
-  typename std::map<T, P*>::iterator it =  // NOLINT(build/include_what_you_use)
-      haystack.find(needle);
-
-  if (it == haystack.end()) {
-    return haystack.insert(it, std::make_pair(needle, new P(needle)))->second;
-  } else {
-    return it->second;
-  }
-}
 }  // end of namespace imp
 
 #if NODE_MODULE_VERSION > NODE_0_10_MODULE_VERSION
--- a/nan_callbacks_12_inl.h
+++ b/nan_callbacks_12_inl.h
@@ -166,11 +166,12 @@ namespace imp {
 static
 void FunctionCallbackWrapper(const v8::FunctionCallbackInfo<v8::Value> &info) {
   v8::Local<v8::Object> obj = info.Data().As<v8::Object>();
-  FunctionWrapper *wrapper = static_cast<FunctionWrapper*>(
-      obj->GetAlignedPointerFromInternalField(kFunctionIndex));
+  FunctionCallback callback = reinterpret_cast<FunctionCallback>(
+      reinterpret_cast<intptr_t>(
+          obj->GetInternalField(kFunctionIndex).As<v8::External>()->Value()));
   FunctionCallbackInfo<v8::Value>
       cbinfo(info, obj->GetInternalField(kDataIndex));
-  wrapper->callback(cbinfo);
+  callback(cbinfo);
 }
 
 typedef void (*NativeFunction)(const v8::FunctionCallbackInfo<v8::Value> &);
@@ -183,9 +184,10 @@ void GetterCallbackWrapper(
   v8::Local<v8::Object> obj = info.Data().As<v8::Object>();
   PropertyCallbackInfo<v8::Value>
       cbinfo(info, obj->GetInternalField(kDataIndex));
-  GetterWrapper *wrapper = static_cast<GetterWrapper*>(
-      obj->GetAlignedPointerFromInternalField(kGetterIndex));
-  wrapper->callback(property.As<v8::String>(), cbinfo);
+  GetterCallback callback = reinterpret_cast<GetterCallback>(
+      reinterpret_cast<intptr_t>(
+          obj->GetInternalField(kGetterIndex).As<v8::External>()->Value()));
+  callback(property.As<v8::String>(), cbinfo);
 }
 
 typedef void (*NativeGetter)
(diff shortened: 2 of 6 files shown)


xzyfer:
I'll give it a shot

kkoopa:
Also try without the static.

xzyfer:
Will do

xzyfer:
It's `SIGSEGV`s without the `static` in my test case. With the `static` my test case passes. I'm currently trying this patch over our production system.

kkoopa:
Check the symbol table as well. Problem with adding the static is that it makes lots of copies of the map. It should be one per translation unit.

kkoopa:
Can you test a third option? move the static map outside the function and make the function non-static

xzyfer:
> Check the symbol table as well

Sorry I'm not sure how to do this.

> move the static map outside the function and make the function non-static

Trying now.

kkoopa:
`nm -gC addon.node`

xzyfer:
> move the static map outside the function and make the function non-static

Having trouble getting this to compile. Likely my fault. Can you show me a diff of what you're thinking?

xzyfer:
> nm -gC addon.node

OSX doesn't support the `-C` flag. I've output in this gist - https://gist.github.com/xzyfer/eaabbc1e2851215f787b

kkoopa:
Nevermind moving the map.

I don't see it linked at all in the gist, which is somewhat strange.

kkoopa:
@bnoordhuis Why is clang so bad?

xzyfer:
I'm compiling against iojs v2.5.0 for reference.

nan_callbacks.h
```diff
@@ -101,7 +101,7 @@ static const int kIndexPropertyQueryIndex =      5;
 static const int kIndexPropertyFieldCount =      6;
 
 template<typename T, typename P>
-P *GetWrapper(T needle) {
+static inline P *GetWrapper(T needle) {
```
bnoordhuis: Just `inline` is sufficient but each compilation unit will now end up with its own instance of `haystack`.  I don't think that's the intended behavior.

To be honest, I don't really understand why `haystack` exists.  Can't it just be removed?

kkoopa: Just inline apparently led to SIGSEGV. The problem is likely that clang includes static locals many times. The haystack exists to enforce singleton behavior for callbacks, which is necessary since the wrappers are never freed. The wrappers, in turn, are necessary, because SetInternalFieldPointer only accepts data pointers, not function pointers.

xzyfer: For reference, this is the stacktrace for the SIGSEGV when the static is removed.

```
PID 94064 received SIGSEGV for address: 0x0
0   segfault-handler.node               0x0000000100ffa90f _ZL16segfault_handleriP9__siginfoPv + 287
1   libsystem_platform.dylib            0x00007fff8e80a5aa _sigtramp + 26
2   ???                                 0x0000000000000783 0x0 + 1923
3   binding.node                        0x0000000103107d8c _ZNSt3__16__treeINS_12__value_typeIPFvRKN3Nan20FunctionCallbackInfoIN2v85ValueEEEEPNS2_3imp15FunctionWrapperEEENS_19__map_value_compareISA_SE_NS_4lessISA_EELb1EEENS_9allocatorISE_EEE20__node_insert_uniqueENS_21__tree_const_iteratorISE_PNS_11__tree_nodeISE_PvEElEESQ_ + 56
4   binding.node                        0x0000000103107d32 _ZNSt3__16__treeINS_12__value_typeIPFvRKN3Nan20FunctionCallbackInfoIN2v85ValueEEEEPNS2_3imp15FunctionWrapperEEENS_19__map_value_compareISA_SE_NS_4lessISA_EELb1EEENS_9allocatorISE_EEE15__insert_uniqueINS_4pairISA_SD_EEEENS_15__tree_iteratorISE_PNS_11__tree_nodeISE_PvEElEENS_21__tree_const_iteratorISE_ST_lEEOT_ + 62
5   binding.node                        0x0000000103107c3b _ZN3Nan3imp7FactoryIN2v816FunctionTemplateEE3NewEPFvRKNS_20FunctionCallbackInfoINS2_5ValueEEEENS2_5LocalIS6_EENSC_INS2_9SignatureEEE + 235
6   binding.node                        0x00000001031023c1 _Z14RegisterModuleN2v86HandleINS_6ObjectEEE + 52
7   node                                0x0000000100601047 _ZN4node6DLOpenERKN2v820FunctionCallbackInfoINS0_5ValueEEE + 725
8   node                                0x00000001001749ee _ZN2v88internal25FunctionCallbackArguments4CallEPFvRKNS_20FunctionCallbackInfoINS_5ValueEEEE + 158
9   node                                0x000000010019c97b _ZN2v88internalL19HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateERNS0_12_GLOBAL__N_116BuiltinArgumentsILNS0_21BuiltinExtraArgumentsE1EEE + 811
10  node                                0x000000010019f1cd _ZN2v88internalL21Builtin_HandleApiCallEiPPNS0_6ObjectEPNS0_7IsolateE + 61
11  ???                                 0x000039d1e69060bb 0x0 + 63573679169723
```

_It's bedtime here, I'll check back in a couple hours_

bnoordhuis: > SetInternalFieldPointer only accepts data pointers, not function pointers.

`object->SetAlignedPointerInInternalField(index, reinterpret_cast<void*>(callback))`?  Or is there some subtlety here I'm misunderstanding?

kkoopa: It is not guaranteed to be correct. According to the standard, `void *` is a data pointer, not a function pointer. Function pointers may be of different size or refer to other memory areas than data pointers.

https://stackoverflow.com/questions/12358843/why-are-function-pointers-and-data-pointers-incompatible-in-c-c

bnoordhuis: Yes, but see my comment about half the world stopping if it didn't work.  This is one of those areas where it's okay to ignore the spec and get on with work.

bnoordhuis: Also, the behavior is well-defined on POSIX and Windows and that's all we care about.  The few people still running VMS or DOS can fend for themselves. :-)

kkoopa: Fine.

kkoopa:
There. Updated.

nan_callbacks_12_inl.h
```diff
@@ -166,11 +166,11 @@ namespace imp {
 static
 void FunctionCallbackWrapper(const v8::FunctionCallbackInfo<v8::Value> &info) {
   v8::Local<v8::Object> obj = info.Data().As<v8::Object>();
-  FunctionWrapper *wrapper = static_cast<FunctionWrapper*>(
-      obj->GetAlignedPointerFromInternalField(kFunctionIndex));
+  FunctionCallback callback =  reinterpret_cast<FunctionCallback>(reinterpret_cast<intptr_t>(
```
bnoordhuis: I don't think you need the extra intptr_t cast.

bnoordhuis:
LGTM with a suggestion.  Some of the lines look really long.

nan_callbacks_12_inl.h
kkoopa: It can warn otherwise.

lovell:
My initial tests with io.js v3.2.0 and clang-600.0.57 suggest the segfault is no more, thanks @kkoopa. I'll be able to run more thorough tests in a few hours.

kkoopa:
No, this does not work. Unaligned pointerness makes it impossible to use `SetInternalPointer`, and I guess the same goes for `v8::External`, leading to runtime faults instead.

kkoopa:
There, that should do it.

saper:
So we basically get rid of those wrappers, at the cost of converting data<->code pointers via casts, right?

kkoopa:
Yes, which is rather horrible and not allowed or well-defined in general. Clang should really get its act together if it wants to compile with the big boys.

bnoordhuis:
> Unaligned pointerness makes it impossible to use SetInternalPointer

Was it actually making tests fail?  Functions are normally aligned on 4 or 16 byte boundaries.

saper:
Not sure it is clang's fault, maybe we overstretched compatibility between libstdc++ and libc++.

kkoopa:
@bnoordhuis Yes, it failed with OS X on Travis. `v8::External` did however work.

xzyfer:
I can confirm I can no longer cause the segfault with this patch applied.

lovell:
I added a [test](https://github.com/lovell/sharp/blob/master/test/unit/require.js) to sharp that, when compiled with clang, triggers the segfault on [nan v2](https://travis-ci.org/lovell/sharp-osx-ci/builds/68804844).

Switching to the [kkoopa/nan#clang_sucks](https://travis-ci.org/lovell/sharp-osx-ci/builds/77601537) fork/branch works perfectly.

(Both versions pass the MSVC/Windows/Appveyor and gcc/Centos/Snap based CI builds without regression.)

Thanks again for all your help @kkoopa!

xzyfer:
Thanks for that test @lovell. I'll add one to node-sass as well.

kkoopa:
Merged 11902dade135c409d21f0d9de051f3dedb319068

PR #442, opened 2015-09-08
Test on Node.js 4.0.

kkoopa:
I don't think it's worth testing on io.js any more.


bnoordhuis:
I speculate there will be one or two more io.js v3 releases so I wouldn't axe it from the test matrix just yet.

kkoopa:
Alright. I'll update this and add it back in a couple of days, when the CI environments actually can install v4.

wadey:
Node v4 works on both appveyor now:

- https://ci.appveyor.com/project/WadeSimmons/node-microtime/build/48
- https://github.com/wadey/node-microtime/blob/master/.appveyor.yml#L39

kkoopa:
Excellent.

PR #443, opened 2015-09-08
EscapableHandleScope in Nan::NewBuffer for Node 0.8 and 0.10

mikemorris:
I'm guessing https://github.com/mikemorris/nan/blob/master/test/js/buffer-test.js#L19 didn't catch this because the scope doesn't change?

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1026,12 +1026,12 @@ class Utf8String {
     , node::Buffer::free_callback callback
     , void *hint
   ) {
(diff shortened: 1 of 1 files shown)


kkoopa:
The V8 versions in 0.8 and 0.10 were always very sloppy with the scopes. A problem would likely only manifest if a garbage collection happened in between scope destruction and use. I'm not entirely sure.

PR #447, opened 2015-09-10
Remove duplicate v8 function reference

mscdex:
The function name has always been `AdjustAmountOfExternalAllocatedMemory()` for all versions supported by nan.


kkoopa:
Ah, yes. The (omitted) point should have been that the argument type has changed from `int` to `size_t` and that NAN keeps it as `int`. Could you add a note about that as well?

kkoopa:
Merged as 9347dc9f43bd4e5e54926fa38dd915b7d3fa3c08

Issue #448, opened 2015-09-10
'make test' fails with errors in header files for node v4.0


smrtlion:
Hi,

I was trying to run the test suite that comes with nan with node v4.0.0 but I get the following when invoking 'make test' (this is done on an xLinux machine):

nodebot@machine01:/svt/nodebot/test-crypt/nan$ make test >> test.log
gyp info it worked if it ends with ok
gyp info using node-gyp@3.0.1
gyp info using node@4.0.0 | linux | ia32
gyp info spawn python2
gyp info spawn args [ '/svt/nodebot/test-crypt/nan/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'make',
gyp info spawn args   '-I',
gyp info spawn args   '/svt/nodebot/test-crypt/nan/test/build/config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/svt/nodebot/test-crypt/nan/node_modules/node-gyp/addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/home/nodebot/.node-gyp/4.0.0/include/node/common.gypi',
gyp info spawn args   '-Dlibrary=shared_library',
gyp info spawn args   '-Dvisibility=default',
gyp info spawn args   '-Dnode_root_dir=/home/nodebot/.node-gyp/4.0.0',
gyp info spawn args   '-Dnode_gyp_dir=/svt/nodebot/test-crypt/nan/node_modules/node-gyp',
gyp info spawn args   '-Dnode_lib_file=node.lib',
gyp info spawn args   '-Dmodule_root_dir=/svt/nodebot/test-crypt/nan/test',
gyp info spawn args   '--depth=.',
gyp info spawn args   '--no-parallel',
gyp info spawn args   '--generator-output',
gyp info spawn args   'build',
gyp info spawn args   '-Goutput_dir=.' ]
gyp info spawn make
gyp info spawn args [ 'BUILDTYPE=Release', '-C', 'build' ]
In file included from /home/nodebot/.node-gyp/4.0.0/include/node/node.h:42:0,
                 from ../../nan.h:24,
                 from ../cpp/accessors.cpp:9:
/home/nodebot/.node-gyp/4.0.0/include/node/v8.h:336:1: error: expected unqualified-id before ‘using’
[...]

PR #450, opened 2015-09-12
Check for C++11 compiler first

agnat:
As discussed in #448.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -20,10 +20,29 @@
 #ifndef NAN_H_
 #define NAN_H_
 
+#include <node_version.h>
+
+#define NODE_0_10_MODULE_VERSION 11
+#define NODE_0_12_MODULE_VERSION 14
+#define ATOM_0_21_MODULE_VERSION 41
+#define IOJS_1_0_MODULE_VERSION  42
+#define IOJS_1_1_MODULE_VERSION  43
+#define IOJS_2_0_MODULE_VERSION  44
+#define IOJS_3_0_MODULE_VERSION  45
+
+#ifdef _MSC_VER
+# define NAN_HAS_CPLUSPLUS_11 (_MSC_VER >= 1800)
+#else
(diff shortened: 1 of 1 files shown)


rvagg:
lgtm

greenboxal:
It seems to be worse than that. I can't build nan 2.0.9 using GCC 4.8 against node 0.12 or lower.

https://travis-ci.org/pagarme/node-rsa-keygen/jobs/80027572

There's a lot of errors regarding double brackets on `pre43` code. I fear that this could extend to v8 code also, breaking any possibility of using newer compilers with it.

kkoopa:
Something's wrong under AppVeyor. Does msvs2013 not set the proper define? It is C++11-capable, as it can compile the code, so something's wrong with the detection logic.

kkoopa:
Yes, typical Microsoft... Never get anything done right.
https://connect.microsoft.com/VisualStudio/feedback/details/763051/a-value-of-predefined-macro-cplusplus-is-still-199711l

> Thanks for your feedback. Could you please provide some screenshots to help us better understand this scenario?

need to add `#ifdef _MSC_VER && _MSC_VER < 1800`

agnat:
Really?!?...wow... just... wow. lol.

Just let me real quick...

agnat:
Like this?

kkoopa:
The jury is still out, but I guess the outcome will be pleasing.

kkoopa:
Seems to work.

brodybits:
We should also add:
```c++
#define NODE_4_00_MODULE_VERSION 46
```
see #477

kkoopa:
Yes, that should be added, although it ought to be `NODE_4_0_MODULE_VERSION`.

kkoopa:
Merged as 48cc1d8543aff67c2181c73caddcc94ba8fd2230

PR #452, opened 2015-09-12
persistent: remove double heading

LinusU:
I think that this heading was accidentally added twice since it exists further down as well.


rvagg:
thanks, landed @ 24f8b20

PR #457, opened 2015-09-14
doc: fix minor formatting errors

thefourtheye:
This patch does the following

 - Completes the backtick in `PropertyCallbackInfo`

 - Separates code from the explanation in "Method declaration" section

 - Aligns the parameters in the signatures to the first parameter

 - Separates two signatures with `;` so that they both will be Syntax highlighted.


kkoopa:
Thanks

Issue #455, opened 2015-09-13
Expose a portable method for reading the bytes out of typedarrays


mikolalysenko:
What are the correct invocations for getting the bytes out of a typedarray?  This broke with migration to v8 version 4 in iojs 3 and up.

Going forward what is the right strategy for reading and writing typedarray data in NAN 2+?

kkoopa:
NAN does not support typed arrays. Use buffers.

mikolalysenko:
@kkoopa what about systems that would need typedarrays, for example numerical computing or WebGL bindings?

kkoopa:
It should be doable with buffers at the C++ level, else write such a method and use it. Typed arrays are not available on all supported platforms and will not be supported by NAN.

rvagg:
@mikolalysenko since TypedArrays are now basically interchangable with Buffer because they share the same underlying implementation you should be able to write such code to work with Buffer exclusively, you can even instantiate a Buffer with a TypedArray now.

If you don't care about older versions of Node that don't support TypedArray then perhaps you shouldn't be concerned about using NAN for those specific instances anyway and just work directly with the V8 APIs that you want in those cases.

mikolalysenko:
@rvagg right, but there are other versions of node that do use typedarrays, for example node 0.10, node 0.12, iojs 1 and iojs 2.  Things changed in iojs 3 and node 4.  Before, you could get the storage of a typedarray using something like the following:

```c++
unsigned int offset = array->Get(NanNew<v8::String>("byteOffset"))->Uint32Value();
void* data = (void*) &((char*) array->GetIndexedPropertiesExternalArrayData())[offset];
```

Starting with v8 version 4, this changed.  Now you need to do the following:

```c+
unsigned int offset = array->ByteOffset();
void* data = (void*)((char*)array->Buffer()->GetContents().Data() + offset);
```

Now you could use preprocessor macros to hack around this, but isn't this the problem NAN is supposed to solve?

mikolalysenko:
Note that for headless-gl, I want to support all versions of node >= 0.10.  It seems reasonable to have an interface that works for all of these cases.

mikolalysenko:
Or illustrated in a different way, we could break down the history of node into 3 epochs:

```
        Before typedarrays        External typedarray API     New typedarray API
     o---------------------------o---------------------------o-------------------> Now
     ^                           ^                           ^
Birth of node                    node 0.8                    v8 v4
```

[...]

PR #459, opened 2015-09-14
Add helper method for accessing contents of typedarrays

mikolalysenko:
This PR adds a portable helper method for reading the contents of a typedarray (ArrayBufferView) from C++, fixing https://github.com/nodejs/nan/issues/455.

I tried to follow the API conventions from `Utf8String` for this module closely.  Note that this does not make a copy of the typedarray data or hold a reference to the underlying array object, so it would require the user to know what they are doing and not try to access these bytes outside the natural lifetime of the object.

### DIFF
--- a/nan_typedarray_contents.h
+++ b/nan_typedarray_contents.h
@@ -70,8 +70,8 @@ class TypedArrayContents {
   }
 
   NAN_INLINE size_t length() const            { return length_; }
-  NAN_INLINE T* const operator*()             { return data_;   }
-  NAN_INLINE const T* const operator*() const { return data_;   }
+  NAN_INLINE T* operator*()             { return data_;   }
+  NAN_INLINE const T* operator*() const { return data_;   }
 
  private:
   NAN_DISALLOW_ASSIGN_COPY_MOVE(TypedArrayContents)
--- a/nan_typedarray_contents.h
+++ b/nan_typedarray_contents.h
@@ -58,7 +58,17 @@ class TypedArrayContents {
 
 #endif
 
+
+#if defined(_MSC_VER)
+    assert(reinterpret_cast<uintptr_t>(data_) % __alignof(T) == 0);
+#elif defined(__GNUC__)
+    assert(reinterpret_cast<uintptr_t>(data_) % __alignof__(T) == 0);
+#elif NODE_MODULE_VERSION >= IOJS_3_0_MODULE_VERSION
+    assert(reinterpret_cast<uintptr_t>(data_) % alignof(T) == 0);
+#else
     assert(reinterpret_cast<uintptr_t>(data_) % sizeof(T) == 0);
+#endif
+
   }
 
   NAN_INLINE size_t length() const            { return length_; }
--- a/nan_typedarray_contents.h
+++ b/nan_typedarray_contents.h
@@ -32,7 +32,7 @@ class TypedArrayContents {
       data_   = reinterpret_cast<T*>(data + byte_offset);
     }
 
-#elif NODE_MODULE_VERSION >= NODE_0_8_MODULE_VERSION
+#else
 
     if (from->IsObject() && !from->IsNull()) {
 
@@ -56,18 +56,9 @@ class TypedArrayContents {
       }
     }
 
-#else
-
-    //TypedArrays not supported on node < 0.8
-    assert(false);
-
 #endif
 
-#if NODE_MODULE_VERSION >= IOJS_3_0_MODULE_VERSION
-    assert(reinterpret_cast<uintptr_t>(data_) % alignof(T) == 0);
-#else
     assert(reinterpret_cast<uintptr_t>(data_) % sizeof(T) == 0);
-#endif
   }
 
   NAN_INLINE size_t length() const            { return length_; }
--- a/nan_typedarray_contents.h
+++ b/nan_typedarray_contents.h
@@ -1,0 +1,82 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+#ifndef NAN_TYPEDARRAY_CONTENTS_H_
+#define NAN_TYPEDARRAY_CONTENTS_H_
+
+template<typename T>
+class TypedArrayContents {
+ public:
+  NAN_INLINE explicit TypedArrayContents(v8::Local<v8::Value> from) :
+      length_(0), data_(NULL) {
+
+#if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                      \
+  (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
+
+    if(from->IsArrayBufferView()) {
+      v8::Local<v8::ArrayBufferView> array =
+        v8::Local<v8::ArrayBufferView>::Cast(from);
+
+      int byteLength = array->ByteLength();
+      int byteOffset = array->ByteOffset();
+      v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
+
+      void* data = buffer->GetContents().Data();
+      length_ = byteLength / sizeof(T);
+      data_   = (T*)((void*)((char*)data + byteOffset));
+    }
+
+#elif NODE_MODULE_VERSION >= NODE_0_8_MODULE_VERSION
+
+    if (from->IsObject() && !from->IsNull()) {
+
+      v8::Local<v8::Object> array = v8::Local<v8::Object>::Cast(from);
+
+      v8::Local<v8::Value> byteLength = array->Get(
+        Nan::New<v8::String>("byteLength").ToLocalChecked());
+      v8::Local<v8::Value> byteOffset = array->Get(
+        Nan::New<v8::String>("byteOffset").ToLocalChecked());
+
+      if(byteLength->IsUint32() && byteOffset->IsUint32()) {
+
+        unsigned int length = byteLength->Uint32Value();
+        void* data = array->GetIndexedPropertiesExternalArrayData();
+
+        if(data) {
+          length_ = length / sizeof(T);
+          data_   = (T*)((void*)((char*)data));
+        }
+      }
+    }
+
+#endif
+  }
+
+  NAN_INLINE int length() const         { return length_; }
+  NAN_INLINE T* operator*()             { return data_;   }
+  NAN_INLINE const T* operator*() const { return data_;   }
+
+ private:
+   int  length_;
+   T*   data_;
+};
+
+#endif  // NAN_TYPEDARRAY_CONTENTS_H_
--- a/nan_typedarray_contents.h
+++ b/nan_typedarray_contents.h
@@ -56,6 +56,11 @@ class TypedArrayContents {
       }
     }
 
+#else
+
+    //TypedArrays not supported on node < 0.8
+    assert(false);
+
 #endif
 
 #if NODE_MODULE_VERSION >= IOJS_3_0_MODULE_VERSION
--- a/nan_typedarray_contents.h
+++ b/nan_typedarray_contents.h
@@ -1,23 +1,10 @@
-// Copyright Joyent, Inc. and other Node contributors.
-//
-// Permission is hereby granted, free of charge, to any person obtaining a
-// copy of this software and associated documentation files (the
-// "Software"), to deal in the Software without restriction, including
-// without limitation the rights to use, copy, modify, merge, publish,
-// distribute, sublicense, and/or sell copies of the Software, and to permit
-// persons to whom the Software is furnished to do so, subject to the
-// following conditions:
-//
-// The above copyright notice and this permission notice shall be included
-// in all copies or substantial portions of the Software.
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
-// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
-// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
-// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
-// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
-// USE OR OTHER DEALINGS IN THE SOFTWARE.
+/*********************************************************************
+ * NAN - Native Abstractions for Node.js
+ *
+ * Copyright (c) 2015 NAN contributors
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
 
 #ifndef NAN_TYPEDARRAY_CONTENTS_H_
 #define NAN_TYPEDARRAY_CONTENTS_H_
@@ -31,17 +18,18 @@ class TypedArrayContents {
 #if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                      \
   (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
 
-    if(from->IsArrayBufferView()) {
+    if (from->IsArrayBufferView()) {
       v8::Local<v8::ArrayBufferView> array =
         v8::Local<v8::ArrayBufferView>::Cast(from);
 
-      int byteLength = array->ByteLength();
-      int byteOffset = array->ByteOffset();
+      const size_t    byteLength = array->ByteLength();
+      const ptrdiff_t byteOffset = array->ByteOffset();
       v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
 
-      void* data = buffer->GetContents().Data();
+      char* data = reinterpret_cast<char*>(buffer->GetContents().Data());
+
       length_ = byteLength / sizeof(T);
-      data_   = (T*)((void*)((char*)data + byteOffset));
+      data_   = reinterpret_cast<T*>(data + byteOffset);
     }
 
 #elif NODE_MODULE_VERSION >= NODE_0_8_MODULE_VERSION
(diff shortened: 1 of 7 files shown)


nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
+// Copyright Joyent, Inc. and other Node contributors.
```
bnoordhuis: Why (c) Joyent?

nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
...
+
+#if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                      \
+  (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
+
+    if(from->IsArrayBufferView()) {
+      v8::Local<v8::ArrayBufferView> array =
+        v8::Local<v8::ArrayBufferView>::Cast(from);
+
+      int byteLength = array->ByteLength();
+      int byteOffset = array->ByteOffset();
```
bnoordhuis: Can you use `size_t` here and in general?  Maybe make them `const` while you're here.

nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
...
+template<typename T>
+class TypedArrayContents {
+ public:
+  NAN_INLINE explicit TypedArrayContents(v8::Local<v8::Value> from) :
+      length_(0), data_(NULL) {
+
+#if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                      \
+  (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
+
+    if(from->IsArrayBufferView()) {
```
bnoordhuis: Minor style infraction: space after `if`.

test/cpp/typedarrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <stdint.h>
+
+using namespace Nan;  // NOLINT(build/namespaces)
+
+NAN_METHOD(ReadU8) {
+  Nan::TypedArrayContents<uint8_t> data(info[0]);
+
+  v8::Local<v8::Array> result = Nan::New<v8::Array>(data.length());
+  for (int i=0; i<data.length(); ++i) {
+    result->Set(i, Nan::New<v8::Number>((*data)[i]));
```
kkoopa: `Nan::Set(result, ...`

nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
...
+      v8::Local<v8::ArrayBufferView> array =
+        v8::Local<v8::ArrayBufferView>::Cast(from);
+
+      int byteLength = array->ByteLength();
+      int byteOffset = array->ByteOffset();
+      v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
+
+      void* data = buffer->GetContents().Data();
+      length_ = byteLength / sizeof(T);
+      data_   = (T*)((void*)((char*)data + byteOffset));
```
bnoordhuis: This will result in undefined behavior if the pointer's alignment doesn't satisfy `alignas(T)`.

Aside, can you use C++ casts here?

test/cpp/typedarrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+  info.GetReturnValue().Set(result);
+}
+
+NAN_METHOD(ReadI32) {
+  Nan::TypedArrayContents<int32_t> data(info[0]);
+
+  v8::Local<v8::Array> result = Nan::New<v8::Array>(data.length());
+  for (int i=0; i<data.length(); ++i) {
+    result->Set(i, Nan::New<v8::Number>((*data)[i]));
```
kkoopa: `Nan::Set`

nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
...
+    if (from->IsObject() && !from->IsNull()) {
+
+      v8::Local<v8::Object> array = v8::Local<v8::Object>::Cast(from);
+
+      v8::Local<v8::Value> byteLength = array->Get(
+        Nan::New<v8::String>("byteLength").ToLocalChecked());
+      v8::Local<v8::Value> byteOffset = array->Get(
+        Nan::New<v8::String>("byteOffset").ToLocalChecked());
+
+      if(byteLength->IsUint32() && byteOffset->IsUint32()) {
```
bnoordhuis: Style error.

test/cpp/typedarrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+  info.GetReturnValue().Set(result);
+}
+
+
+NAN_METHOD(ReadFloat) {
+  Nan::TypedArrayContents<float> data(info[0]);
+
+  v8::Local<v8::Array> result = Nan::New<v8::Array>(data.length());
+  for (int i=0; i<data.length(); ++i) {
+    result->Set(i, Nan::New<v8::Number>((*data)[i]));
```
kkoopa: `Nan::Set`

nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
...
+
+        unsigned int length = byteLength->Uint32Value();
```
bnoordhuis: `const uint32_t`?

test/cpp/typedarrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+  info.GetReturnValue().Set(result);
+}
+
+NAN_METHOD(ReadDouble) {
+  Nan::TypedArrayContents<double> data(info[0]);
+
+  v8::Local<v8::Array> result = Nan::New<v8::Array>(data.length());
+  for (int i=0; i<data.length(); ++i) {
+    result->Set(i, Nan::New<v8::Number>((*data)[i]));
```
kkoopa: `Nan::Set`

nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
...
+        void* data = array->GetIndexedPropertiesExternalArrayData();
+
+        if(data) {
```
bnoordhuis: Style error.  I'm going to stop pointing them out from now on.

nan_typedarray_contents.h
```diff
@@ -0,0 +1,82 @@
...
+#endif
+  }
+
+  NAN_INLINE int length() const         { return length_; }
+  NAN_INLINE T* operator*()             { return data_;   }
+  NAN_INLINE const T* operator*() const { return data_;   }
+
+ private:
+   int  length_;
+   T*   data_;
```
bnoordhuis: Consider making these `const` (`T* const` in the case of `data_`.)

test/cpp/typedarrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <stdint.h>
+
+using namespace Nan;  // NOLINT(build/namespaces)
+
+NAN_METHOD(ReadU8) {
+  Nan::TypedArrayContents<uint8_t> data(info[0]);
+
+  v8::Local<v8::Array> result = Nan::New<v8::Array>(data.length());
+  for (int i=0; i<data.length(); ++i) {
```
bnoordhuis: Style errors: needs more spaces.

kkoopa:
Would it make sense to make the class non-heap-allocatable? or is there a legitimate use case for having it on the heap, what with the lifetime issues and that?

nan_typedarray_contents.h
mikolalysenko: I just copy-pasted the boilerplate from nan.h.  If you have an alternative license I should use, let me know.

nan_typedarray_contents.h
mikolalysenko: Sure.

nan_typedarray_contents.h
bnoordhuis: Did you mean nan_string_bytes.h?  I don't see Joyent copyright boilerplate anywhere else.

Aside, I speculate @indutny included it because it was copied from joyent/node but I think we can just remove it; most of that code was originally written by me.

indutny: I'm fine with removing this.

nan_typedarray_contents.h
mikolalysenko: Using C++ casts is no big deal.  Regarding alignment, what do you think the right behavior should be?  I can think of at least 3 options:

1.  Do nothing.
2.  Detect bad alignment at run time and return NULL.
3.  Make a local copy with the correct alignment.

nan_typedarray_contents.h
mikolalysenko: Yes, I just grabbed a random file and must've picked the wrong one.

mikolalysenko:
I agree that there doesn't seem to be a good reason for heap ownership, so I think it seems reasonable to disable it.  If you did need to track the data on the heap anyway, it would be no big deal to instantiate the class locally and copy the data over manually.

nan_typedarray_contents.h
mikolalysenko: Actually there is a 4th option:

* Use `std::align` to coerce the pointer to an aligned value.

mikolalysenko:
Ok, I fixed up some of the style issues and switched some of the types and macros around to be more idiomatic/descriptive.

I still haven't done anything with alignment, since this requires making a decision on what to do in the case of unaligned data.  I think that doing nothing is reasonable, but if we commit to this, then it should be documented in `v8_misc.md`.

indutny:
Just curious, what are you going to do about alignment?

mikolalysenko:
@indutny currently leaning toward doing nothing, but returning NULL seems ok too.

indutny:
Aah, I re-read the comments. Everything is clear now, nvm.

nan_typedarray_contents.h
```diff
@@ -0,0 +1,77 @@
...
+
+    if (from->IsArrayBufferView()) {
+      v8::Local<v8::ArrayBufferView> array =
+        v8::Local<v8::ArrayBufferView>::Cast(from);
+
+      const size_t    byteLength = array->ByteLength();
+      const ptrdiff_t byteOffset = array->ByteOffset();
+      v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
+
+      char* data = reinterpret_cast<char*>(buffer->GetContents().Data());
```
bnoordhuis: Can you use `static_cast` here?

nan_typedarray_contents.h
```diff
@@ -0,0 +1,77 @@
...
+
+#if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                      \
+  (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
+
+    if (from->IsArrayBufferView()) {
+      v8::Local<v8::ArrayBufferView> array =
+        v8::Local<v8::ArrayBufferView>::Cast(from);
+
+      const size_t    byteLength = array->ByteLength();
+      const ptrdiff_t byteOffset = array->ByteOffset();
```
bnoordhuis: Minor style nit but I think nan predominantly uses snake_case, not camelCase, for local variables.

nan_typedarray_contents.h
```diff
@@ -0,0 +1,77 @@
...
+      v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
+
+      char* data = reinterpret_cast<char*>(buffer->GetContents().Data());
+
+      length_ = byteLength / sizeof(T);
+      data_   = reinterpret_cast<T*>(data + byteOffset);
```
bnoordhuis: I think I'd insert an assert that checks the alignment.  You could use `alignof(T)` here but not in general because it's a C++11 construct (io.js v3 and node v4 imply C++11 support but that's not the case with node v0.10 and v0.12.)

Maybe settle for `assert(reinterpret_cast<uintptr_t>(data + byte_offset) % sizeof(T) == 0)`.  There are perhaps still one or two edge cases that will slip through, though.  I'm thinking of things like performing SSE operations on an array of doubles that only has an alignment of 8.

test/cpp/typedarrays.cpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  info.GetReturnValue().Set(result);
+}
+
+NAN_METHOD(ReadI32) {
+  Nan::TypedArrayContents<int32_t> data(info[0]);
+
+  v8::Local<v8::Array> result = Nan::New<v8::Array>(data.length());
+  for (size_t i=0; i<data.length(); ++i) {
```
bnoordhuis: Can you fix the style issues in the test?

nan_typedarray_contents.h
mikolalysenko: What about situations where the user would read out the contents of a typed array and then immediately copy it to local storage?  For example, maybe you want to read out the typedarray data and then immediately pass it to the OpenGL driver, which internally just copies it out.

bnoordhuis: Then you'd just use a `TypedArrayContents<char>`, wouldn't you?

mikolalysenko: Fair enough.

mikolalysenko:
Ok, I added an assert to check the alignment and made a few more style changes in the test cases.

mikolalysenko:
Ah whoops, I messed up the rebase and named that last commit incorrectly.  It should say "assert false when typedarrays missing" or something to that effect.

kkoopa:
[AppVeyor says no-go](https://ci.appveyor.com/project/RodVagg/nan/build/1007/job/uvctrj09asbd5qyq#L242).

```
c:\projects\nan\nan_typedarray_contents.h(67): error C2275: 'T' : illegal use of this type as an expression [C:\projects\nan\test\build\typedarrays.vcxproj]
          ..\cpp\typedarrays.cpp(16) : see declaration of 'T'
          c:\projects\nan\nan_typedarray_contents.h(15) : while compiling class template member function 'Nan::TypedArrayContents<uint8_t>::TypedArrayContents(v8::Local<v8::Value>)'
          ..\cpp\typedarrays.cpp(16) : see reference to function template instantiation 'Nan::TypedArrayContents<uint8_t>::TypedArrayContents(v8::Local<v8::Value>)' being compiled
          ..\cpp\typedarrays.cpp(16) : see reference to class template instantiation 'Nan::TypedArrayContents<uint8_t>' being compiled
c:\projects\nan\nan_typedarray_contents.h(67): error C3861: 'alignof': identifier not found [C:\projects\nan\test\build\typedarrays.vcxproj]
```

mikolalysenko:
Ok, I took out `alignof` since it seems that doesn't work on windows.  We'll see if the new version passes.

kkoopa:
Windows should have `__alignof`.

mikolalysenko:
Alright, I added in some more precise alignments checks for various platforms.

mikolalysenko:
@kkoopa can you start the appveyor CI?  It looks like the node v0.12 case failed due to a network problem during the npm install.

kkoopa:
Can't. Push something to make it run.

mikolalysenko:
Ok, I pushed a doc tweak and it passed this time.

nan_typedarray_contents.h
```diff
@@ -0,0 +1,89 @@
...
+
+#if defined(_MSC_VER)
+    assert(reinterpret_cast<uintptr_t>(data_) % __alignof(T) == 0);
+#elif defined(__GNUC__)
+    assert(reinterpret_cast<uintptr_t>(data_) % __alignof__(T) == 0);
+#elif NODE_MODULE_VERSION >= IOJS_3_0_MODULE_VERSION
+    assert(reinterpret_cast<uintptr_t>(data_) % alignof(T) == 0);
+#else
+    assert(reinterpret_cast<uintptr_t>(data_) % sizeof(T) == 0);
+#endif
```
bnoordhuis: I'd hoist this to the top of the constructor and replace the NODE_MODULE_VERSION check with `#if __cplusplus >= >= 201103L` (and make it the first one.)

PS: Both clang and gcc have `__alignof(...)` so you should be able to fold the gcc and VS code paths.

mikolalysenko: If we hoist it to the top of the constructor, then `data_` won't be initialized.  It is probably better to check the assert after we computed the byteOffset and finished all the unboxing logic.

bnoordhuis: Yes, that's what I mean; the assert should precede assigning to `data_`.

mikolalysenko: So, what you would want is to maybe read data_ and length_ into temporary variables, then write the result only if the assertion passes?

mikolalysenko:
Is this PR ok?  I would like to use it in headless-gl so that it can support node v4.  If the commit log is too messy I can do a rebase and squish it into a new PR.

kkoopa:
Would it not be more appropriate to have it return a `Maybe<T>` instead of possibly returning `NULL`. It seems a better fit with the rest that way.

mikolalysenko:
@kkoopa I'm not sure what the benefit of wrapping everything in a `Maybe<>` would be, besides making things more complicated.  `Utf8String` doesn't do this, and I think there is a strong parallel between that usage and the situation here.

kkoopa:
The difference is that Utf8String cannot return `NULL`. The point of wrapping everything in Maybes is to get rid of `NULL`.

mikolalysenko:
Well, why stop there.  We could even add some overloads for `const` and all permutations of `Maybe<>` as an argument:

```c++
template<typename T> Nan::Maybe< Nan::TypedArrayContents<T> > getTypedArrayContents(Nan::Maybe< Nan::Local<Value> > &object);

template<typename T> Nan::Maybe< const Nan::TypedArrayContents<T> > getTypedArrayContents(const Nan::Maybe< Nan::Local<Value> > &object);

template<typename T> Nan::Maybe< Nan::TypedArrayContents<T> > getTypedArrayContents(Nan::Local<Value> &object);

template<typename T> Nan::Maybe< const Nan::TypedArrayContents<T> > getTypedArrayContents(const Nan::Local<Value> &object);
```

mikolalysenko:
Alternatively, we could maybe use a different interface.  Perhaps the following approach would be simpler:

```c++
template<typename T>
  bool Nan::getTypedArrayContents(Nan::Local<Value> array, size_t& length, T*& data);
```

Where this method returns `false` if the object is not a valid typedarray.

jfirebaugh:
@kkoopa Seems to me that `Utf8String` does provide a precent for using NULL. From [the docs](https://github.com/nodejs/nan/blob/master/doc/v8_misc.md#nanutf8string):

> If conversion to a string fails (e.g. due to an exception in the toString() method of the object) then the length() method returns 0 and the * operator returns NULL.

I'm :+1: on merging this in the current state.

mikolalysenko:
I'm ok with changing the interface too, just let me know how you want it to look.

kkoopa:
@jfirebaugh Good point, although the documentation might not be accurate here. At least I don't remember enforcing it anywhere: https://github.com/nodejs/nan/blob/master/nan.h#L861-L901

Who knows when V8 will change `Utf8Value` to return a `Maybe`? Seems like it should be changed, to make it consistent with the new ways, but that is a different matter.

I don't have particularly strong feelings either way, it's mostly caution with what will be the best way of introducing a new API and in what form, as there will be no option for take-backs until the next major version.

Perhaps @bnoordhuis could weigh in on this?

mikolalysenko:
Hold the press on this one, I found a failing case.  On node v0.12, it seems that there is some optimization where typed arrays aren't allocated until they are written to.  I'm not sure what is causing this problem, but it is not present in earlier versions and does not fail in node v4.

mikolalysenko:
Figured out a workaround.

mikolalysenko:
I think the CI failed due to a network issue.  If you want I can poke this PR with some doc tweak to rerun it.

kkoopa:
Poke ahead. There are occasional network issues.

mikolalysenko:
I'm not sure what is causing the problems on node v0.8, as far as I know nothing should have changed.  It looks like travis is failing to install the module `are-we-there-yet`.  Locally, I can run the tests against 0.8 and everything passes with no problems.

kkoopa:
Could it be that the travis config does not update npm for node 0.8? If it works locally it is probably good... 0.8 can act flaky with npm sometimes. I've seen that before I think.

mikolalysenko:
Yeah, it is just node v0.8 on CI.  Everything works fine locally.

LinusU:
@mikolalysenko If you want to trigger ci without adding more commits you can change the latest one instead.

```sh
git commit --amend --no-edit
```

Then force push to overwrite it

```sh
git push --force
```

Just a small tip :)

kkoopa:
Seems npm has been updated to 3.3 and they broke that damn are-we-there-yet thing, *again*. Why can't they stop touching it or simply remove it completely?

bsergean:
Should an issue be created in are-we-there-yet, or in npm ?

https://github.com/iarna/are-we-there-yet/issues

tmpvar:
:+1:, what can I do to help get this landed?

rvagg:
https://github.com/npm/npm/search?q=are-we-there-yet&type=Issues&utf8=%E2%9C%93 - there's plenty of issues & discussion about the npm problem

mikolalysenko:
I removed the extra const qualifiers on the value types returned by the accessors, since they aren't needed and they also generate warnings when compiling with GCC on Linux.

kkoopa:
It should have both versions of the operators. Both the const- and non-const versions.

mikolalysenko:
The issue is not making the contents of the pointer const, but rather making the pointer itself const.  Returning a const pointer isn't needed since you can freely cast it to a non-const pointer.  While doing this would have no semantic effect, it does trigger a warning in gcc on Linux (but not in clang on OS X) which you would have to ignore.

Forcing users to ignore warnings seems like it could cause problems, especially if you are building with -Werror.

kkoopa:
Ah, yes. I did not read the diff thoroughly.

kkoopa:
If you rebase and squash, I guess it is good to merge.

mikolalysenko:
Closing this PR, moved all the changes into a separate branch.

Issue #462, opened 2015-09-15
PropertyCallbackInfo data pointer gets corrupted


jjrv:
This only affects NAN with Node 0.10 on Windows and somehow not on Linux.

When using `Nan::SetAccessor`, the 5th argument (arbitrary data) gets corrupted inside `GetterCallbackWrapper` [on this line](https://github.com/nodejs/nan/blob/40754451bb304d7b78ffdecfd9c8b6e5e56b4093/nan_callbacks_pre_12_inl.h#L305) before reaching the callback.

[Reason is a stray `&` here](https://github.com/nodejs/nan/blob/40754451bb304d7b78ffdecfd9c8b6e5e56b4093/nan_callbacks_pre_12_inl.h#L152). Compare to [another similar line that is correct](https://github.com/nodejs/nan/blob/40754451bb304d7b78ffdecfd9c8b6e5e56b4093/nan_callbacks_pre_12_inl.h#L103).

PR #463, opened 2015-09-15
Fix issue #462 where PropertyCallbackInfo data is not stored safely.

jjrv:


kkoopa:
LGTM if CI is happy.

nan_callbacks_pre_12_inl.h
```diff
@@ -149,7 +149,7 @@ class FunctionCallbackInfo {
 template<typename T>
 class PropertyCallbackInfoBase {
   const v8::AccessorInfo &info_;
-  const v8::Local<v8::Value> &data_;
+  const v8::Local<v8::Value> data_;
 
  public:
   explicit inline PropertyCallbackInfoBase(
```
kkoopa: Update the constructor here to take a Local and not a reference. Same goes for superclasses.

jjrv:
Better like this? There's still some Set and Reset functions remaining that take a reference to v8::Local. But I didn't find any cases where it would store such a reference except that single first one.

kkoopa:
It looks fine at a glance. These constructors should never be directly used by user code anyway, so changing them is fine. Set and Reset functions should take references, as that is (supposed to be) part of the emulated V8 API.

kkoopa:
Merged as 55f50adedd543098526c7b9f4fffd607d3f9861f fefd813d432a5368a87da61eae48587b0f9e874b

PR #477, opened 2015-09-30
Add NODE_4_00_MODULE_VERSION definition

brodybits:


brodybits:
This change should be applied in a different location due to #450.

Issue #466, opened 2015-09-17
The documentation for Nan::New is inconsistent


murgatroid99:
It says

> The following types will be contained within a `Nan::MaybeLocal<T>`: `v8::String`, `v8::Date`, `v8::RegExp`, `v8::Script`, `v8::UnboundScript`.

but the signatures given are

```cpp
v8::Local<v8::String> Nan::New<T>(std::string const& value);
v8::Local<v8::String> Nan::New<T>(const char * value, int length);
v8::Local<v8::String> Nan::New<T>(const char * value);
v8::Local<v8::String> Nan::New<T>(const uint16_t * value);
```

kkoopa:
Will fix.

brodybits:
I am not so sure that the documentation is wrong. Looking at the following sections:

[nan_new.h#L42-L44](./nan_new.h#L42-L44):
```c++
template <typename T> struct MaybeFactoryBase {
  typedef MaybeLocal<T> return_t;
};
```
[nan_new.h#L158-L172](./nan_new.h#L158-L172):
```c++
template <>
struct Factory<v8::String> : MaybeFactoryBase<v8::String> {
  static inline return_t New();
  static inline return_t New(const char *value, int length = -1);
  static inline return_t New(const uint16_t *value, int length = -1);
  static inline return_t New(std::string const& value);

  static inline return_t New(v8::String::ExternalStringResource * value);
  static inline return_t New(ExternalOneByteStringResource * value);
};
```
and [nan_new.h#L290-L300](./nan_new.h#L290-L300) (and further):
```c++
inline
imp::Factory<v8::String>::return_t
New(std::string const& value) {  // NOLINT(build/include_what_you_use)
  return New<v8::String>(value);
}

inline
imp::Factory<v8::String>::return_t
New(const char * value, int length) {
  return New<v8::String>(value, length);
}
```

I would conclude that the `v8::String` variants of `Nan::New<T>` also return `v8::MaybeLocal<T>` and *not* `v8::Local<T>`.

As an experiment, I tried building with `Factory<v8::String>` derived from `FactoryBase<v8::String>` instead of `MaybeFactoryBase<v8::String>` but encountered some errors since certain V8 functions are returning `MaybeLocal<v8::String>`.

murgatroid99:
[...]

PR #480, opened 2015-10-02
doc: fix for string variants of Nan::New

brodybits:
Fixed the inconsistency reported in #466, according to the API in `nan_new.h`.

Also removed trailing whitespaces.


kkoopa:
Thanks, LGTM.

kkoopa:
Merged as 0a2af1e4f0d9acf18f157b3c08dcece5b17374ca

PR #489, opened 2015-10-10
Nogcmacro

kkoopa:
Tried to remove the dependence on macros.
The gc callback thing feels a bit clumsy.


nan.h
```diff
@@ -136,15 +136,77 @@ namespace Nan {
...
+#if NODE_MODULE_VERSION < NODE_0_10_MODULE_VERSION
+namespace imp {
+  template<addon_reg_func_t F>
+  static inline void addon_reg_func(v8::Handle<v8::Object> target) {
+    v8::HandleScope scope;
+    F(v8::Local<v8::Object>::New(target));
+  }
+}
+
+# define NAN_MODULE(modname, regfunc)                                          \
```
agnat: While we are at it, we could DRY up and drop `modname` all together. Newer versions of node-gyp define [`NODE_GYP_MODULE_NAME`](https://github.com/nodejs/node-gyp/blob/master/addon.gypi#L14). For older versions it is easily emulated since we have a `.gypi` file.

Just a thought, though.

nan.h
```diff
@@ -136,15 +136,77 @@ namespace Nan {
...
+  template<addon_reg_func_t F>
+  static inline void addon_reg_func(v8::Handle<v8::Object> target) {
+    v8::Isolate *isolate = v8::Isolate::GetCurrent();
+    v8::HandleScope scope(isolate);
+    F(v8::Local<v8::Object>::New(isolate, target));
+  }
+}
+
+# define NAN_MODULE(modname, regfunc)                                          \
+  extern "C" {                                                                 \
```
agnat: Do we have to duplicate this macro? Isn't it enough to call it with `Nan::imp::addon_reg_func<regfunc>` as an argument?

agnat:
> The gc callback thing feels a bit clumsy.

I see you pulled the direct call trick ... other than that. Please elaborate...

nan.h
kkoopa: Ah, that may be the case. But it looks so ugly.

nan.h
agnat: > ... since we have a .gypi file.

Oh, we don't... I believe I proposed this once but it didn't work out. In any case _not_ repeating _modname_ would be nice to have.

nan.h
agnat: Really?

````
#define NAN_MODULE(modname, regfunc) \
  NODE_MODULE(modname, Nan::imp::addon_reg_func<regfunc>)
````

... looks ok to me. Users won't even see it.

kkoopa: Ah, that way it is fine, yes.

kkoopa:
I would like to get the callback as a function argument, like in all other parts of the API, but that is not possible. This will remain a slightly odd case.

nan.h
kkoopa: don't know how to achieve it

agnat: Hehe... yeah... by _introducing_ our own `.gypi`. So it probably is something for next major.

kkoopa:
Yeah, this is all mostly for the next major, whenever that is. Not worth breaking compatibility over these small things alone. Just figured I'd explore the possibilities. This seems to be the best way of defining callbacks when no arbitrary data arguments are available.

kkoopa:
The ugliness could be avoided by using a static map of gc callbacks, but I don't know if it is worth it.

agnat:
Hm, yeah... a homegrown v-table... not sure... in my book that would be even more ugly as it introduces a call overhead of O(n log(n)) (I think) with n being the number of NAN based modules. Not that anyone would notice but I like the template version better...

agnat:
BTW, I don't think it's ugly... it's merely a bit inconsistent. Actually I think I like it. The pattern has interesting properties.

kkoopa:
This would only be for the gc callbacks. There n would grow with the number of gc callbacks. How likely are you to use many of such an obscure item? Inconsistency is arguably ugliness, but I prefer this over macros. C++11 lifts a lot of these restrictions, but we can't use that everywhere yet, which is kind of limiting.

PR #491, opened 2015-10-11
Basic documentation of object wrappers on its own page, with sample code

brodybits:
Keeps documentation of `Nan::ObjectWrapper` on its own page, along with examples to reflect the documentation at https://nodejs.org/api/addons.html:
- basic wrapped object
- wrapped object factory
- passing wrapped objects around

The examples are derived from the existing tests, with some adaptations and `using namespace` statement(s) removed. I tested the C++ code in the examples (with all `using namespace` statement(s) removed) to be sure that they will compile and run correctly. To be honest they are not so easy to understand for a beginner but I figure they would be better than having no samples.

### DIFF
--- a/doc/object_wrappers.md
+++ b/doc/object_wrappers.md
@@ -1,0 +1,248 @@
+## Object Wrappers
+
+ - <a href="#api_nan_object_wrap"><b><code>Nan::ObjectWrap</code></b></a>
+
+
+<a name="api_nan_object_wrap"></a>
+### Nan::ObjectWrap()
+
+A reimplementation of `node::ObjectWrap` that adds some API not present in older versions of Node. Should be preferred over `node::ObjectWrap` in all cases for consistency.
+
+Definition:
+
+```c++
+class ObjectWrap {
+ public:
+  ObjectWrap();
+
+  virtual ~ObjectWrap();
+
+  template <class T>
+  static inline T* Unwrap(v8::Local<v8::Object> handle);
+
+  inline v8::Local<v8::Object> handle();
+
+  inline Nan::Persistent<v8::Object>& persistent();
+
+ protected:
+  inline void Wrap(v8::Local<v8::Object> handle);
+
+  inline void MakeWeak();
+
+  /* Ref() marks the object as being attached to an event loop.
+   * Refed objects will not be garbage collected, even if
+   * all references are lost.
+   */
+  virtual void Ref();
+
+  /* Unref() marks an object as detached from the event loop.  This is its
+   * default state.  When an object with a "weak" reference changes from
+   * attached to detached state it will be freed. Be careful not to access
+   * the object after making this call as it might be gone!
+   * (A "weak reference" means an object that only has a
+   * persistant handle.)
+   *
+   * DO NOT CALL THIS FROM DESTRUCTOR
+   */
+  virtual void Unref();
+
+  int refs_;  // ro
+};
+```
+
+See the Node documentation on [Wrapping C++ Objects](https://nodejs.org/api/addons.html#addons_wrapping_c_objects) for more details.
+
+### Examples
+
+#### Basic
+
+```c++
+class MyObject : public Nan::ObjectWrap {
+ public:
+  static NAN_MODULE_INIT(Init) {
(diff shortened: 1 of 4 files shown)


doc/object_wrappers.md
```diff
@@ -0,0 +1,247 @@
...
+Use the `MyFactoryObject` class above along with the following:
+
+```c++
+static NAN_METHOD(Sum) {
+  Nan::MaybeLocal<v8::Object> maybe1 = Nan::To<v8::Object>(info[0]);
+  Nan::MaybeLocal<v8::Object> maybe2 = Nan::To<v8::Object>(info[1]);
+
+  // Quick check:
+  if (maybe1.IsEmpty() || maybe2.IsEmpty()) {
+    info.GetReturnValue().Set(Nan::Undefined());
```
kkoopa: The default return value is undefined, so this whole line is unnecessary. On another note, it may also be slightly better to use the primitive setters directly e.g. `ReturnValue::SetUndefined`.

doc/object_wrappers.md
```diff
@@ -0,0 +1,247 @@
...
+    Nan::Set(target, Nan::New("MyObject").ToLocalChecked(), tpl->GetFunction());
+  }
+
+ private:
+  explicit MyObject(double value = 0) : value_(value) {}
+  ~MyObject() {}
+
+  static NAN_METHOD(New) {
+    if (info.IsConstructCall()) {
+      double value = info[0]->IsUndefined() ? 0 : info[0]->NumberValue();
```
kkoopa: Prefer `Nan::To<double>` over `NumberValue`

doc/object_wrappers.md
```diff
@@ -0,0 +1,247 @@
...
+ public:
+  static NAN_MODULE_INIT(Init) {
+    v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(New);
+    tpl->SetClassName(Nan::New("MyObject").ToLocalChecked());
+    tpl->InstanceTemplate()->SetInternalFieldCount(1);
+
+    SetPrototypeMethod(tpl, "getHandle", GetHandle);
+
+    constructor().Reset(tpl->GetFunction());
+    Nan::Set(target, Nan::New("MyObject").ToLocalChecked(), tpl->GetFunction());
```
kkoopa: `Nan::GetFunction(tpl).ToLocalChecked()`

kkoopa:
This is good to merge after those small nits have been fixed.

brodybits:
I made the changes and also made very similar changes to `test/cpp/objectwraphandle.cpp` to make it more future-proof as well.

doc/object_wrappers.md
```diff
@@ -0,0 +1,248 @@
...
+ public:
+  static NAN_MODULE_INIT(Init) {
+    v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(New);
+    tpl->SetClassName(Nan::New("MyObject").ToLocalChecked());
+    tpl->InstanceTemplate()->SetInternalFieldCount(1);
+
+    SetPrototypeMethod(tpl, "getHandle", GetHandle);
+
+    constructor().Reset(Nan::GetFunction(tpl).ToLocalChecked());
+    Set(target, Nan::New("MyObject").ToLocalChecked(),
```
kkoopa: This should have the namespace explicit, as the rest of the example does.

doc/object_wrappers.md
```diff
@@ -0,0 +1,248 @@
...
+#### Factory of wrapped objects
+
+```c++
+class MyFactoryObject : public Nan::ObjectWrap {
+ public:
+  static NAN_MODULE_INIT(Init) {
+    v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(New);
+    tpl->InstanceTemplate()->SetInternalFieldCount(1);
+
+    SetPrototypeMethod(tpl, "getValue", GetValue);
```
kkoopa: namespace here too

brodybits:
My bad! I just fixed lines 70 and 131 to use the `Nan` namespace, and also added a comment at line 212 to mark the case where the return value is `undefined` by default.

kkoopa:
Very good. Thanks a lot.

PR #492, opened 2015-10-11
Add convenience getter for Local from Persistent

kkoopa:
This is found in newer versions of V8.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -169,15 +169,12 @@ typedef v8::String::ExternalOneByteStringResource
 
 #if (NODE_MODULE_VERSION > NODE_0_10_MODULE_VERSION)
 template<typename T>
-class NonCopyablePersistentTraits :
-    public v8::NonCopyablePersistentTraits<T> {};
+class NonCopyablePersistentTraits;
 template<typename T>
-class CopyablePersistentTraits :
-    public v8::CopyablePersistentTraits<T> {};
+class CopyablePersistentTraits;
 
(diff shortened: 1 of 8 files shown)


rvagg:
Pretty big set of changes, I won't have time to review and am probably not the best for this one anyway, maybe @bnoordhuis or @agnat if they have time.

Missing docs though, is it just `Get()` that we want to add or should more of this be officially documented, (like `.persistent`?)

kkoopa:
Just Get, the rest is private, so no point in documenting anything else, since it is not accessible. There were many changes, since I redid the newer implementation not to inherit from v8::PeristentBase<> but to compose it.

brodybits:
I wonder if we can simply add `Get` to `Nan::Persistent` and `Nan::Global` instead? This would preserve `Nan::PersistentBase` and its subclasses as subclasses of `v8::PersistentBase` FWIW.

kkoopa:
Yes, that would work, but I wanted to try and redo the implementation to make 
it more similar to the other one. Now the hierarchy is the same across 
versions, as would be expected.

brodybits:
Sounds reasonable to me. There is an assumption that the user would never need access to the `v8::PersistentBase` object, hope this will remain true.

kkoopa:
That assumption is already there for versions less than 0.12, this makes it 
explicit in newer versions too. v8::PersistentBase cannot even be directly 
instantiated.

PR #494, opened 2015-10-14
Adding DCO and CoC

mikeal:


rvagg:
Initial agreement of the group when deciding to move to the nodejs org was to not adopt a CoC, which at the time was within the rules as long as we'd discussed it.

rvagg:
Would like to hear from @kkoopa, @bnoordhuis and @agnat on this PR, perhaps they have no objections or can see a sensible path forward and it's just me that's causing trouble by even raising possible objections.

bnoordhuis:
We have a DCO, it's in the README.  I don't particularly miss a CoC.

kkoopa:
Ditto. Common sense should be enough.

agnat:
I don't particularly miss a CoC either, but then members of the majority never do since they never feel excluded. So, I'm currently mildly in favor of merging this and be done with it. 

At the very least we should add one line about what to do if you feel harassed/excluded/&c.

mikeal:
> which at the time was within the rules as long as we'd discussed it.

Technically I don't think this was ever the policy, although you could argue that the policy has failed. Let me explain.

The process for WG's was documented to be "copy the contribution policy (which includes the CoC) to your WG." Of course we had no good language in there for individuals projects under WGs or projects that transfer into the project. But even in the WGs that were created almost none of them copied over those agreements and we've never effectively kept track. I wrote the policy and I don't think I've ever copied over these documents in the 40+ repos I've created here.

This is all to say that the process we designed hasn't worked and that copying or requiring writing a new policy in ever repo/WG is ineffective and we should find a better solution.

However, for legal reasons we do need the DCO to be in *every* repo and we actually need it to be in a CONTRIBUTING.md file because that's the file GitHub points users to when filing a pull request.

I'm starting to think a minimalist CoC that is foundation wide is a better approach to all of this. Several other foundations have done this and, because violations are rare, it's a nice service to have something to fall back on rather than having an up-front requirement we make every project and WG deal with.

kkoopa:
My specific concerns are:
 - No politics
 - No agendas
 - No whining

kkoopa:
Why did you close it? My point was that we are volunteering our spare time here for solving technical issues, I imagine none of us really wishes to deal with any extra drama.

If, say, a pull request were to be rejected based on merit, or lack thereof, I do not wish someone with hostile intentions to have fodder for suing me or hanging me out to dry based on hurt feelings. The current node.js code of conduct does not have safeguards for not abusing the code of conduct or harassing contributors with inane issues and imagined slights. This is what I meant with the above three bullet points. If they could be incorporated into the minimalist CoC you mentioned, I would probably not have further reservations.

The DCO is fine and can be put in CONTRIBUTING.md if it is more appropriate than having it in README.md.

mikeal:
I closed it because this PR won't be merged as is and I need to send a new one moving the DCO to the CONTRIBUTING file.

> I imagine none of us really wishes to deal with any extra drama.

If you continue this line of commentary you will certainly attract drama, another reason I closed and locked it as this is the predictable outcome of this thread were it to continue.

PR #496, opened 2015-10-15
Add Function::Call wrapper

kkoopa:


joshperry:
I'm using this code in a project and somewhere between v0.10 and v0.12 the `recv` function was changed from `Object` to `Value` so I'm going to change the parameter type in the pre_46 ifdef in my code.

I'm not familiar enough with the progression of v8/node to know. Do you know if this ifdef needs to be modified slightly to build properly on v0.12?

kkoopa:
No, `Object` is a descendant of `Value`. I've updated the branch.

PR #499, opened 2015-10-21
Allow Value as target for function calls

### LINKED ISSUE #497: Receiver parameter for CallAsFunction should be Local<Value>
The receiver can be null, so the receiver parameter for `Nan::CallAsFunction` should be `Local<Value>` not `Local<Object>`.  The parameter is declared as `Local<Value>` in v8, it's just Nan which has the overly-restrictive type.

kkoopa:
Fixes #497

R: @bnoordhuis

Also, Node should be fixed to allow `v8::Value` in `node::MakeCallback`.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -806,17 +806,23 @@ class TryCatch {
     , int argc
     , v8::Local<v8::Value>* argv) {
 #if NODE_MODULE_VERSION < IOJS_3_0_MODULE_VERSION
-    v8::Local<v8::Object> target_ = target->IsUndefined() || target->IsNull() ?
-        v8::Isolate::GetCurrent()->GetCurrentContext()->Global() :
-            target->ToObject();
+    v8::Local<v8::Object> target_obj;
+    if (target->IsUndefined() || target->IsNull()) {
+      target_obj = v8::Isolate::GetCurrent()->GetCurrentContext()->Global();
+    } else {
+      target_obj = target->ToObject();
+    }
     return New(node::MakeCallback(
-        v8::Isolate::GetCurrent(), target_, func, argc, argv));
+        v8::Isolate::GetCurrent(), target_obj, func, argc, argv));
 #else
-    v8::Local<v8::Object> target_ = target->IsUndefined() || target->IsNull() ?
-        v8::Isolate::GetCurrent()->GetCurrentContext()->Global() :
-            Nan::To<v8::Object>(target).ToLocalChecked();
+    v8::Local<v8::Object> target_obj;
+    if (target->IsUndefined() || target->IsNull()) {
+      target_obj = v8::Isolate::GetCurrent()->GetCurrentContext()->Global();
+    } else {
+      target_obj = Nan::To<v8::Object>(target).ToLocalChecked();
+    }
     return node::MakeCallback(
-        v8::Isolate::GetCurrent(), target_, func, argc, argv);
+        v8::Isolate::GetCurrent(), target_obj, func, argc, argv);
 #endif
   }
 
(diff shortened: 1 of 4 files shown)


nan.h
```diff
@@ -801,44 +801,62 @@ class TryCatch {
...
   }
 
   NAN_INLINE v8::Local<v8::Value> MakeCallback(
-      v8::Local<v8::Object> target
+      v8::Local<v8::Value> target
     , v8::Local<v8::String> symbol
     , int argc
     , v8::Local<v8::Value>* argv) {
 #if NODE_MODULE_VERSION < IOJS_3_0_MODULE_VERSION
+    v8::Local<v8::Object> target_ = target->IsUndefined() || target->IsNull() ?
```
bnoordhuis: target_obj?  Ending in an underscore suggests it's a member.

bnoordhuis:
The ternaries are making my eyeballs bleed but LGTM.

nan.h
kkoopa: I see it as privatish, but you are right. However, that would make the line too long so I would have to use if-else like a chump...

bnoordhuis: My eyeballs would thank you for it.

kkoopa: Yeah, I'll fix it if you fix `node::MakeCallback` in time for 5.0.

bnoordhuis: There was a reason I couldn't change it in io.js but I forgot what it was...

kkoopa: That seems odd, as this seems to work :/

bnoordhuis: I remember now: node.js looks up the object's context with `recv->CreationContext()`.  That doesn't work when recv is a Value.

kkoopa:
Fixed the ugliness. Is this still good to go regarding MakeCallback?

bnoordhuis:
LGTM with the caveat that the handles from the Global() and ToObject() calls will leak into the handle scope of the caller.  If that is bad, I'd add HandleScopes / EscapableHandleScopes.

kkoopa:
But, but... They should not be creating _new_ handles... I'll add the scopes immediately.

kkoopa:
That should do it.

bnoordhuis:
LGTM

kkoopa:
Thanks for taking the time to look through this. I'll flatten on merge.

nan.h
```diff
@@ -801,44 +801,83 @@ class TryCatch {
...
     , int argc
     , v8::Local<v8::Value>* argv) {
+    v8::Isolate *isolate = v8::Isolate::GetCurrent();
+    v8::EscapableHandleScope scope(isolate);
+    v8::Local<v8::Object> target_obj;
 #if NODE_MODULE_VERSION < IOJS_3_0_MODULE_VERSION
-    return New(node::MakeCallback(
-        v8::Isolate::GetCurrent(), target, func, argc, argv));
+    if (target->IsUndefined() || target->IsNull()) {
+      target_obj = isolate->GetCurrentContext()->Global();
```
trevnorris: In strict mode wouldn't devs expect the calling context to be `undefined`?

kkoopa: Seems you are right according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode (unless something has changed in ES 6, but then we should probably go with ES 5 behavior).

Is there a way of knowing if a function is strict or not from V8?

kkoopa: Here's the 4.6 implementation of `Function::Call`, it seems it explicitly converts to an (internal) `Object` using `Utils::OpenHandle`.
https://github.com/nodejs/node/blob/master/deps/v8/src/api.cc#L4387

`Utils::OpenHandle` is here
https://github.com/nodejs/node/blob/master/deps/v8/src/api.h#L390-L398

But I don't really get how that `->IsX` check would pass. Are `v8::Undefined` and `v8::Null` both represented as a `nullptr` inside a `Local<>`?

Would it be possible to just hack this by doing a `reinterpret_cast<v8::Local<v8::Object>*>(&target)` and get V8 to do the right thing?

bnoordhuis: Not directly but you can deduce it by trying to access the `function.caller` property, it throws an exception for strict functions.

trevnorris: I believe the `val_` pointer is, but internally the`Value` has a check to see what the type is. So it can determine that way.

kkoopa: @bnoordhuis  Dealing with exceptions for every function call sounds like it would be quite expensive, no?

Here's the same line for `Function::Call` in V8 from 0.8.28
https://github.com/nodejs/node/blob/v0.8.28-release/deps/v8/src/api.cc#L3639

And here's `Utils::OpenHandle`
https://github.com/nodejs/node/blob/v0.8.28-release/deps/v8/src/api.h#L297-L302
There does not seem to be a type check there... I guess the only way of finding out if that crude hack I suggested works is by testing., right?

kkoopa: Doing a straight cast led to a segfault. Can't access `function.caller`, because I don't have access to the current closure from within `Nan::MakeCallback`. Even if I could, it would still not be possible to make target `Undefined`, since `node::MakeCallback` only accepts `Object`.

I did some testing, in strict mode, doing `Function::Call` with `Undefined` as target produced `undefined`, while setting `Null` as target produced `null`. In normal mode, both produce `global`.

All in all, it seems the current behavior is the only one that can be consistently implemented and this PR can be closed. No `Value` targets for anyone.

PR #502, opened 2015-10-26
Add prototype method

noamokman:
Hi there!
I added a missing method on the prototype.
Looks like you need to register all functions.

Thanks!


brodybits:
Good catch, that was my bad in PR #491.

PR #505, opened 2015-10-28
Use NAN_HAS_CPLUSPLUS_11 more

kkoopa:


nan_typedarray_contents.h
```diff
@@ -57,19 +57,19 @@ class TypedArrayContents {
 
 #endif
 
-#if defined(_MSC_VER) || defined(__GNUC__)
+#if defined(_MSC_VER) && _MSC_VER >= 1900 || __cplusplus >= 201103L
```
rvagg: shouldn't this use `NAN_HAS_CPLUSPLUS_11`?

kkoopa: Heh, that's what I thought too, which is why I started messing around with it. However, Visual Studio does not support the `alignof` operator prior to VS 2015. `NAN_HAS_CPLUSPLUS_11` tests for VS 2013, as that is sufficient to compile V8.

kkoopa: However, I see I could break this up into two commits to make things more logical.

PR #510, opened 2015-11-04
Fixed package.json read

CKarper:
You probably meant to read the tool's package.json, not whichever package.json happens to be in CWD.


kkoopa:
Thanks,  but do note that said tool is not actively maintained. However, https://github.com/nodesource/upgrade-utils is.

PR #514, opened 2015-11-08
AsyncProgressWorker callback context isolation fix

bfrancojr:
The current AsyncProgressWorker class implementation has one single pointer to define the callback context (asyncdata_ member var) which can be overwritten in subsequents calls to progress.Send method even before the callback to be called.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1617,40 +1617,40 @@ class Callback {
    public:
     // You could do fancy generics with templates here.
     void Send(const char* data, size_t size) {
-        asyncdata_t* asyncdata = new asyncdata_t;
-        asyncdata->data = new char[size];
-        memcpy(asyncdata->data, data, size);
-        asyncdata->size = size;
-        asyncdata->handle = new uv_async_t;
-        asyncdata->worker = that_;
-        uv_async_init(
-            uv_default_loop()
-            , asyncdata->handle
-            , AsyncProgress_
-        );
-        asyncdata->handle->data = asyncdata;
-        uv_async_send(asyncdata->handle);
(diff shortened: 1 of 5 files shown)


kkoopa:
@bnoordhuis Can you review this? I have a hard time figuring out if it is safe.

nan.h
```diff
@@ -1592,51 +1592,64 @@ class Callback {
...
     // You could do fancy generics with templates here.
     void Send(const char* data, size_t size) const {
-        that_->SendProgress_(data, size);
+      asyncdata_t* asyncdata = new asyncdata_t;
+      asyncdata->data = new char[size];
+      memcpy(asyncdata->data, data, size);
+      asyncdata->size = size;
+      asyncdata->handle = new uv_async_t;
+      asyncdata->worker = that_;
+      uv_async_init(
```
bnoordhuis: This is unsound.  `uv_async_init()` needs to be called on the event loop thread (the main thread), only `uv_async_send()` is thread-safe.

bnoordhuis:
I'd stick with the current approach but add a `std::deque` or something like that.

Keep in mind that libuv can and will coalesce `uv_async_send()` calls so you can't assume that one call to `uv_async_send()` means one item popped off the queue.

bfrancojr:
Hi @bnoordhuis,

Could you please review my changes? I'm using std::deque now.

Thanks!

bnoordhuis:
@bfrancojr Can you rebase your work?  It contains unrelated changes.

bfrancojr:
I'm going to submit a new PR. Please ignore this one.
Thanks.

PR #515, opened 2015-11-10
AsyncProgressWorker callback context isolation fix (reviewed)

bfrancojr:
The current AsyncProgressWorker class implementation has one single pointer to define the callback context (asyncdata_ member var) which can be overwritten in subsequents calls to progress.Send method even before the callback to be called.
This PR follows @bnoordhuis suggestion and uses std::deque to temporarily store async data.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -49,6 +49,7 @@
 #include <cstring>
 #include <climits>
 #include <cstdlib>
+#include <deque>
 #if defined(_MSC_VER)
 # pragma warning( push )
 # pragma warning( disable : 4530 )
@@ -1591,39 +1592,38 @@ class Callback {
 
 /* abstract */ class AsyncProgressWorker : public AsyncWorker {
  public:
+
+  typedef std::deque<std::string> async_queue_t;
+
   explicit AsyncProgressWorker(Callback *callback_)
-      : AsyncWorker(callback_), asyncdata_(NULL), asyncsize_(0) {
-    async = new uv_async_t;
+      : AsyncWorker(callback_) {
     uv_async_init(
         uv_default_loop()
-      , async
+      , &async
       , AsyncProgress_
     );
-    async->data = this;
+    async.data = this;
 
     uv_mutex_init(&async_lock);
   }
 
   virtual ~AsyncProgressWorker() {
     uv_mutex_destroy(&async_lock);
-
-    if (asyncdata_) {
-      delete[] asyncdata_;
-    }
   }
 
   void WorkProgress() {
-    uv_mutex_lock(&async_lock);
-    char *data = asyncdata_;
-    size_t size = asyncsize_;
-    asyncdata_ = NULL;
-    uv_mutex_unlock(&async_lock);
-
     // Dont send progress events after we've already completed.
     if (callback) {
-        HandleProgressCallback(data, size);
+      uv_mutex_lock(&async_lock);
+      while (!queue.empty()) {
+        const std::string& async_data = queue.front();
+        uv_mutex_unlock(&async_lock);
+        HandleProgressCallback(async_data.data(), async_data.size());
+        uv_mutex_lock(&async_lock);
+        queue.pop_front();
+      }
+      uv_mutex_unlock(&async_lock);
     }
-    delete[] data;
   }
 
   class ExecutionProgress {
@@ -1631,7 +1631,7 @@ class Callback {
    public:
     // You could do fancy generics with templates here.
     void Send(const char* data, size_t size) const {
-        that_->SendProgress_(data, size);
+      that_->SendProgress_(data, size);
     }
 
    private:
@@ -1644,29 +1644,20 @@ class Callback {
   virtual void HandleProgressCallback(const char *data, size_t size) = 0;
 
   virtual void Destroy() {
-      uv_close(reinterpret_cast<uv_handle_t*>(async), AsyncClose_);
+    uv_close(reinterpret_cast<uv_handle_t*>(&async), AsyncClose_);
   }
 
  private:
   void Execute() /*final override*/ {
-      ExecutionProgress progress(this);
-      Execute(progress);
+    ExecutionProgress progress(this);
+    Execute(progress);
   }
 
   void SendProgress_(const char *data, size_t size) {
-    char *new_data = new char[size];
-    memcpy(new_data, data, size);
-
     uv_mutex_lock(&async_lock);
-    char *old_data = asyncdata_;
-    asyncdata_ = new_data;
-    asyncsize_ = size;
+    queue.push_back(std::string(data, size));
     uv_mutex_unlock(&async_lock);
-
-    if (old_data) {
-      delete[] old_data;
-    }
-    uv_async_send(async);
+    uv_async_send(&async);
   }
 
   NAN_INLINE static NAUV_WORK_CB(AsyncProgress_) {
@@ -1678,14 +1669,12 @@ class Callback {
   NAN_INLINE static void AsyncClose_(uv_handle_t* handle) {
     AsyncProgressWorker *worker =
             static_cast<AsyncProgressWorker*>(handle->data);
-    delete reinterpret_cast<uv_async_t*>(handle);
     delete worker;
   }
 
-  uv_async_t *async;
+  uv_async_t async;
+  async_queue_t queue;
   uv_mutex_t async_lock;
-  char *asyncdata_;
-  size_t asyncsize_;
 };
 
 NAN_INLINE void AsyncExecute (uv_work_t* req) {
--- a/nan.h
+++ b/nan.h
@@ -1610,6 +1610,11 @@ class Callback {
     uv_mutex_destroy(&async_lock);
   }
 
+  virtual void WorkComplete() {
+    WorkProgress();
+    AsyncWorker::WorkComplete();
+  }
+
   void WorkProgress() {
     // Dont send progress events after we've already completed.
     if (callback) {
--- a/nan.h
+++ b/nan.h
@@ -1592,7 +1592,6 @@ class Callback {
 
 /* abstract */ class AsyncProgressWorker : public AsyncWorker {
  public:
-
   typedef std::deque<std::string> async_queue_t;
 
   explicit AsyncProgressWorker(Callback *callback_)
--- a/test/cpp/asyncprogressworker.cpp
+++ b/test/cpp/asyncprogressworker.cpp
@@ -11,6 +11,8 @@
 #define Sleep(x) usleep((x)*1000)
 #endif
 #include <nan.h>
+#include <string>
+#include <sstream>
 
 using namespace Nan;  // NOLINT(build/namespaces)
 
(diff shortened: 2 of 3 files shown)


nan.h
```diff
@@ -1591,39 +1592,43 @@ class Callback {
...
-    uv_mutex_lock(&async_lock);
-    char *data = asyncdata_;
-    size_t size = asyncsize_;
-    asyncdata_ = NULL;
-    uv_mutex_unlock(&async_lock);
-
     // Dont send progress events after we've already completed.
     if (callback) {
-        HandleProgressCallback(data, size);
+        while(!queue.empty()) {
```
bnoordhuis: You need to hold the lock before calling `queue.empty()`.

nan.h
```diff
@@ -1678,14 +1679,15 @@ class Callback {
   NAN_INLINE static void AsyncClose_(uv_handle_t* handle) {
     AsyncProgressWorker *worker =
             static_cast<AsyncProgressWorker*>(handle->data);
-    delete reinterpret_cast<uv_async_t*>(handle);
-    delete worker;
+    while(!worker->queue.empty()) {
```
bnoordhuis: Style: space before '('

nan.h
```diff
@@ -1591,39 +1592,43 @@ class Callback {
...
     if (callback) {
-        HandleProgressCallback(data, size);
+        while(!queue.empty()) {
+            uv_mutex_lock(&async_lock);
+            asyncdata_t *asyncdata = queue.front();
+            queue.pop_front();
+            uv_mutex_unlock(&async_lock);
+            HandleProgressCallback(asyncdata->data, asyncdata->size);
+            delete asyncdata;
+        }
```
bnoordhuis: You might as well fix the indent errors in this method and Destroy() while you're here.

nan.h
```diff
@@ -1678,14 +1679,15 @@ class Callback {
   NAN_INLINE static void AsyncClose_(uv_handle_t* handle) {
     AsyncProgressWorker *worker =
             static_cast<AsyncProgressWorker*>(handle->data);
-    delete reinterpret_cast<uv_async_t*>(handle);
-    delete worker;
+    while(!worker->queue.empty()) {
+      delete worker->queue.front();
+      worker->queue.pop_front();
+    }
   }
```
bnoordhuis: I think this introduces a memory leak.  What deletes the AsyncProgressWorker now?

nan.h
```diff
@@ -1591,39 +1592,43 @@ class Callback {
...
+      , &async
       , AsyncProgress_
     );
-    async->data = this;
+    async.data = this;
 
     uv_mutex_init(&async_lock);
   }
 
   virtual ~AsyncProgressWorker() {
```
bnoordhuis: /cc @kkoopa - Hadn't realized this before but this destructor should probably be private, with Destroy() being the only way to delete an instance.  This:
```c++
{
  AsyncProgressWorker temp(callback);
}
```
Is unsound because the `uv_async_t` handle is not closed.

nan.h
```diff
@@ -1591,39 +1592,43 @@ class Callback {
 
 /* abstract */ class AsyncProgressWorker : public AsyncWorker {
  public:
+
+  typedef struct asyncdata_t {
+    char* data;
+    size_t size;
+    AsyncProgressWorker* worker;
+  } asyncdata_t;
```
agnat: Please use a C++ struct instead of a C-style struct typedef.

nan.h
```diff
@@ -20,6 +20,7 @@
 #ifndef NAN_H_
 #define NAN_H_
 
+#include <deque>
```
agnat: This include statement should go with the others around line 50.

nan.h
kkoopa: Yes, probably. Although is not meant to be stack allocated either. The only legitimate use is to feed it heap allocated to AsyncQueWorker for subsequent deletion.

bnoordhuis: In that case I'd enforce that by making the destructor private.  Doesn't have to be done in this PR though.

kkoopa: Yes, all of AsyncWorker / AsyncProgressWorker / AsyncQueueWorker should be redesigned at some future point.

bfrancojr:
Hey @bnoordhuis, @kkoopa and @agnat,

Could you please review my latest changes?
I also fixed another memory leak regarding asyncdata->data pointer.

Thanks!

nan.h
```diff
@@ -1591,39 +1592,51 @@ class Callback {
...
 /* abstract */ class AsyncProgressWorker : public AsyncWorker {
  public:
+
+  struct AsyncData {
+    AsyncData(const char* data_, size_t size_)
+        : data(new char[size_]), size(size_) {
+      memcpy(data, data_, size_);
+    }
+    ~AsyncData() {
+      delete data;
```
bnoordhuis: `delete[]`

nan.h
```diff
@@ -1591,39 +1592,51 @@ class Callback {
...
+    }
+    char* data;
+    size_t size;
```
bnoordhuis: If you assign in the constructor (good!) consider making these `const` (i.e. `char* const` and `const size_t`.)

nan.h
agnat: Great catch. Missed that.

nan.h
```diff
@@ -1591,39 +1592,51 @@ class Callback {
...
+  };
```
bnoordhuis: I suppose a case could be made that this is just reinventing `std::string(s, n)`.

nan.h
```diff
@@ -1591,39 +1592,51 @@ class Callback {
...
     if (callback) {
-        HandleProgressCallback(data, size);
+      uv_mutex_lock(&async_lock);
+      while (!queue.empty()) {
+        AsyncData *async_data = queue.front();
+        queue.pop_front();
+        uv_mutex_unlock(&async_lock);
+        HandleProgressCallback(async_data->data, async_data->size);
+        uv_mutex_lock(&async_lock);
+        delete async_data;
```
bnoordhuis: I've move the delete before the call to `uv_mutex_lock()`, it doesn't need to be inside the critical section.

nan.h
agnat: Lol. True, now with the (unused) worker member gone...

nan.h
```diff
@@ -1591,39 +1593,38 @@ class Callback {
...
-    size_t size = asyncsize_;
-    asyncdata_ = NULL;
-    uv_mutex_unlock(&async_lock);
-
     // Dont send progress events after we've already completed.
     if (callback) {
-        HandleProgressCallback(data, size);
+      uv_mutex_lock(&async_lock);
+      while (!queue.empty()) {
+        std::string async_data = queue.front();
```
agnat: I'm not quite sure, but I think we don't have to copy the string here. I think it is Ok to hold a `const&` to the string and delay the `pop_front()` until we re-aquire the lock late int the loop.

BTW, I really like this:

screen shot 2015-11-12 at 09 24 55

Thanks, for following through, @bfrancojr. Much appreciated.

nan.h
```diff
@@ -1678,14 +1670,15 @@ class Callback {
   NAN_INLINE static void AsyncClose_(uv_handle_t* handle) {
     AsyncProgressWorker *worker =
             static_cast<AsyncProgressWorker*>(handle->data);
-    delete reinterpret_cast<uv_async_t*>(handle);
+    while (!worker->queue.empty()) {
+      worker->queue.pop_front();
+    }
```
agnat: This is no longer necessary.

nan.h
```diff
@@ -49,6 +49,8 @@
 #include <cstring>
 #include <climits>
 #include <cstdlib>
+#include <deque>
+#include <string>
```
agnat: Superfluous include. `string` is included just a few lines below. ;)

bnoordhuis:
LGTM.  Can you squash the commits?

bfrancojr:
Although this PR fixes one issue I was facing, the callback function is not
being called at least once as libuv's documentation states.
I had to add a usleep(500) after the progress.Send call as workaround.
Looks like uv_async_send is not reliable at all as it is implemented now.

On Thu, Nov 12, 2015 at 1:47 AM, Ben Noordhuis <notifications@github.com>
wrote:

> LGTM. Can you squash the commits?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nodejs/nan/pull/515#issuecomment-156051628>.
>



-- 
Bene.

bnoordhuis:
I think it's unlikely that the issue is with libuv because we'd have received no end of bug reports if that were the case.  Quick sanity check, what happens when you move the call to uv_async_send() inside the critical section?

nan.h
```diff
@@ -1591,39 +1592,38 @@ class Callback {
...
-    uv_mutex_lock(&async_lock);
-    char *data = asyncdata_;
-    size_t size = asyncsize_;
-    asyncdata_ = NULL;
-    uv_mutex_unlock(&async_lock);
-
     // Dont send progress events after we've already completed.
     if (callback) {
-        HandleProgressCallback(data, size);
+      uv_mutex_lock(&async_lock);
```
reqshark: locking is slow, is this justified?

bfrancojr:
It is a little better.
Now, if there is one single call to progress.Send, no callback is called but if the thread makes 2 or more calls to progress.Send, it works as expected.

agnat:
At which point did your original solution stop working for you?

> Any idea/help will be greatly appreciated.

Maybe write a test we can look at?

bfrancojr:
I've added more test case that simulates simular behavior I'm getting.
It fails under OSX 10.11 but works under Ubuntu 14.04.

agnat:
Hm... I see... I think this behavior is by design and it seems legit to me. When the work is done the whole thing is destroyed and pending queue items are lost. You know it is done because your callback got called without an error argument. The progress events are merely hints.

test/cpp/asyncprogressworker.cpp
```diff
@@ -47,6 +47,36 @@ class ProgressWorker : public AsyncProgressWorker {
...
+        Callback *callback
+        , Callback *progress
+        , int iters)
+        : AsyncProgressWorker(callback), progress(progress)
+        , iters(iters) {}
+    ~ProgressWorkerNoWait() {}
+
+    void Execute (const AsyncProgressWorker::ExecutionProgress& progress) {
+        for (int i = 0; i < iters; ++i) {
+            progress.Send(reinterpret_cast<const char*>(&i), sizeof(int));
```
agnat: I'm not ok with this. Use proper string conversions.

agnat:
> [...] which can be overwritten in subsequents calls to progress.Send method even before the callback to be called.

You definitely fixed that, but I'm starting to believe that that was never the problem. What good is outdated progress information? Maybe you don't need an `AsyncProgressWorker` but an `AsyncObjectFactory`? Could you take a second to describe what you are actually trying to do? I just want to avoid landing the wrong fix.

bfrancojr:
I've fixed a race condition problem and it was a real bug, IMHO.
After having the proper messaging passing implementation in place, I've noticed unpredictable results  which, in my opinion, is not desirable. If that is the case, features like that should be deprecated or at least,  the documentation should warning developers about it. Again, I'm just looking for a consistent behavior across platforms.
Regarding callback error argument, I couldn't find anything in the code linking uv_async_send, callbacks and error handling. Could you please enlighten me about this?

agnat:
Hehe, ok. I might be misreading the code ... but I think not. @bnoordhuis, @kkoopa help me out here. Could you guys take a look and tell me where I'm wrong?

1. `AsyncWorker` is very simple. It calls a callback when the work is done. This callback is the primary and authoritative source of progress information: We are done. (If the work fails it calls the callback with an error argument. I only mentioned that for completeness.)

2. `AsyncProgressWorker` adds _current_ progress information on top of that. And now follow me here. It is totally Ok to overwrite the progress data, because that _is_ the _current_ progress. If the JavaScript part of the application is to busy to get all the values all the better. That's why we do concurrency. I claim the only thing wrong with it is the method names starting with `Send(...)`. The word "send" is closely associated with message passing which is not what is does.

3. The behavior you seek, delivering each and every "progress message", is easily implemented. All you need is to flush the queue (deliver each item that is) at an appropriate point before the callback is called. But...

4. There is no added value in this. All the values in the queue except the last one are historic, superseded and outdated. In case the work is already done I don't really need to know that it was 80% some time ago.

5. Although this specific case is easily mitigated (see above) you'll never get complete consistency when multithreading is involved. You will always have observable differences even across different runs on the _same_ platform.

6. There is really nothing wrong with `uv_async_send()`. Believe.

I totally see the utility of an `AsyncWorker` that supports message passing. So, here is a proposal. In this PR we only land the clean-ups you did on the fly. Not allocating members on the heap, using `std::string` instead of `char*, size_t`, &c. But we keep it lazy and lossy. All the queue goodness could go into another PR  implementing `template <typename T> AsyncObjectFactory` (or what have you) which passes objects of type `T` using a `std::deque`, with final flushing and all.

However, if @bnoordhuis or @kkoopa are pro-progress-queue I won't raise hell. ;)

bnoordhuis:
I only reviewed the PR on its technical merit, I don't have an opinion on whether it's needed or not.  I assume @kkoopa thinks so because he asked me to review it but I'll let him speak for himself. :-)

kkoopa:
I know very little of libuv and its intricacies, so I tend to defer to what I see as better judgement in this case. If there is some class with high enough utility, not eadily realizable with the current AsyncWorkers, I am in favor of adding them if there are complicated steps involved.  But if something can be trivially implemented by simply extending one of the existing workers, and / or it has low usage likelihood, it is better to keep it unimplemented.

agnat:
lol. _I_ only reviewed the PR on its technical merit... initially. I did so because it involves multithreading and another pair of eyes couldn't hurt, no?

@kkoopa Ok, that's regarding my proposal to add another worker. I guess the key question is this: Do we want full blown progress message passing with delivery guaranties or are we Ok with the current lazy approach of only keeping the latest-and-greatest with no guaranties except we'll call you when we're done?

kkoopa:
I guess your analysis of what the current implementatuon does is quite alright. I honestly don't know if there is any other legitimate use case which is common enough to warrant inclusion in NAN or not.

mathiask88:
@kkoopa @agnat I want to implement a worker that does an epoll on a gpio pin, reads a byte from i2c on every interrupt and send that byte to the user callback. That's the kind of use case you are talking about, isn't it? But is it a common use case? I don't think so.

kkoopa:
Does not sound like too common a use case outside of robotics / raspberry pi / etc. -- where a library for that segment could provide said functionality.

mathiask88:
@kkoopa More or less. There is an epoll and an i2c module, but then the flow would be `native epoll -> JS land i2c.read -> native i2c.read -> js land` which would be too slow in my case where I want to face ~3000 interrupts & reads per second. So something like `native epoll -> native i2c read -> callback to js land` would be preferable. But I agree with you that this is no use case that would justify a new worker class. I just wanted to say "Hey, I got a use case for that" :)

PR #518, opened 2015-11-11
Fix Global::Pass

kkoopa:


test/cpp/persistent.cpp
```diff
@@ -49,30 +49,46 @@ NAN_METHOD(CopyablePersistent) {
   info.GetReturnValue().Set(New(p));
 }
 
+template<typename T> Global<T> passer(v8::Local<T> handle) {
+  return Global<T>(handle).Pass();
+}
+
+NAN_METHOD(PassGlobal) {
+  info.GetReturnValue().Set(passer(New(12)));
```
bnoordhuis: Clearly that should be 42.

bnoordhuis:
LGTM

kkoopa:
Just noticed that the commit message is wrong. But someting is weird with this test, if I move the added line somewhere else, some of the other tests fail on 0.10 and it is not evident why.

kkoopa:
There, fixed the persistent tests.

bnoordhuis:
Even more LGTM.

PR #519, opened 2015-11-13
fix doc wrt to if Nan::Persistent is copyable

chad3814:
the text is exactly wrong, the code snippet is correct.

### DIFF
--- a/doc/persistent.md
+++ b/doc/persistent.md
@@ -155,7 +155,7 @@ See the V8 documentation for [`CopyablePersistentTraits`](https://v8docs.nodesou
 <a name="api_nan_persistent"></a>
 ### Nan::Persistent
 
-A type of `PersistentBase` which allows copy and assignment. Copy, assignment and destructor behavior is controlled by the traits class `M`.
+A type of `PersistentBase` which does not allows copy nor assignment. Copy, assignment and destructor behavior is controlled by the traits class `M`.
 
 Definition:
 


kkoopa:
No, that only applies to Nan::Persistent<T> which is merely a special case of Nan::Persistent<T, M>. The former implicitly uses Nan::NonCopyablePersistentTraits.

chad3814:
This doc is confusing:
https://github.com/nodejs/nan/blob/master/doc/persistent.md#nanpersistent

>### Nan::Persistent
>
>A type of `PersistentBase` which allows copy and assignment. Copy, assignment and destructor behavior is controlled by the traits class `M`.

the sentence "A type of `PersistentBase` which allows copy and assignment." is misleading when the header is just `Nan::Persistent` which defaults to non-copyable.

kkoopa:
But it clearly says the copy, assignment and constructor behavior is controlled by the traits class. The Persistent classes allow for copying and assignment, but they are associated with traits classes and it is those traits classes that either allow or disallow something.

chad3814:
I'm honestly not trying to make trouble, but it is very confusing to me. The default, if you don't specify a trait class is non-copyable. Maybe if it said "A type of `PersistentBase` which can allow copy and assignment with the `Nan::CopyablePersistentTraits` class.

kkoopa:
Under NonCopyablePersistentTraits it says that it is the default traits for Persistent. The definition also shows that the default traits is NonCopyablePersistentTraits. I guess a sentence could be added for Persistent stating that NonCopyablePersistentTraits is the default, with a link there.

PR #529, opened 2016-01-05
Rename GC*logueCallback to GCCallback for >4.0

matthewloring:
Types were unified in https://codereview.chromium.org/1298113003
This change went into v8 4.6 which made it into Node 5. The
GC*logueCallback types have gone away completely in 4.9.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -598,25 +598,33 @@ class TryCatch {
 # define NAN_GC_CALLBACK(name)                                                 \
     void name(v8::Isolate *isolate, v8::GCType type, v8::GCCallbackFlags flags)
 
+#if NODE_MODULE_VERSION <= NODE_4_0_MODULE_VERSION
+  typedef v8::Isolate::GCEpilogueCallback GCEpilogueCallback;
+  typedef v8::Isolate::GCPrologueCallback GCPrologueCallback;
+#else
+  typedef v8::Isolate::GCCallback GCEpilogueCallback;
+  typedef v8::Isolate::GCCallback GCPrologueCallback;
+#endif
+
   NAN_INLINE void AddGCEpilogueCallback(
-      v8::Isolate::GCEpilogueCallback callback
+      GCEpilogueCallback callback
     , v8::GCType gc_type_filter = v8::kGCTypeAll) {
     v8::Isolate::GetCurrent()->AddGCEpilogueCallback(callback, gc_type_filter);
   }
 
   NAN_INLINE void RemoveGCEpilogueCallback(
-      v8::Isolate::GCEpilogueCallback callback) {
+      GCEpilogueCallback callback) {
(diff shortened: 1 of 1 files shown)


nan.h
```diff
@@ -598,25 +598,33 @@ class TryCatch {
 # define NAN_GC_CALLBACK(name)                                                 \
     void name(v8::Isolate *isolate, v8::GCType type, v8::GCCallbackFlags flags)
 
+#if NODE_MODULE_VERSION <= NODE_4_0_MODULE_VERSION
+  typedef v8::Isolate::GCEpilogueCallback GC_EPILOGUE_CALLBACK_TYPE;
+  typedef v8::Isolate::GCPrologueCallback GC_PROLOGUE_CALLBACK_TYPE;
```
agnat: Please use `CamelCase` for (new) typedefs. (`UPPER_CASE` types had an earlier life as macros and have not been renamed... yet). Oh, and maybe drop the `_TYPE` suffix. Just use the original names.

agnat:
Thanks, @matthewloring.

LGTM with one nit.

nan.h
matthewloring: Fixed.

kkoopa:
I was going with [a different approach](https://github.com/nodejs/nan/pull/489), but I guess this could work as a temporary fix as it does not change the external API.

agnat:
> I was going with a different approach,

Right. I almost forgot about that one...

> but I guess this could work as a temporary fix as it does not change the external API.

I'm OK with either one. This one offers a smoother transition while #489 _might_ be a bit more future proof since it offers more ... wiggle room. "Might" because predictions are hard, specially if they concern the future.

BTW, does anybody know an open source add-on that uses the GC*logue stuff? I'm kind of curious what people do with it.

matthewloring:
I'm not familiar with how people use this. I only encountered it when things crashed using nan with node 6.0-pre. I'm also happy with either solution, I was not aware of the other approach before starting this PR.

kkoopa:
I can merge this and put out a minor update if things don't work or give deprecation warnings in 5.0. Could even merge it regardless, if it happens to help for prereleases of 6. I have been meaning to get out the next minor for a couple of months already, but not gotten around to it, since nothing has been broken yet, AFAIK. The other PR is scheduled for NAN 3.0, which I intend to hold off until Node 6 in April or so.

@agnat Yes, there is one, for which I originally wrote this functionality. I think it was [this](https://github.com/bretcope/node-gc-profiler). I don't know of anything else that uses it.

matthewloring:
I've been experimenting with a branch of Node on top of v8 4.9 and this change is required to make nan work in this context. I believe it is also giving deprecation warnings in 5.0 based on the original v8 commit but I haven't confirmed this personally. It would definitely help me to get this fix in before nan 3.0.

PR #536, opened 2016-01-28
Add required compiler versions in the error msg

apaatsio:
Save the users a Google search and tell them what compiler version they need.

See: https://github.com/nodejs/nan/issues/448

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -38,7 +38,7 @@
 #endif
 
 #if NODE_MODULE_VERSION >= IOJS_3_0_MODULE_VERSION && !NAN_HAS_CPLUSPLUS_11
-# error This version of node/NAN/v8 requires a C++11 compiler
+# error This version of node/NAN/v8 requires a C++11 compiler (at least g++ 4.8 or clang++ 3.4)
 #endif
 
 #include <uv.h>


rvagg:
This would need an MSVC version too for completeness

kkoopa:
I do not like it. Listing versions explicitly is too much detail. There can 
still be discrepancies, compiler bugs, vendors not applying the right patches, 
wrong flags and so on. Not to mention that no version of msvc truly has proper 
C++11 support yet.

The code which produces the error message mostly relies on feature detection, 
and it is merely there as convenience to limit the number of frivolous issues 
due to not grokking the compiler error which would be produced otherwise. What 
will happen when the next compiler comes along and is not listed?

apaatsio:
Yeah, it's a trade-off between "more useful error message for most users" and "not so useful but accurate error message"

ranisalt:
Clang 3.3 has full C++11 support already, but Nan probably can be compiled with Clang 3.1 since patches 3.2 and 3.3 are not very used ([source](http://clang.llvm.org/cxx_status.html))

agnat:
I'm with @kkoopa on this. We (try to) point the user in a clear direction. Maintaining a list of working compilers is beyond our scope.

PR #537, opened 2016-01-29
fix formatting of the example js file

thefourtheye:


rvagg:
whoops I saw #538 first, would you mind rebasing, I think you have some additional changes to go in, sorry.

thefourtheye:
@rvagg I had the inconsistent spacing and the semi-clone fix only. No problem anyway :-) Rebased. Please check if it is worth including now :-)

PR #538, opened 2016-01-30
Use only spaces for indentation in addon example

addaleax:
Right now, both tabs and spaces are mixed (in some weird way) in this file, making it look strangely in some editors.


rvagg:
tabs?? evil! thanks.

PR #542, opened 2016-02-27
Remove null pointer checks

ranisalt:
There's no need to check if a pointer is equal to NULL prior to deleting it, as deleting a null pointer is a no-op ([source](http://en.cppreference.com/w/cpp/language/delete), search for "expression evaluates to a null pointer value").

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1501,10 +1501,8 @@ class Callback {
 
     if (!persistentHandle.IsEmpty())
       persistentHandle.Reset();
-    if (callback)
-      delete callback;
-    if (errmsg_)
-      delete[] errmsg_;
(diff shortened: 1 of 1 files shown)


kkoopa:
This is probably good to merge, if the specification can be trusted to be correctly implemented. Wonder why there is no Travis report visible.

ranisalt:
Maybe the hook didn't fire correctly but the build is [here](https://travis-ci.org/nodejs/nan/builds/112325580).

agnat:
> There's no need to check if a pointer is equal to NULL prior to deleting it

Indeed. LGTM. Thanks!

PR #545, opened 2016-03-14
Dataless progress callbacks for AsyncProgressWorker

JoshuaWise:
I'd like the ability to efficiently send progress without data. Since `uv_async_send` coalesces callbacks, it's probably must more common for people to implement data callbacks with a queue as a member variable anyways.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1632,6 +1632,9 @@ class Callback {
     friend class AsyncProgressWorker;
    public:
     // You could do fancy generics with templates here.
+    void Signal() const {
+        uv_async_send(that_->async);
+    }
     void Send(const char* data, size_t size) const {
         that_->SendProgress_(data, size);
     }
--- a/nan.h
+++ b/nan.h
@@ -1631,10 +1631,10 @@ class Callback {
   class ExecutionProgress {
     friend class AsyncProgressWorker;
    public:
-    // You could do fancy generics with templates here.
     void Signal() const {
         uv_async_send(that_->async);
     }
+    // You could do fancy generics with templates here.
     void Send(const char* data, size_t size) const {
         that_->SendProgress_(data, size);
     }
--- a/test/cpp/asyncprogressworkersignal.cpp
+++ b/test/cpp/asyncprogressworkersignal.cpp
@@ -35,10 +35,8 @@ class ProgressWorker : public AsyncProgressWorker {
   void HandleProgressCallback(const char *data, size_t size) {
     HandleScope scope;
 
-    v8::Local<v8::Value> argv[] = {
-        New<v8::Boolean>(data == NULL && size == 0)
-    };
-    progress->Call(1, argv);
+    v8::Local<v8::Value> arg = New<v8::Boolean>(data == NULL && size == 0);
+    progress->Call(1, &arg);
   }
 
  private:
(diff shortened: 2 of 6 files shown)


bnoordhuis:
Can you add tests and a decent commit log?  The first line should be a brief summary of the change.

Bikeshed: `Send()` is arguably not the best name for what it does.  `Signal()` might be more accurate.

JoshuaWise:
Okay, I changed the method name to `Signal()`, improved the commit log, and added a test.

nan.h
```diff
@@ -1632,6 +1632,9 @@ class Callback {
     friend class AsyncProgressWorker;
    public:
     // You could do fancy generics with templates here.
```
bnoordhuis: Teeny tiny nit: the comment belongs to the function below, although I'd be fine with removing it, it doesn't add much value.

test/cpp/asyncprogressworkersignal.cpp
```diff
@@ -0,0 +1,66 @@
...
+ * NAN - Native Abstractions for Node.js
+ *
+ * Copyright (c) 2016 NAN contributors
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
+
+#ifndef _WIN32
+#include <unistd.h>
+#define Sleep(x) usleep((x)*1000)
```
bnoordhuis: Can we perhaps break this out into a header file?  I see it's already duplicated three times, this one is the fourth.

test/cpp/asyncprogressworkersignal.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  void HandleProgressCallback(const char *data, size_t size) {
+    HandleScope scope;
+
+    v8::Local<v8::Value> argv[] = {
+        New<v8::Boolean>(data == NULL && size == 0)
+    };
+    progress->Call(1, argv);
```
bnoordhuis: This code is fine as is but there's a shorthand version:
```c++
v8::Local<v8::Value> arg = New<v8::Boolean>(data == NULL && size == 0);
progress->Call(1, &arg);
```
Feel free to ignore. :-)

bnoordhuis:
Mostly LGTM.  Maybe one of the other collaborators can take a peek as well?

kkoopa:
I guess it's good. One thing that is missing is addition of the test/cpp/asyncprogressworkersignal.cpp file to the Makefile to get the linter to run on it.

JoshuaWise:
@kkoopa 
I added `test/cpp/asyncprogressworkersignal.cpp ` to the Makefile.

kkoopa:
Good, this will be merged. However, I don't know whether there will be more feature releases in the 2-series, as Node 6 is coming soon and will require NAN 3. Might well do a final point release of NAN 2 if it seems like the transition to 3 will be demanding, but it is still too early to tell. Do you require this urgently?

JoshuaWise:
Fairly urgently, yes. I don't imagine I'll be able to switch to NAN 3 for current projects unless the transition is trivial, so a final point release of NAN 2 would be much appreciated.

kkoopa:
Alright. A final point release it is then. It might still take a while to tie up unrelated loose ends and see what else to possibly include there, but I'll add this to my list.

PR #553, opened 2016-03-29
Use NewFromUnsigned in ReturnValue<T>::Set(uint32_t i) for pre_12

### LINKED ISSUE #552: ReturnValue<T>::Set(uint32_t i) does not preserve unsignedness on 0.10
One would expect this code:

     info.GetReturnValue().Set((uint32_t)0xffffffff);

to produce the same result as

    info.GetReturnValue().Set(Nan::New((uint32_t)0xffffffff));

on nodejs 0.10, but it does not. In the first case the returned value is -1, in the second case it's 4294967295.
Nodejs 5.0 returns the positive value in both cases.

In `nan_implementation_pre_12_inl.h` we see (the code in `nan_implementation_12_inl.h` is essentially the same):
```
Factory<v8::Uint32>::return_t
Factory<v8::Uint32>::New(uint32_t value) {
  return To<v8::Uint32>(v8::Uint32::NewFromUnsigned(value));
}
```

But the Implementation of `ReturnType<T>` in `nan_callbacks_pre_12_inl.h`:
```
  inline void Set(uint32_t i) {
    TYPE_CHECK(T, v8::Integer);
    value_->Dispose();
    *value_ = v8::Persistent<T>::New(v8::Uint32::New(i));
  }
```
does not use `v8::Uint32::NewFromUnsigned`, but straight `v8::Uint32::New`.
Changing `New` to `NewFromUnsigned` **fixes** the behavior. 

`nan_callbacks_12_inl.h` just forwards to v8:
https://github.com/v8/v8/blob/master/include/v8.h#L7708
which uses `Integer::NewFromUnsigned` (in this latest version).

There has been one noteworthy change to v8's `ReturnValue::Set(uint32_t)` since it's introduction: https://github.com/v8/v8/commit/3c8a79cabadeaa8c356b7789a8b621464d643b2d - but I'm not sure of its implications.

smilingthax:
Fixes #552

### DIFF
--- a/nan_callbacks_pre_12_inl.h
+++ b/nan_callbacks_pre_12_inl.h
@@ -65,7 +65,7 @@ class ReturnValue {
   inline void Set(uint32_t i) {
     TYPE_CHECK(T, v8::Integer);
     value_->Dispose();
-    *value_ = v8::Persistent<T>::New(v8::Uint32::New(i));
+    *value_ = v8::Persistent<T>::New(v8::Uint32::NewFromUnsigned(i));
   }
 
   // Fast JS primitive setters
--- a/test/cpp/returnvalue.cpp
+++ b/test/cpp/returnvalue.cpp
@@ -30,6 +30,10 @@ NAN_METHOD(ReturnGlobal) {
   info.GetReturnValue().Set(global);
 }
 
+NAN_METHOD(ReturnUnsigned) {
+  info.GetReturnValue().Set(0x80000000u);
+}
+
 NAN_MODULE_INIT(Init) {
   global.Reset(New(true));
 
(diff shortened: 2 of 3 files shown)


kkoopa:
Thank you. Don't know why the CI hooks don't show any output in GitHub anymore, but the builds seem to pass. https://travis-ci.org/nodejs/nan/builds/119252584 https://ci.appveyor.com/project/RodVagg/nan/build/1135
The AppVeyor failure is timing related and false.

I'll merge and publish a new version soon.

bnoordhuis:
One more LGTM.

kkoopa:
Thanks, Ben. I'll push out a patch release.

PR #559, opened 2016-04-26
Don't use deprecated v8::Template::Set()

### LINKED ISSUE #558: deprecation warnings when used with node v6.0.0
The default use of `Nan::SetPrototypeMethod` is giving warnings under node 6.0.  I had to change from:

```
Nan::SetPrototypeMethod(tpl, "flood", Flood);
```

to:
```
tpl->PrototypeTemplate()->Set(Nan::New<v8::String>("flood").ToLocalChecked(), Nan::New<v8::FunctionTemplate>(QuadTree::Flood));
```
in order to bypass the warning. It would be nice if SetPrototypeMethod was updated to do this internally.

mdouglass:
Setting non-primitive values on FunctionTemplate and ObjectTemplate
instances is depreciated.
Fixes #558

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1867,13 +1867,12 @@ NAN_INLINE void SetPrototypeMethod(
     v8::Local<v8::FunctionTemplate> recv
   , const char* name, FunctionCallback callback) {
   HandleScope scope;
-  v8::Local<v8::Function> fn = GetFunction(New<v8::FunctionTemplate>(
+  v8::Local<v8::FunctionTemplate> fn = New<v8::FunctionTemplate>(
       callback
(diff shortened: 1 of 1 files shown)


mdouglass:
Likely an incomplete fix and I'm not sure if there are other backwards incompatibility issues that need to be addressed to complete this. It does, however, silence the warning for the simple case I've been testing with.

bnoordhuis:
Sorry, hadn't seen this when I opened #560.  This PR mostly LGTM but, unlike #560, it loses the class name.

mdouglass:
No worries, looks like you're closing in on a better fix, I'll close this.

PR #560, opened 2016-04-26
Don't use deprecated v8::Template::Set().

bnoordhuis:
See [0] and [1]: starting with node.js v6, setting non-primitive values
on FunctionTemplate and ObjectTemplate instances is discouraged; v7 will
downright disallow it.  Update `Nan::SetPrototypeMethod()`.

[0] https://github.com/nodejs/node/issues/6216
[1] https://github.com/nodejs/node/pull/6228

PR #561, opened 2016-04-27
Don't use deprecated v8::Template::Set() in SetMethod

mcollina:
Same as https://github.com/nodejs/nan/pull/560, but for SetMethod.

I think it's ok for releasing this as 2.3.1.

cc @kkoopa @bnoordhuis

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1856,10 +1856,10 @@ NAN_INLINE void SetMethod(
   , const char *name
   , FunctionCallback callback) {
   HandleScope scope;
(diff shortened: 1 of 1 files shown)


kkoopa:
Thanks for reporting, but `Nan::SetMethod` is supposed to do what `node::NODE_SET_METHOD` does. This would not work on a `v8::Object`, so I instead went with adding an overload in a90951e9ea70fa1b3836af4b925322919159100e.

Issue #564, opened 2016-04-29
Consider supporting "void Nan::SetMethod(v8::Local<v8::Function>"


springmeyer:
The change in https://github.com/nodejs/nan/commit/a90951e9ea70fa1b3836af4b925322919159100e
 broke `void Nan::SetMethod(v8::Local<v8::Function>, ...` usage I'd unknowingly been depending on.

Per https://github.com/nodejs/nan/commit/a90951e9ea70fa1b3836af4b925322919159100e#commitcomment-17309590 creating this ticket to discuss if this should be fixed in Nan or if users should workaround by casting to an Object.

kkoopa:
For reference, there is no true ambiguity, a `v8::Function` is a `v8::Object`, so a `v8::Local<v8::Function>` is a `v8::Local<v8::Object>`. Node did that same change in nodejs/node@642076f2aff3cddf93dd7cb6b1a332f8064ab39f and has the same problem. I tried working around it for the `v8::Template` case in bbf82b0b501ef3ed596d38ea9a87f81cbcf4c5e1, which, while ugly, is manageable, since there are only two subclasses of `v8::Template`. However, there are too many subclasses of `v8::Object`, so there has to be a better way. Something with `enable_if` and `is_convertible` might work, but is rather tedious.

cc @bnoordhuis @agnat

agnat:
I agree that unrolling is not really an option. Like you suggest, meta-programing is probably the better approach. I can look into it, but I'll need a couple of days...

> For reference, there is no true ambiguity [...]

Although I like to blame the compiler just like the next guy the compiler is in the right. It is ambiguous. `Function` is an `Object`, it's true but: `Local<Function>` and `Local<Object>` are completely unrelated types. Both types are generated by the same template, but that does not imply any relation. Recall that `template Local` has this nasty templateized constructor... the one that triggers a `STATIC_ASSERT` if the two types are incompatible. This constructor makes it ambiguous since it offers a conversion from some `Local<Foo>` to any other `Local`. Anyway...

PS: `using namespace v8;`

kkoopa:
Yeah, I was using some amount of hyperbole. I did test this issue on clang as well, with the same result. However, if the compiler is not wrong, then the standard is wrong. There should be a convenient way of stating that a function accepts any `A<? extends B>` or similar, with normal, automatic conversion of `A` and `B`, especially in regard to `B`.

[...]

PR #566, opened 2016-05-01
Refactor SetMethod(...)

agnat:
As discussed in #564 here is a first stab at refactoring `SetMethod(...)`. @kkoopa, @bnoordhuis PTAL and let me know what you think.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1852,75 +1852,47 @@ NAN_INLINE void SetInstanceTemplate(
   SetTemplate(templ->InstanceTemplate(), name, value, attributes);
 }
 
-#if NODE_MODULE_VERSION < IOJS_3_0_MODULE_VERSION
-NAN_INLINE void SetMethod(
-    v8::Handle<v8::Object> recv
-  , const char *name
-  , FunctionCallback callback) {
-  HandleScope scope;
-  v8::Local<v8::Function> fn = GetFunction(New<v8::FunctionTemplate>(
-      callback)).ToLocalChecked();
-  v8::Local<v8::String> fn_name = New(name).ToLocalChecked();
-  fn->SetName(fn_name);
-  recv->Set(fn_name, fn);
+namespace imp {
+
+// Note(@agnat): Helper to distinguish different receiver types. The first
+// version deals with receivers derived from v8::Template. The second version
+// handles everything else. The final argument only serves as discriminator and
+// is unused.
+template <typename T>
+NAN_INLINE
+void
+SetMethodAux(T recv,
+             v8::Local<v8::String> name,
+             v8::Local<v8::FunctionTemplate> tpl,
+             v8::Template *) {
+  recv->Set(name, tpl);
 }
 
+template <typename T>
+NAN_INLINE
+void
+SetMethodAux(T recv,
+             v8::Local<v8::String> name,
+             v8::Local<v8::FunctionTemplate> tpl,
+             ...) {
+  recv->Set(name, GetFunction(tpl).ToLocalChecked());
+}
+
+}  // end of namespace imp
+
+template <typename T, template <typename> class HandleType>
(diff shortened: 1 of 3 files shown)


kkoopa:
Clever use of the dummy argument to distinguish types while preserving normal type matching. This seems like an elegant solution.

nan.h
```diff
@@ -1852,75 +1852,47 @@ NAN_INLINE void SetInstanceTemplate(
...
   HandleScope scope;
-  v8::Local<v8::FunctionTemplate> t = New<v8::FunctionTemplate>(callback);
+  v8::Local<v8::FunctionTemplate> tpl = New<v8::FunctionTemplate>(callback);
   v8::Local<v8::String> fn_name = New(name).ToLocalChecked();
-  t->SetClassName(fn_name);
-  templ->Set(fn_name, t);
+  tpl->SetClassName(fn_name);
+  // Note(@agnat): Pass an empty T* as discriminator. See note on
+  // SetMethodAux(...) above
+  imp::SetMethodAux(recv, fn_name, tpl, (T*)0);
```
bnoordhuis: Can you use `static_cast<T*>(0)`?

nan.h
```diff
@@ -1852,75 +1852,47 @@ NAN_INLINE void SetInstanceTemplate(
...
 
+template <typename T, template <typename> class HandleType>
 NAN_INLINE void SetMethod(
-    v8::Local<v8::ObjectTemplate> templ
+    HandleType<T> recv
   , const char *name
   , FunctionCallback callback) {
   HandleScope scope;
-  v8::Local<v8::FunctionTemplate> t = New<v8::FunctionTemplate>(callback);
+  v8::Local<v8::FunctionTemplate> tpl = New<v8::FunctionTemplate>(callback);
```
bnoordhuis: Renaming the local makes the diff a little noisier than it needs to be.

bnoordhuis:
It's going to be interesting to see if VS 201[023] accept this.

kkoopa:
Visual Studio 2013 has no problems. https://ci.appveyor.com/project/RodVagg/nan/build/1157

nan.h
agnat: Yeah, the diff is all over the place. Not sure why though. I didn't rename the local. Diff just *thinks* I did. :-/

I'll see what I can do.

nan.h
agnat: Sure. Personally in this case I like `(T*)0` better because it so ... idiomatic. Anyway...

kkoopa:
Looking good. I like it. Much cleaner than unrolling everything. I would think this should work on older VS compilers than that in 2013, but have no easy way of testing it. However, since this whole workaround is only strictly necessary for Node 6 and later, and already Node 4 requires a C++11 compiler, it could even be ifdefed around in case it turns out that it break some old compiler for some old version of Node and someone cared enough about it.

@bnoordhuis You'll want something like this for Node core as well, since `NODE_SET_METHOD` suffers from the same problem. There should be no worry about outdated VS versions there, as Node cannot be built with less than VS 2013.

mathiask88:
@kkoopa You could try to add the `--msvs_version=2010` argument to the appveyor.yml in this PR and start a build for this PR. They got VS 08 10 12 13 15 installed on their systems.

agnat:
I'm confident that it will just work... right down to historic, single digit versions of VS. ;)

I didn't invent it. I've seen it somewhere... in a high-quality code base... probably boost. If it really _does_ act up I'll come up with something (else). TIMTOWTDI.

I removed my beautiful application of a C cast in C++, ironed the diff to remove some wrinkles like @bnoordhuis suggested, squashed the thing and added a proper commit message. PTAL.

kkoopa:
LGTM. Unless someone raises a valid objection, I'll merge this and push out a new version by tomorrow.

PR #568, opened 2016-05-09
Remove V8 deprecation warnings

matthewloring:
Current deprecation warnings on Node v6 are SetAccessor, CloneElementAt,
and BooleanObject::New.

### DIFF
--- a/nan_implementation_12_inl.h
+++ b/nan_implementation_12_inl.h
@@ -37,7 +37,12 @@ Factory<v8::Boolean>::New(bool value) {
 
 Factory<v8::BooleanObject>::return_t
 Factory<v8::BooleanObject>::New(bool value) {
+#if (NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION)
+  return v8::BooleanObject::New(
+    v8::Isolate::GetCurrent(), value).As<v8::BooleanObject>();
+#else
   return v8::BooleanObject::New(value).As<v8::BooleanObject>();
+#endif
 }
 
 //=== Context ==================================================================
--- a/nan_maybe_43_inl.h
+++ b/nan_maybe_43_inl.h
@@ -217,7 +217,21 @@ NAN_INLINE Maybe<int> GetEndColumn(v8::Local<v8::Message> msg) {
 NAN_INLINE MaybeLocal<v8::Object> CloneElementAt(
     v8::Local<v8::Array> array
   , uint32_t index) {
-  return array->CloneElementAt(GetCurrentContext(), index);
+  v8::Local<v8::Context> context = GetCurrentContext();
+#if (NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION)
+  v8::EscapableHandleScope handle_scope(v8::Isolate::GetCurrent());
+  v8::Local<v8::Value> elem;
+  if (!array->Get(context, index).ToLocal(&elem)) {
+    return MaybeLocal<v8::Object>();
+  }
+  v8::Local<v8::Object> obj;
+  if (!elem->ToObject(context).ToLocal(&obj)) {
+    return MaybeLocal<v8::Object>();
+  }
+  return handle_scope.Escape(obj->Clone());
+#else
+  return array->CloneElementAt(context, index);
+#endif
 }
 
 NAN_INLINE MaybeLocal<v8::Value> Call(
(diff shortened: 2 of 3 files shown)


nan_implementation_12_inl.h
```diff
@@ -37,7 +37,11 @@ Factory<v8::Boolean>::New(bool value) {
 
 Factory<v8::BooleanObject>::return_t
 Factory<v8::BooleanObject>::New(bool value) {
+#if (NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION)
+  return v8::BooleanObject::New(v8::Isolate::GetCurrent(), value).As<v8::BooleanObject>();
```
bnoordhuis: Can you wrap this at 80 columns?

bnoordhuis:
LGTM with a style nit.

matthewloring:
@bnoordhuis The `CloneElementAt` deprecation message says that `"Cloning is not supported."`. How should this be handled here as there is not an alternative that I am aware of.

nan_implementation_12_inl.h
matthewloring: Done.

kkoopa:
What about `WeakCallbackInfo::IsFirstPass`?

matthewloring:
Hmm, I wasn't seeing a warning for that one. I found these warnings by compiling an arbitrary native module with warnings enabled. Is there an easy way to exercise the full API to make sure I've got all the deprecations? It looks like `WeakCallbackInfo::IsFirstPass` falls into the same category as `CloneElementAt` as having no clear alternative API. How do you recommend dealing with these cases?

kkoopa:
The test suite covers most, but it is far from complete. The removed functionality should be reimplemented without breaking the API. Functional equivalence with no unwanted side effects.

Cloning an array element should be fairly straightforward, as it only needs special logic for `v8::Object`s, although I do not know what all is cloned. The weak callback logic will require a fair amount of rewriting, since it appears to have been unreliable forever.

matthewloring:
@kkoopa I've got a reimplementation of `CloneElementAt`. Both `Clone` and `CloneElementAt` eventually call `CopyJSObject` so the should share copy semantics. `IsFirstPass` will require more work. It may be good to address that one in a larger follow on PR.

kkoopa:
Sounds good. IsFirstPass might be reworkable by adding a constant boolean template parameter to the function, essentially turning it into two functions.

nan_maybe_43_inl.h
```diff
@@ -217,7 +217,13 @@ NAN_INLINE Maybe<int> GetEndColumn(v8::Local<v8::Message> msg) {
 NAN_INLINE MaybeLocal<v8::Object> CloneElementAt(
     v8::Local<v8::Array> array
   , uint32_t index) {
+#if (NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION)
+  v8::EscapableHandleScope handle_scope(v8::Isolate::GetCurrent());
+  v8::Local<v8::Object> elem = array->Get(index)->ToObject()->Clone();
```
kkoopa: The non-maybe versions of `Get` and `ToObject` are scheduled for deprecation.

matthewloring: I've updated these to use the maybe versions.

nan_maybe_43_inl.h
```diff
@@ -217,7 +217,21 @@ NAN_INLINE Maybe<int> GetEndColumn(v8::Local<v8::Message> msg) {
 NAN_INLINE MaybeLocal<v8::Object> CloneElementAt(
     v8::Local<v8::Array> array
   , uint32_t index) {
-  return array->CloneElementAt(GetCurrentContext(), index);
+  v8::Local<v8::Context> context = GetCurrentContext();
+#if (NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION)
+  v8::EscapableHandleScope handle_scope(v8::Isolate::GetCurrent());
```
kkoopa: The context handle leaks to the surrounding scope, since it is created before the scope.

matthewloring: Fixed.

nan_maybe_43_inl.h
```diff
@@ -230,7 +230,7 @@ NAN_INLINE MaybeLocal<v8::Object> CloneElementAt(
   }
   return handle_scope.Escape(obj->Clone());
 #else
-  return array->CloneElementAt(context, index);
+  return array->CloneElementAt(GetCurrentContext(), index);
```
kkoopa: I think the handle returned from `GetCurrentContext` now leaks here instead. It would be better to move the handle scope out before the preprocessor block.

matthewloring: Fixed.

matthewloring:
@kkoopa Do you want any further changes on this PR?

kkoopa:
You could escape the scope before returning on line 233, other than that it looks good to merge. However, I have been waiting for someone to review #569, since it also is necessary to avoid deprecation warnings.

matthewloring:
All fixed.

PR #569, opened 2016-05-10
Fix new versions not to use WeakCallbackInfo::IsFirstPass

kkoopa:
Follow-up to #568.

### DIFF
--- a/nan_weak.h
+++ b/nan_weak.h
@@ -90,8 +90,16 @@ class WeakCallbackInfo {
   static WeakCallbackInfo *unwrap(NAN_WEAK_CALLBACK_DATA_TYPE_ data);
 # endif
 #else
+# if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                     \
+  (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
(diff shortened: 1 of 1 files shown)


bnoordhuis:
LGTM, I think.

kkoopa:
Good. Thanks for looking through it. This was the least invasive solution I could come up with. It should even have better performance than the old implementation, thanks to dead code elimination. CI also passes, but in somehow the hook part showing CI status in PRs has disappeared in the last months.
I'll merge this and the related PR this week and publish a new patch version.

@rvagg Could you figure out what's what with the Travis and AppVeyor hooks for PRs?

rvagg:
Unfortunately the nodejs org doesn't allow third-party applications, it's been a policy for a while but tbh I don't recall how it started or exactly what the justification is. That's why Travis and AppVeyor can't post status back via the GH API.

kkoopa:
Aha, well that explains why it stopped working. Cannot access be granted for specific repositories? The current state is quite error prone, as it is easy to forget checking builds manually.

bnoordhuis:
It depends on the permissions that the third-party application wants.  In the case of Travis CI, it wanted (and presumably still wants) permissions that lets it access private repositories.  Because our private repositories contain sensitive information, and because we can't trust a third-party not to get hacked, that makes it a no-go.

PR #572, opened 2016-05-25
Update nan_object_wrap.h

laurie71:
Add a const accessor. This was needed to access the handle in a const context.

### DIFF
--- a/nan_object_wrap.h
+++ b/nan_object_wrap.h
@@ -39,6 +39,11 @@ class ObjectWrap {
   }
 
 
+  inline v8::Local<v8::Object> const_handle() const {
+    return New(handle_);
+  }
+
+
   inline v8::Local<v8::Object> handle() {
     return New(persistent());
   }


nan_object_wrap.h
```diff
@@ -39,6 +39,11 @@ class ObjectWrap {
   }
 
 
+  inline v8::Local<v8::Object> const_handle() const {
+    return New(handle_);
+  }
+
+
```
bnoordhuis: Why not make the existing `handle()` method const?

bnoordhuis:
A regression test would be nice.  Also, "Update nan_object_wrap.h" is not a great status line for a commit log.

kkoopa:
Since the interface is the same as node::ObjectWrap, is there a reason that is not const?

bnoordhuis:
Legacy, mostly.  It could be made const now.

kkoopa:
Fixed in #574.

PR #574, opened 2016-05-28
Make ObjectWrap::handle() const

kkoopa:
Correct solution to #572 with regression test.


bnoordhuis:
LGTM

agnat:
What @bnoordhuis says.

kkoopa:
Thanks for quick action.

Issue #576, opened 2016-05-31
nan 2.3.1 broke ref on node v0.10 / gcc 4.3.2


ds82:
```
> ref@1.3.2 install /home/buildbot/jenkins/workspace/rea-jet-firmware-etx/origin/V3.61/titan-web/modules/rea-renderer/node_modules/ref
> node-gyp rebuild

make: Entering directory `/home/buildbot/jenkins/workspace/rea-jet-firmware-etx/origin/V3.61/titan-web/modules/rea-renderer/node_modules/ref/build'
  CXX(target) Release/obj.target/binding/src/binding.o
../node_modules/nan/nan.h: In function ‘void init(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE)’:
../node_modules/nan/nan.h:1877: sorry, unimplemented: inlining failed in call to ‘void Nan::imp::SetMethodAux(T, v8::Local<v8::String>, v8::Local<v8::FunctionTemplate>, ...) [with T = v8::Handle<v8::Object>]’: function not inlinable
../node_modules/nan/nan.h:1894: sorry, unimplemented: called from here
../node_modules/nan/nan.h:1877: sorry, unimplemented: inlining failed in call to ‘void Nan::imp::SetMethodAux(T, v8::Local<v8::String>, v8::Local<v8::FunctionTemplate>, ...) [with T = v8::Handle<v8::Object>]’: function not inlinable
../node_modules/nan/nan.h:1894: sorry, unimplemented: called from here
../node_modules/nan/nan.h:1877: sorry, unimplemented: inlining failed in call to ‘void Nan::imp::SetMethodAux(T, v8::Local<v8::String>, v8::Local<v8::FunctionTemplate>, ...) [with T = v8::Handle<v8::Object>]’: function not inlinable
../node_modules/nan/nan.h:1894: sorry, unimplemented: called from here
../node_modules/nan/nan.h:1877: sorry, unimplemented: inlining failed in call to ‘void Nan::imp::SetMethodAux(T, v8::Local<v8::String>, v8::Local<v8::FunctionTemplate>, ...) [with T = v8::Handle<v8::Object>]’: function not inlinable
../node_modules/nan/nan.h:1894: sorry, unimplemented: called from here
[...]

PR #577, opened 2016-05-31
Replace NAN_INLINE with 'inline' keyword.

bnoordhuis:
There is no reason to mark everything  `__attribute__((always_inline))`,
just use the regular keyword and let the compiler decide.  Avoids code
bloat and build errors when inlining fails due to compiler limitations.

Fixes: https://github.com/nodejs/nan/issues/576

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -84,14 +84,6 @@
 
 namespace Nan {
 
-#if defined(__GNUC__) && !(defined(DEBUG) && DEBUG)
-# define NAN_INLINE inline __attribute__((always_inline))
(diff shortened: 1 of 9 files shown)


kkoopa:
Good, apart from the actual removal of the macro. Some addons use it themselves, since it is exported due to being a macro. Add a comment that it should be removed in the next major.

bnoordhuis:
Fix-up added.

agnat:
> just use the regular keyword and let the compiler decide.

+1. LGTM.

PR #579, opened 2016-06-03
Correct NAN_PROPERTY_SETTER_ARGS_TYPE to match with prototype needed by Nan::SetAccessor().

Flarna:
If I use NAN_PROPERTY_SETTER I get compile errors (at least in MSVC) that functions doesn't match to Nan::SetAccessor().
I adapted NAN_PROPERTY_SETTER_ARGS_TYPE to correct this.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1260,7 +1260,7 @@ typedef const PropertyCallbackInfo<v8::Value>&
     NAN_PROPERTY_GETTER_ARGS_TYPE;
 typedef void NAN_PROPERTY_GETTER_RETURN_TYPE;
 
-typedef const PropertyCallbackInfo<v8::Value>&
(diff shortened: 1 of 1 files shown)


kkoopa:
No, this is wrong. It is correct as it is. A property setter is not an accessor, but a named or indexed property interceptor. This is explained in the documentation and the examples. You are getting compile errors because you are combining the wrong things. It is NAN_SETTER that goes with SetAccessor.

Flarna:
ok thanks. I clearly mixed this up.
sorry!

kkoopa:
An honest mistake. It is not the first time this particular issue has caused confusion, since interceptors are fairly uncommon. They correspond to ES6 proxies.

PR #581, opened 2016-06-12
Linted code

kkoopa:


bnoordhuis:
The explicitness vs. implicitness of constructors seems a bit random.  What's the pattern?

kkoopa:
V8 has MaybeLocal do conversion. I was a bit befuddled regarding Persistents, but tried to stick to that of V8 3.19 or so, which was the last one with single argument constructors. Some of them were explicit, others were not.

bnoordhuis:
Most things are explicit now and personally I think that's better, particularly with persistent handles, because there is less opportunity for accidental memory leaks.  Possibly less of an issue with v4 and newer than it is with v0.10 and v0.12 because v8::Persistent is by default non-copyable.

`MaybeLocal`, like you point out, is the odd duck out for some reason.

nan_persistent_12_inl.h
```diff
@@ -14,11 +14,12 @@ template<typename T, typename M> class Persistent :
...
 
-  template<typename S> inline Persistent(v8::Local<S> that) :
+  template<typename S> explicit inline Persistent(v8::Local<S> that) :
       v8::Persistent<T, M>(v8::Isolate::GetCurrent(), that) {}
 
   template<typename S, typename M2>
-  inline Persistent(const v8::Persistent<S, M2> &that) :
+  inline
+  Persistent(const v8::Persistent<S, M2> &that) :  // NOLINT(runtime/explicit)
       v8::Persistent<T, M2>(v8::Isolate::GetCurrent(), that) {}
```
kkoopa: I guess this could be made explicit.

nan_persistent_12_inl.h
```diff
@@ -62,11 +63,12 @@ class Global : public v8::Global<T> {
...
   inline Global() : v8::Global<T>() {}
 
-  template<typename S> inline Global(v8::Local<S> that) :
+  template<typename S> inline explicit Global(v8::Local<S> that) :
     v8::Global<T>(v8::Isolate::GetCurrent(), that) {}
 
   template<typename S>
-  inline Global(const v8::PersistentBase<S> &that) :
+  inline
+  Global(const v8::PersistentBase<S> &that) :  // NOLINT(runtime/explicit)
```
kkoopa: While this could be made explicit, it is not strictly necessary, as `Globals` are reset when they go out of scope. Maybe we should prefer explicit.

nan_persistent_12_inl.h
```diff
@@ -96,11 +98,12 @@ class Global : public v8::UniquePersistent<T> {
...
   inline Global() : v8::UniquePersistent<T>() {}
 
-  template<typename S> inline Global(v8::Local<S> that) :
+  template<typename S> inline explicit Global(v8::Local<S> that) :
     v8::UniquePersistent<T>(v8::Isolate::GetCurrent(), that) {}
 
   template<typename S>
-  inline Global(const v8::PersistentBase<S> &that) :
+  inline
+  Global(const v8::PersistentBase<S> &that) :  // NOLINT(runtime/explicit)
```
kkoopa: This should follow L71

nan_persistent_pre_12_inl.h
```diff
@@ -196,20 +196,20 @@ class Global : public PersistentBase<T> {
...
   template <typename S>
-  inline Global(v8::Local<S> that)
+  inline explicit Global(v8::Local<S> that)
       : PersistentBase<T>(v8::Persistent<T>::New(that)) {
     TYPE_CHECK(T, S);
   }
 
   template <typename S>
-  inline Global(const PersistentBase<S> &that)
+  inline Global(const PersistentBase<S> &that)  // NOLINT(runtime/explicit)
```
kkoopa: This should follow the newer `Global` in the other file.

nan_persistent_pre_12_inl.h
```diff
@@ -196,20 +196,20 @@ class Global : public PersistentBase<T> {
...
     : PersistentBase<T>(that) {
     TYPE_CHECK(T, S);
   }
   /**
    * Move constructor.
    */
-  inline Global(RValue rvalue)
+  inline Global(RValue rvalue)  // NOLINT(runtime/explicit)
```
kkoopa: This should be a conversion, as it emulates a move constructor.

kkoopa:
There, turned out there were rather many wrong things which had been masked by implicit conversion. Now it should be sorted out.

nan_persistent_12_inl.h
```diff
@@ -14,25 +14,31 @@ template<typename T, typename M> class Persistent :
  public:
   inline Persistent() : v8::Persistent<T, M>() {}
 
-  template<typename S> inline Persistent(v8::Local<S> that) :
+  template<typename S> inline explicit  Persistent(v8::Local<S> that) :
```
bnoordhuis: There are two spaces here.

nan_persistent_12_inl.h
```diff
@@ -81,6 +87,8 @@ class Global : public v8::Global<T> {
     v8::PersistentBase<T>::Reset(v8::Isolate::GetCurrent(), other);
   }
 
+  Global Pass() { return static_cast<Global&&>(*this); }  // NOLINT(build/c++11)
```
bnoordhuis: Shouldn't that be in an `#if __cplusplus >= 201103L` to avoid breaking g++ 4.7 and older?  v0.10 and v0.12 support g++ 4.2.

kkoopa: This implementation of `Global` is conditionally compiled only for V8 4.3 and up.

nan_persistent_12_inl.h
```diff
@@ -93,28 +101,57 @@ class Global : public v8::Global<T> {
...
 
-  template <typename S>
+  template<typename S>
   inline void Reset(const v8::PersistentBase<S> &other) {
     v8::PersistentBase<T>::Reset(v8::Isolate::GetCurrent(), other);
   }
 
+#if NODE_MODULE_VERSION == NODE_0_12_MODULE_VERSION
+  inline void Empty() {
+    reinterpret_cast<v8::Persistent<T>*>(this)->ClearAndLeak();
```
bnoordhuis: This is bordering on illegal, of course, what with `UniquePersistent` not being an instance of `Persistent`.  I assume that's the only way to leak the internal pointer?

bnoordhuis:
I think I understand why the code is doing what it's doing but it's really skirting UB territory now, isn't it?

nan_persistent_12_inl.h
kkoopa: Yes, the only other way would be to directly memset the whole object to 0. So this is the lesser of two evils.

kkoopa:
Skirting, but never crossing. Nothing is virtual and no child has added any data members. Additionally, the horrible things only occur for old versions, which will not change retroactively.

nan_persistent_12_inl.h
kkoopa: I can also add that `v8::Persistent::ClearAndLeak` was moved to `v8::PersistentBase::Empty` after V8 @ Node 0.12, but already in V8 @ io.js 1.0. It was changed to no longer return the encapsulated `T*`. However, since `v8::Persistent` is a `v8::PersistentBase` and `v8::UniquePersistent` is a `v8::PersistentBase` and the sole `val_` member resides in `v8::PersistentBase`, this is correct (in the loosest sense of correct). Nothing is virtual.

kkoopa:
I changed the casts to `static_cast` and added some tests to help put your mind at ease.

nan_maybe_pre_43_inl.h
```diff
@@ -16,9 +16,10 @@ class MaybeLocal {
 
   template<typename S>
 # if NODE_MODULE_VERSION >= NODE_0_12_MODULE_VERSION
-  inline MaybeLocal(v8::Local<S> that) : val_(that) {}
+  inline MaybeLocal(v8::Local<S> that) :  // NOLINT(runtime/explicit)
+      val_(that) {}
 # else
-  inline MaybeLocal(v8::Local<S> that) :
+  inline MaybeLocal(v8::Local<S> that) :  // NOLINT(runtime/explicit)
       val_(*reinterpret_cast<v8::Local<T>*>(&that)) {}
```
bnoordhuis: Can you use `val_(v8::Local<T>(that))` here?  V8's constructor does a type check that catches most mistakes.

nan_persistent_12_inl.h
```diff
@@ -81,47 +85,80 @@ class Global : public v8::Global<T> {
...
+        rvalue.object_))->ClearAndLeak();
+# endif
+  }
+
+  template<typename S>
+  inline Global &operator=(v8::UniquePersistent<S> other) {
+    return static_cast<Global&>(v8::UniquePersistent<S>::operator=(other));
+  }
+
+  inline operator RValue() { return RValue(this); }
```
bnoordhuis: Is the implicit conversion operator necessary?  I don't see it being used anywhere.

Aside: it looks like if you add a private `Global(Global* that)` constructor, you can drop the `RValue` type and simply call `that->Empty()`.

nan_maybe_pre_43_inl.h
kkoopa: That would leak a handle to the surrounding scope. I can do a `TYPE_CHECK` in the body instead.

nan_persistent_12_inl.h
kkoopa: It is necessary for move emulation. The compiler sees the (private) copy constructor of non-const ref. It does not work with a temporary, so it looks for other options. It finds the `Global(RValue)` constructor and looks for a conversion to `RValue`. The conversion is provided by the conversion operator, so this succeeds. The `Global(RValue)` constructor then moves ownership of the encapsulated value to `this`.

nan_maybe_pre_43_inl.h
bnoordhuis: The conversion constructor doesn't allocate a HandleScope slot.  A `TYPE_CHECK` is fine too, though.

bnoordhuis:
LGTM

kkoopa:
Thanks for looking through this.

I did find an edge case in that it is now not reasonably possible to actually copy a `Global` from another `Global`. It will require a `static_cast` from `Global` to `PersistentBase`, which is presumably unavoidable and can be lived with. However, due to the way things are set up, that does not work either, since `Global` inherits from `v8::PersistentBase` but not from `Nan::PersistentBase`.

The best solution ought to be the normalization process from #492, which would set up the correct inheritance hierarchy, remove some code duplication and make everything better in the long run. I will open a new PR with these commits plus the additional ones.

PR #582, opened 2016-06-12
Add Callback::Reset

kkoopa:
This is useful if you want to set and unset a `Callback`.

CI: https://travis-ci.org/nodejs/nan/builds/137235515

### DIFF
--- a/doc/callback.md
+++ b/doc/callback.md
@@ -35,6 +35,10 @@ class Callback {
 
   bool IsEmpty() const;
 
+  void Reset(const v8::Local<v8::Function> &fn);
+
+  void Reset();
+
   v8::Local<v8::Value> Call(v8::Local<v8::Object> target,
                             int argc,
                             v8::Local<v8::Value> argv[]) const;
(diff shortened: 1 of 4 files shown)


rvagg:
lgtm, but maybe `Reset` would be more familiar?

kkoopa:
I thought of Reset, but decided not to. Reset should also allow setting the function, this would make SetFunction useless, removing SetFunction would be a major change.

It could also be thought of as clearing the underlying Persistent, which it does not.

As such, I thought Clear was the better name for the time being.

bnoordhuis:
You could add a Reset() overload that takes a `Local<Function>` and make SetFunction() a deprecated alias.  Not saying you should, just that it's possible. :-)

kkoopa:
I was thinking of doing that in autumn or so for the next major, sans deprecation. At the same time parts should be maybefied and more stuff fixed overall. This is a stopgap measure.

I also don't remember why the Function is stored as a property on a Persistent Object. Why is it not stored directly? If that were the case, Reset would be wholly sensible.

kkoopa:
There, now it is called `Reset`, and the `Callback` is cleaner.

bnoordhuis:
LGTM

PR #588, opened 2016-06-30
AsyncProgressWorker: add template types for .send

eljefedelrodeodeljefe:
In order to implement a quick way to a JS stream, sending arbitrary types in ProgressWorker would be useful. This adds support for exactly that and a Stream example.

However for prettiness and compat, help would be appreciated. I couldn't make fully backwards compatible unfortunately. Hence this would semver major, and would require changes as displayed in the adapted tests.

**From commit:**
This allows for arbitrary types to be sent from .send. To achieve this, the
abstract class can be implemented with a template itself. Either at
implementation or instantiation a type MUST be declared though.

Due to the limitations and / or design of C++ it is not easily possible to
fallback to the old behaviour. In consequence this is a semver major change.
An easy semi-fallback would be to implement the class as

      class ProgressWorker : public AsyncProgressWorker<> {}

which falls back to the `char` case. However HandleProgressCallback can then not
take `const char` as first argument, and needs char.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1589,12 +1589,9 @@ class Callback {
   char *errmsg_;
 };
 
-// temaplate default is mainly set to have as little friction to prior
-// implementation of AsyncProgressWorkerBase. The problem arises when the user needs
-// to send non-uniform types like the previous cons declaration in
-// HandleProgressCallback. Though this attempts backwards compatibility, cpp
-// does not allow for overfloaded pure virtual methods and conditional const or
-// de-const-ing.
+// The template default is mainly set to have as little friction to the prior
+// implementations of AsyncProgressWorker. In conjunction with the respective
+// typedef below, it allows for backwards compatibility.
 template<class T = char>
 /* abstract */ class AsyncProgressWorkerBase : public AsyncWorker {
  public:
@@ -1624,7 +1621,7 @@ template<class T = char>
     asyncdata_ = NULL;
     uv_mutex_unlock(&async_lock);
 
-    // Dont send progress events after we've already completed.
+    // Don't send progress events after we've already completed.
     if (callback) {
         HandleProgressCallback(data, size);
     }
--- a/nan.h
+++ b/nan.h
@@ -1590,15 +1590,15 @@ class Callback {
 };
 
 // temaplate default is mainly set to have as little friction to prior
-// implementation of AsyncProgressWorker. The problem arises when the user needs
+// implementation of AsyncProgressWorkerBase. The problem arises when the user needs
 // to send non-uniform types like the previous cons declaration in
 // HandleProgressCallback. Though this attempts backwards compatibility, cpp
 // does not allow for overfloaded pure virtual methods and conditional const or
 // de-const-ing.
 template<class T = char>
-/* abstract */ class AsyncProgressWorker : public AsyncWorker {
+/* abstract */ class AsyncProgressWorkerBase : public AsyncWorker {
  public:
-  explicit AsyncProgressWorker(Callback *callback_)
+  explicit AsyncProgressWorkerBase(Callback *callback_)
       : AsyncWorker(callback_), asyncdata_(NULL), asyncsize_(0) {
     async = new uv_async_t;
     uv_async_init(
@@ -1611,7 +1611,7 @@ template<class T = char>
     uv_mutex_init(&async_lock);
   }
 
-  virtual ~AsyncProgressWorker() {
+  virtual ~AsyncProgressWorkerBase() {
     uv_mutex_destroy(&async_lock);
 
     delete[] asyncdata_;
@@ -1632,24 +1632,24 @@ template<class T = char>
   }
 
   class ExecutionProgress {
-    friend class AsyncProgressWorker;
+    friend class AsyncProgressWorkerBase;
    public:
     void Signal() const {
         uv_async_send(that_->async);
     }
 
-    void Send(T* data, size_t size) const {
+    void Send(const T* data, size_t size) const {
         that_->SendProgress_(data, size);
     }
 
    private:
-    explicit ExecutionProgress(AsyncProgressWorker* that) : that_(that) {}
+    explicit ExecutionProgress(AsyncProgressWorkerBase* that) : that_(that) {}
     NAN_DISALLOW_ASSIGN_COPY_MOVE(ExecutionProgress)
-    AsyncProgressWorker* const that_;
+    AsyncProgressWorkerBase* const that_;
   };
 
   virtual void Execute(const ExecutionProgress& progress) = 0;
-  virtual void HandleProgressCallback(T *data, size_t size) = 0;
+  virtual void HandleProgressCallback(const T *data, size_t size) = 0;
 
   virtual void Destroy() {
       uv_close(reinterpret_cast<uv_handle_t*>(async), AsyncClose_);
@@ -1661,7 +1661,7 @@ template<class T = char>
       Execute(progress);
   }
 
-  void SendProgress_(T *data, size_t size) {
+  void SendProgress_(const T *data, size_t size) {
     T *new_data = new T[size];
     memcpy(new_data, data, size);
 
@@ -1676,14 +1676,14 @@ template<class T = char>
   }
 
   inline static NAUV_WORK_CB(AsyncProgress_) {
-    AsyncProgressWorker *worker =
-            static_cast<AsyncProgressWorker*>(async->data);
+    AsyncProgressWorkerBase *worker =
+            static_cast<AsyncProgressWorkerBase*>(async->data);
     worker->WorkProgress();
   }
 
   inline static void AsyncClose_(uv_handle_t* handle) {
-    AsyncProgressWorker *worker =
-            static_cast<AsyncProgressWorker*>(handle->data);
+    AsyncProgressWorkerBase *worker =
+            static_cast<AsyncProgressWorkerBase*>(handle->data);
     delete reinterpret_cast<uv_async_t*>(handle);
     delete worker;
   }
@@ -1693,6 +1693,8 @@ template<class T = char>
   T *asyncdata_;
   size_t asyncsize_;
 };
+
+typedef AsyncProgressWorkerBase<> AsyncProgressWorker;
 
 inline void AsyncExecute (uv_work_t* req) {
   AsyncWorker *worker = static_cast<AsyncWorker*>(req->data);
--- a/nan.h
+++ b/nan.h
@@ -1640,7 +1640,7 @@ template<class T = char>
     }
 
    private:
-    explicit ExecutionProgress(AsyncProgressWorkerBase* that) : that_(that) {}
+    explicit ExecutionProgress(AsyncProgressWorkerBase *that) : that_(that) {}
     NAN_DISALLOW_ASSIGN_COPY_MOVE(ExecutionProgress)
     AsyncProgressWorkerBase* const that_;
   };
@@ -1694,6 +1694,8 @@ template<class T = char>
   size_t asyncsize_;
 };
 
+// This ensures compatibility to the previous un-templated AsyncProgressWorker
+// class definition.
 typedef AsyncProgressWorkerBase<> AsyncProgressWorker;
 
 inline void AsyncExecute (uv_work_t* req) {
--- a/nan.h
+++ b/nan.h
@@ -1660,7 +1660,10 @@ template<class T = char>
 
   void SendProgress_(const T *data, size_t size) {
     T *new_data = new T[size];
-    memcpy(new_data, data, size);
+    {
+      T *it = new_data;
+      std::copy(data, data + size, it);
+    }
 
     uv_mutex_lock(&async_lock);
     T *old_data = asyncdata_;
--- a/nan.h
+++ b/nan.h
@@ -1589,9 +1589,7 @@ class Callback {
   char *errmsg_;
 };
 
-// The template default is mainly set to have as little friction to the prior
-// implementations of AsyncProgressWorker. In conjunction with the respective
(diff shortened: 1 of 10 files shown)


nan.h
```diff
@@ -1589,6 +1589,13 @@ class Callback {
   char *errmsg_;
 };
 
+// temaplate default is mainly set to have as little friction to prior
```
bnoordhuis: s/temaplate/Template/

nan.h
```diff
@@ -1589,6 +1589,13 @@ class Callback {
...
+// implementation of AsyncProgressWorker. The problem arises when the user needs
+// to send non-uniform types like the previous cons declaration in
```
bnoordhuis: cons?

nan.h
```diff
@@ -1589,6 +1589,13 @@ class Callback {
...
+// HandleProgressCallback. Though this attempts backwards compatibility, cpp
+// does not allow for overfloaded pure virtual methods and conditional const or
```
bnoordhuis: overloaded

nan.h
```diff
@@ -1589,6 +1589,13 @@ class Callback {
...
+// de-const-ing.
+template<class T = char>
 /* abstract */ class AsyncProgressWorker : public AsyncWorker {
```
bnoordhuis: I'd add a `struct AsyncProgressWorker : public AsyncProgressWorker<char> {};` for the sake of compatibility.

test/cpp/asyncprogressworkerstream.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#ifndef _WIN32
+#include <unistd.h>
+#define Sleep(x) usleep((x)*1000)
+#endif
+#include <nan.h>
+
+using namespace Nan;  // NOLINT(build/namespaces)
+
+// custom data type. This servers as an example of how external
```
bnoordhuis: s/custom/Custom/ and s/servers/serves/

test/cpp/asyncprogressworkerstream.cpp
```diff
@@ -0,0 +1,83 @@
...
+// libraries could be hooked in, populate their objects and send them to JS
+struct data_t {
+  int index;
+  int data;
+};
+
+// unlike test ayyncprogressworker.cpp this test is explictely templated
```
bnoordhuis: s/explictely/explicitly/.  I'll stop pointing it out from now on but can you capitalize and punctuate comments?

test/cpp/asyncprogressworkerstream.cpp
```diff
@@ -0,0 +1,83 @@
...
+    : AsyncProgressWorker<T>(callback), progress(progress)
+    , milliseconds(milliseconds), iters(iters) {}
+  ~ProgressWorker() {}
+
+  void Execute (const typename AsyncProgressWorker<T>::ExecutionProgress& progress) {
+    data_t data;
+    for (int i = 0; i < iters; ++i) {
+      data.index = i;
+      data.data = i * 2;
+      progress.Send(&data, sizeof(data_t));
```
bnoordhuis: `sizeof(data)`

bnoordhuis:
Basic concept seems okay to me.

nan.h
```diff
@@ -1654,12 +1661,12 @@ class Callback {
       Execute(progress);
   }
 
-  void SendProgress_(const char *data, size_t size) {
-    char *new_data = new char[size];
+  void SendProgress_(T *data, size_t size) {
+    T *new_data = new T[size];
     memcpy(new_data, data, size);
```
bnoordhuis: One potential pitfall is that memcpy and non-POD types don't mix well.

test/cpp/asyncprogressworkerstream.cpp
kkoopa: with space, it is a keyword, not a function

nan.h
kkoopa: Well, it is guaranteed to work for anything that satisfies `std::is_trivially_copy_constructable`, but that trait is not available in C++03. I do agree that this cannot stand as-is.

kkoopa:
If this is going to be a breaking change, it will not happen until NAN 3, which is likely to be a near complete rewrite again, since a lot of legacy support will be dropped and many things will become unnecessary. However, that will not happen until  autumn.

To get this in now, would it not be possible to break out `AsyncProgressWorker` into a base class and one with the current interface, then add this as a sibling? Or use a specialization for the `char` case.

test/cpp/asyncprogressworkerstream.cpp
eljefedelrodeodeljefe: Will review all texts again. Not the best day for spelling, sorry.

eljefedelrodeodeljefe:
@kkoopa I was playing around with this in more than one way, Including a base class. Easiest would have seemed to me, to overload `HandleProgressCallback`, but I couldn't make work properly. I guess also that is not the way cpp was designed. If you and @bnoordhuis like the general idea and have the intention to merge it in some way, I give compat another run. There must be _one_ way.

nan.h
eljefedelrodeodeljefe: I will likely have questions about this at some later point, if google doesn't yield enough information. I assume trying to compile with C++03 will break it, hence makes it testable?

bnoordhuis: The simple solution would be to use `std::copy()` from `<algorithm>`, i.e.:
```c++
{
  T *it = new_data;
  std::copy(data, data + size, it);
}
```

kkoopa: Why not just do T *it = new T(*data);

Flarna:
Isn't it possible to use `const T*` as argument of `HandleProgressCallback()`, `SendProgress_()` and `Send()` to avoid the compatibility issue?

nan.h
kkoopa: It will unfortunstely not break compilation. The problem is that it is undefined behavior unless src is trivially copy constructible. That template could be used in a static assert to ensure a compilation error if invoked with a bad argument, but since said template is not available it cannot be used directly.

bnoordhuis: I must confess I don't quite understand what you're getting at.

kkoopa: Nevermind. I missed that it is supposed to be an array. I was only reading the e-mail notifications, missing out on the full context.

Flarna: 3rd argument of memcpy takes number of bytes but size is the number of elements now therefore this would work only for types with sizeof(T) == 1.

eljefedelrodeodeljefe:
@Flarna good point. Overlooked that. @kkoopa that leaves the API intact. To use the API addition one would need to extend from `AsyncProgressWorkerBase` though. There now is a typedef for the old API https://github.com/nodejs/nan/pull/588/files#diff-558c37be49a47656e3ff41806c24bd33R1697

eljefedelrodeodeljefe:
If I read this right maybe even the renaming is not necessary, but I can't make anything practical out of [type aliases](http://en.cppreference.com/w/cpp/language/type_alias) anymore. Suggestions are welcomed.

@bnoordhuis addressed nits. Pending the `memcpy`-thing

nan.h
```diff
@@ -1654,12 +1658,15 @@ class Callback {
...
 
-  void SendProgress_(const char *data, size_t size) {
-    char *new_data = new char[size];
-    memcpy(new_data, data, size);
+  void SendProgress_(const T *data, size_t size) {
+    T *new_data = new T[size];
+    {
+      T *it = new_data;
+      std::copy(data, data + size, it);
+    }
```
eljefedelrodeodeljefe: @bnoordhuis I assume you intended this? 

With that everything should be addressed cc @kkoopa

bnoordhuis: Yep, that's what I meant.

kkoopa:
On the subject of type aliases, they are a C++11 feature. Thus, no good.

test/cpp/asyncprogressworkerstream.cpp
```diff
@@ -0,0 +1,83 @@
...
+      data.data = i * 2;
+      progress.Send(&data, sizeof( data ));
+      Sleep(milliseconds);
+    }
+  }
+
+  void HandleProgressCallback(const T *data, size_t size) {
+    HandleScope scope;
+    v8::Local<v8::Object> obj = Nan::New<v8::Object>();
+    obj->Set(
```
kkoopa: Use `Nan::Set` here and below.

nan.h
```diff
@@ -1604,45 +1608,45 @@ class Callback {
...
-    // You could do fancy generics with templates here.
-    void Send(const char* data, size_t size) const {
+
+    void Send(const T* data, size_t size) const {
         that_->SendProgress_(data, size);
     }
 
    private:
-    explicit ExecutionProgress(AsyncProgressWorker* that) : that_(that) {}
+    explicit ExecutionProgress(AsyncProgressWorkerBase* that) : that_(that) {}
```
kkoopa: Switch the asterisk and the space. It was wrong before, better make it right this time around. 
`AsyncProgressWorkerBase *that`

nan.h
```diff
@@ -1604,45 +1608,45 @@ class Callback {
...
     NAN_DISALLOW_ASSIGN_COPY_MOVE(ExecutionProgress)
-    AsyncProgressWorker* const that_;
+    AsyncProgressWorkerBase* const that_;
```
kkoopa: I guess this can leave the asterisk where it is, since it would look daft to have it anywhere else while preserving the semantics. As far as I know, this is the only instance in the entire code base of a pointer which is const.

nan.h
```diff
@@ -1669,24 +1676,26 @@ class Callback {
...
   }
 
   uv_async_t *async;
   uv_mutex_t async_lock;
-  char *asyncdata_;
+  T *asyncdata_;
   size_t asyncsize_;
 };
 
+typedef AsyncProgressWorkerBase<> AsyncProgressWorker;
```
kkoopa: I would like a comment here repeating that it is supposed to be there for compatibility.

nan.h
eljefedelrodeodeljefe: * left it here as it is
* changed it above as noted

Others:
* added second comment for compat
* used Nan::Set

kkoopa:
LGTM apart from one small thing:
Please add the new cpp file to the Makefile for linting purposes (it does not use enough wildcard patterns, so there is an explicit list of all files to be  linted).

eljefedelrodeodeljefe:
done. Amended a line that that was 80+ chars.

eljefedelrodeodeljefe:
CI fails (at least) on [Windows for 0.8](https://ci.appveyor.com/project/RodVagg/nan/build/1202/job/pbfxbvkq9oiidphw), because of the Stream-inheriting. Otherwise green.
There is another test that inherits manually from EventEmitter, however same logic doesn't work for `streams` in this case. I know @bnoordhuis has expertise on this issue: can you advise? Just leave it out? I could rather add a full example in `node-addon-examples`

kkoopa:
I do not really do JavaScript, so I cannot advise on that, but I will however note that tests should be as correct and proper as possible in regard to actual usage. Although not (initially) intended as such, they have become a de facto reference and set of examples of how to do addons using NAN, so they should reflect best practices (save for the `using namespace Nan`, which I consider grandfathered for tests) and work in all cases. When something unnatural is done for testing purposes it should be noted that it is for testing purposes only and that it should never be done by anybody else.

Also, everything should work on all supported versions, incuding 0.8, so do make the JavaScript side work, even on 0.8. Everybody do not read node-addon-examples.

From what I can tell from a quick Google search, you would likely need to use the readable-stream module to get streams working on 0.8. It can be added as a dev-dependency. Since there are hopefully very few users of 0.8 left, it would probably be good to somehow condition the usage of said module depending on the version of the runtime, so as to not give people an incorrect impression as to its necessity.

eljefedelrodeodeljefe:
Didn't have any assumptions about policy. Checks out for me. Made it run on 0.8.X locally, by adding the devDependency and conditionally requires it. Thx.

test/js/asyncprogressworkerstream-test.js
```diff
@@ -9,9 +9,14 @@
 const test     = require('tap').test
     , testRoot = require('path').resolve(__dirname, '..')
     , bindings = require('bindings')({ module_root: testRoot, bindings: 'asyncprogressworkerstream' })
-    , Readable = require('stream').Readable
     , util = require('util');
 
+var Readable
+if (process.versions.node.split('.')[1] <= 8)
```
kkoopa: Add a conjunction on the major being 0, else this would trigger for any minor version of 8, e.g. a possible 4.8.

eljefedelrodeodeljefe: Quite right, thanks. Amended.

eljefedelrodeodeljefe:
Added the documentation change.

eljefedelrodeodeljefe:
ping @kkoopa anymore I can do?

kkoopa:
I'll try to get at preparing a new release over the weekend. There was however one thing on my mind:
I think it would be better to remove the default template argument in `AsyncProgressWorkerBase` and change the typedef to `typedef AsyncProgressWorkerBase<char> AsyncProgressWorker`.

eljefedelrodeodeljefe:
Had this in my mind too. Just pushed it. Also I am gonna attempt writing some benchmarks for it. Curious how well the worker performs.

nan.h
```diff
@@ -1592,7 +1592,7 @@ class Callback {
 // The template default is mainly set to have as little friction to the prior
```
kkoopa: There is no template default any more.

kkoopa:
Great, thanks for a good contribution.

eljefedelrodeodeljefe:
thx @kkoopa :). Pushed the removal of the comment.

kkoopa:
Merged as 1242c9a11a7ed481c8f08ec06316385cacc513d0

Issue #587, opened 2016-06-29
Wrap v8::Object::GetPrivate


cscott:
v8 renamed `v8::Object::SetHiddenValue`/`GetHiddenValue`/`DeleteHiddenValue` to `v8::Object::GetPrivate`/`SetPrivate`/`DeletePrivate` (and added `HasPrivate`) and changed the return types from `Local` to `MaybeLocal`.  Seems like this is a good candidate for wrapping with `nan`, especially since the very latest versions of v8 seem to have deleted the old `SetHiddenValue` versions, which will break a lot of my modules.  (This change hasn't quite made it into node yet AFAICT.)

kkoopa:
Since that change will break ABI compatibility, Node will not use it until v7, which is scheduled for some time in autumn. This will be taken care of then.

maxkorp:
So unfortunately electron (as of 1.3) has already jumped to v8 5.2.x.y, in which this is broken, and also seems to have a different signature (key/value, instead of context/key/value).

maxkorp:
Obviously, this is not set up using the global exported version constants and is not out in it's own file or anything of the sort, but the gist of what we've ended up with in nodegit to fix this is as such

```cpp
#if (NODE_MODULE_VERSION > 48)
  v8::Local<v8::Value> GetPrivate(v8::Local<v8::Object> object,
                                      v8::Local<v8::String> key) {
    v8::Isolate* isolate = v8::Isolate::GetCurrent();
    v8::Local<v8::Context> context = isolate->GetCurrentContext();
    v8::Local<v8::Private> privateKey = v8::Private::ForApi(isolate, key);
    v8::Local<v8::Value> value;
    v8::Maybe<bool> result = object->HasPrivate(context, privateKey);
    if (!(result.IsJust() && result.FromJust()))
      return v8::Local<v8::Value>();
    if (object->GetPrivate(context, privateKey).ToLocal(&value))
      return value;
    return v8::Local<v8::Value>();
  }

  void SetPrivate(v8::Local<v8::Object> object,
                      v8::Local<v8::String> key,
                      v8::Local<v8::Value> value) {
    if (value.IsEmpty())
      return;
    v8::Isolate* isolate = v8::Isolate::GetCurrent();
    v8::Local<v8::Context> context = isolate->GetCurrentContext();
    v8::Local<v8::Private> privateKey = v8::Private::ForApi(isolate, key);
    object->SetPrivate(context, privateKey, value);
  }
#else
[...]

PR #599, opened 2016-08-18
Add getPrivate/setPrivate methods

maxkorp:
Addressing this: https://github.com/nodejs/nan/issues/587

Of note, I set this to use get/setPrivate on module version 49+, although this should work with older versions as well (I know it does not work in with modules <= 46, but is already marked deprecated in modules 48, I have not tested against 47). 

I've tried to maintain coding standards, although to be totally frank I'm completely unsure that i put this in the correct place in the code at all, and that I handled it properly (some functions seem to be redeclared entirely, while others have a single declaration with different contents).

Regarding testing, I've not added anything to the test suites, but I've been able to build and use nodegit using this change on node 0.12.15, 4.4.7, 6.2.2, 6.3.1, and electron 1.2.8, and 1.3.3, all succesfully (note that building nodegit for node 6.3 wont work on windows for separate reasons). I'd be happy to try to add something to the tests if this is a welcome change.
I've got a commit of nodegit working this way here: https://github.com/nodegit/nodegit/tree/nan-test

I assume some documentation would also need to be added for this.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -2191,7 +2191,7 @@ inline void SetCallAsFunctionHandler(
 
 //=== HiddenValue/Private ======================================================
 
-v8::Local<v8::Value> GetPrivate(
+inline v8::Local<v8::Value> GetPrivate(
     v8::Local<v8::Object> object,
     v8::Local<v8::String> key) {
 #if (NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION)
@@ -2210,7 +2210,7 @@ v8::Local<v8::Value> GetPrivate(
 #endif
 }
 
-void SetPrivate(
+inline void SetPrivate(
     v8::Local<v8::Object> object,
     v8::Local<v8::String> key,
     v8::Local<v8::Value> value) {
--- a/nan.h
+++ b/nan.h
@@ -2189,6 +2189,42 @@ inline void SetCallAsFunctionHandler(
 
 #include "nan_object_wrap.h"  // NOLINT(build/include)
 
+//=== HiddenValue/Private ======================================================
+
+v8::Local<v8::Value> GetPrivate(
+    v8::Local<v8::Object> object,
+    v8::Local<v8::String> key) {
+#if (NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION)
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::Private> privateKey = v8::Private::ForApi(isolate, key);
+  v8::Local<v8::Value> value;
+  v8::Maybe<bool> result = object->HasPrivate(context, privateKey);
+  if (!(result.IsJust() && result.FromJust()))
+    return v8::Local<v8::Value>();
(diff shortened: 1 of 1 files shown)


saper:
I think we can use `NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION` here, so that will be changed already for node 6.x which is running v8 v5.0 (it was deprecated in v8 v4.9). node 5.x was v8 v4.6-something.

nan.h
```diff
@@ -2189,6 +2189,42 @@ inline void SetCallAsFunctionHandler(
...
+v8::Local<v8::Value> GetPrivate(
+    v8::Local<v8::Object> object,
+    v8::Local<v8::String> key) {
+#if (NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION)
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::Private> privateKey = v8::Private::ForApi(isolate, key);
+  v8::Local<v8::Value> value;
+  v8::Maybe<bool> result = object->HasPrivate(context, privateKey);
+  if (!(result.IsJust() && result.FromJust()))
```
bnoordhuis: You can simplify this to `if (!result.FromMaybe(false))`.

nan.h
```diff
@@ -2189,6 +2189,42 @@ inline void SetCallAsFunctionHandler(
...
 
+//=== HiddenValue/Private ======================================================
+
+v8::Local<v8::Value> GetPrivate(
+    v8::Local<v8::Object> object,
+    v8::Local<v8::String> key) {
+#if (NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION)
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::Private> privateKey = v8::Private::ForApi(isolate, key);
```
bnoordhuis: Minor style nit but I think nan predominantly uses snake_case for locals, not camelCase.

maxkorp:
So, is this a desirable change for v6?

saper:
I believe this should be changed for node v6 and all the future versions.

maxkorp:
Now that Node 7 is out using 5.4.500.36 and this can be tested, what do I need to do to get this in motion?

JoshuaWise:
I also need this to add Node 7 support to one of my modules.

JoshuaWise:
Question:
Is it necessary to include `object->HasPrivate` in the implementation for `GetPrivate`? It seems like `object->GetPrivate` already returns a maybe.

kkoopa:
I'm lacking unit tests for the new functionality, that's about it.

agnat:
> I assume some documentation would also need to be added for this.

@maxkorp, could you look into that too?

JoshuaWise:
I crave this so bad.

JoshuaWise:
I believe this issue can be closed, because of https://github.com/nodejs/nan/pull/630

maxkorp:
Bah, I completely dropped the ball on this one. Sorry about that guys, and thanks Koopa for following through on the tests.

PR #602, opened 2016-08-30
Add optional data parameter to SetPrototypeMethod

spinda:
The `v8::FunctionTemplate` that `Nan::SetPrototypeMethod` creates can take some user data to thread through to the callback, but `SetPrototypeMethod` doesn't currently expose this parameter like some of the other NAN functions. I'm using a patched version of `SetPrototypeMethod` to work around this in my own project, but it's a simple change to upstream and would reduce a little of my future maintenance burden.


kkoopa:
This seems to be valid, but could you add a regression test as well?

spinda:
Closing since this was rolled into #876.

PR #603, opened 2016-09-03
Add example

untoldone:
Might not be the best example, but this took a lot longer than I had hoped to figure out for myself. Hope sharing it is helpful. If this isn't the best place to put this, would really appreciate an issue opened on the linked project with where it could be surfaced.


untoldone:
Given it's age, I assume this is no longer relevant

PR #611, opened 2016-09-26
Add New<[typedarray]> methods

### LINKED ISSUE #521: More TypedArray helpers?
Right now in node-canvas we've got stuff like [this](https://github.com/Automattic/node-canvas/blob/master/src/ImageData.cc#L62-L68):

```cpp
#if NODE_MAJOR_VERSION == 0 && NODE_MINOR_VERSION <= 10
    Local<Int32> sizeHandle = Nan::New(length);
    Local<Value> caargv[] = { sizeHandle };
    clampedArray = global->Get(Nan::New("Uint8ClampedArray").ToLocalChecked()).As<Function>()->NewInstance(1, caargv);
#else
    clampedArray = Uint8ClampedArray::New(ArrayBuffer::New(Isolate::GetCurrent(), length), 0, length);
#endif
```

and [this](https://github.com/Automattic/node-canvas/blob/master/src/ImageData.cc#L70-L75):

```cpp
#if NODE_MAJOR_VERSION == 0 && NODE_MINOR_VERSION <= 10
 if (info[0]->ToObject()->GetIndexedPropertiesExternalArrayDataType() == kExternalPixelArray) {
    clampedArray = info[0]->ToObject();
    length = clampedArray->GetIndexedPropertiesExternalArrayDataLength();
#else
 if (info[0]->IsUint8ClampedArray() && info[1]->IsUint32()) {
    clampedArray = info[0].As<Uint8ClampedArray>();
    length = clampedArray->Length();
#endif
```

Since `TypedArrayContents` was [added](https://github.com/nodejs/nan/pull/486) (thank you!), wondering if there's interest in adding more helpers for TypedArrays? Maybe something like:

* `Nan::New<TypedArrayT>(Nan::New<ArrayBuffer>(byteLength[, offset[, length]]))`
* `Nan::New<TypedArrayT>(length)`

Not so sure on argument testing. It would be nice if NAN_METHOD unified the interface so `info[x]->IsUint8Arary()` just worked. Otherwise is the only option `Nan::IsUint8ClampedArray`, `Nan::IsUint8Array`, ... and `Nan::TypedArrayLength(TypedArray)` ?

zbjornson:
This is not ready to merge yet; soliciting feedback first. (At the least, this needs the pre-0.12 implementation (unless NAN is also dropping 0.10 support?) and real tests for the ternary form.)

Adds `Nan::New<[typedarray]>(size_t length)` and `Nan::New<[typedarray]>(v8::Local<v8::ArrayBuffer>, size_t offset, size_t length)` for all of the typed array types.

This is a very repetitive set of functions to add. _I'm open to suggestions for better methods than macros for de-duping._ I struggled for a bit to try to templatize these but couldn't figure out how to specialize `Factory` via another level of templating. As-is, there's not really a reason to have `TypedArrayFactory` in addition to the macros, except that it would dedupe code between the 0.12+ and 0.10 impl (when it's added); if not, the macros could be changed to be `TYPED_ARRAY_DECL(v8::Uint8Array, uint8_t)` for example.

I'm also interested in exposing the C types for the typed array types, since v8 does not do this. That is, make it so you could do something like `NAN_TYPED_ARRAY_CTYPE(v8::Local<v8::Uint8Array>)` and get `uint8_t`. _Is there a better API to offer for this than a macro?_

Thoughts?

Fixes #521.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -185,6 +185,17 @@ template<typename T, typename M = NonCopyablePersistentTraits<T> >
 class Persistent;
 #endif  // NODE_MODULE_VERSION
 
+template <typename T> struct TypedArrayCType;
+template<> struct TypedArrayCType<v8::Float32Array> { typedef float type; };
+template<> struct TypedArrayCType<v8::Float64Array> { typedef double type; };
+template<> struct TypedArrayCType<v8::Int8Array> { typedef int8_t type; };
+template<> struct TypedArrayCType<v8::Uint8Array> { typedef uint8_t type; };
+template<> struct TypedArrayCType<v8::Uint8ClampedArray> { typedef uint8_t type; };
+template<> struct TypedArrayCType<v8::Int16Array> { typedef int16_t type; };
+template<> struct TypedArrayCType<v8::Uint16Array> { typedef uint16_t type; };
+template<> struct TypedArrayCType<v8::Int32Array> { typedef int32_t type; };
+template<> struct TypedArrayCType<v8::Uint32Array> { typedef uint32_t type; };
+
 #if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 4 ||                      \
   (V8_MAJOR_VERSION == 4 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 3))
 # include "nan_maybe_43_inl.h"  // NOLINT(build/include)
--- a/nan_implementation_12_inl.h
+++ b/nan_implementation_12_inl.h
@@ -26,6 +26,36 @@ Factory<v8::Array>::New(int length) {
   return v8::Array::New(v8::Isolate::GetCurrent(), length);
 }
 
+//=== TypedArrays ==============================================================
+
+#define TYPED_ARRAY_DECL(T)                                                   \
+  Factory<T>::return_t                                                        \
+  Factory<T>::New(size_t len) {                                               \
+    size_t byteLength = len * sizeof(TypedArrayFactory<T>::value_type);       \
+    v8::Local<v8::ArrayBuffer> buffer = v8::ArrayBuffer::New(                 \
+      v8::Isolate::GetCurrent(), byteLength);                                 \
+    return T::New(buffer, 0, len);                                            \
+  }                                                                           \
+                                                                              \
+  Factory<T>::return_t                                                        \
+  Factory<T>::New(v8::Local<v8::ArrayBuffer> buffer,                          \
+                  size_t offset,                                              \
+                  size_t len) {                                               \
+    return T::New(buffer, offset, len);                                       \
+  }
+
+TYPED_ARRAY_DECL(v8::Float32Array)
+TYPED_ARRAY_DECL(v8::Float64Array)
+TYPED_ARRAY_DECL(v8::Int8Array)
+TYPED_ARRAY_DECL(v8::Uint8Array)
+TYPED_ARRAY_DECL(v8::Uint8ClampedArray)
+TYPED_ARRAY_DECL(v8::Int16Array)
+TYPED_ARRAY_DECL(v8::Uint16Array)
+TYPED_ARRAY_DECL(v8::Int32Array)
+TYPED_ARRAY_DECL(v8::Uint32Array)
+
+#undef TYPED_ARRAY_DECL
+
 //=== Boolean ==================================================================
 
 Factory<v8::Boolean>::return_t
--- a/nan_implementation_12_inl.h
+++ b/nan_implementation_12_inl.h
@@ -31,7 +31,7 @@ Factory<v8::Array>::New(int length) {
 #define TYPED_ARRAY_DECL(T)                                                   \
   Factory<T>::return_t                                                        \
   Factory<T>::New(size_t len) {                                               \
-    size_t byteLength = len * sizeof(TypedArrayFactory<T>::value_type);       \
+    size_t byteLength = len * sizeof(TypedArrayCType<T>::type);       \
     v8::Local<v8::ArrayBuffer> buffer = v8::ArrayBuffer::New(                 \
       v8::Isolate::GetCurrent(), byteLength);                                 \
     return T::New(buffer, 0, len);                                            \
--- a/nan_implementation_12_inl.h
+++ b/nan_implementation_12_inl.h
@@ -28,33 +28,22 @@ Factory<v8::Array>::New(int length) {
 
 //=== TypedArrays ==============================================================
 
-#define TYPED_ARRAY_DECL(T)                                                   \
-  Factory<T>::return_t                                                        \
-  Factory<T>::New(size_t len) {                                               \
-    size_t byteLength = len * sizeof(TypedArrayCType<T>::type);       \
-    v8::Local<v8::ArrayBuffer> buffer = v8::ArrayBuffer::New(                 \
-      v8::Isolate::GetCurrent(), byteLength);                                 \
-    return T::New(buffer, 0, len);                                            \
-  }                                                                           \
-                                                                              \
-  Factory<T>::return_t                                                        \
-  Factory<T>::New(v8::Local<v8::ArrayBuffer> buffer,                          \
-                  size_t offset,                                              \
-                  size_t len) {                                               \
-    return T::New(buffer, offset, len);                                       \
-  }
+template <typename T>
+typename TypedArrayFactory<T>::return_t
+TypedArrayFactory<T>::New(size_t len) {
+  size_t byteLength = len * sizeof(TypedArrayCType<T>::type);
+  v8::Local<v8::ArrayBuffer> buffer = v8::ArrayBuffer::New(
+    v8::Isolate::GetCurrent(), byteLength);
+  return T::New(buffer, 0, len);
+}
 
-TYPED_ARRAY_DECL(v8::Float32Array)
-TYPED_ARRAY_DECL(v8::Float64Array)
-TYPED_ARRAY_DECL(v8::Int8Array)
-TYPED_ARRAY_DECL(v8::Uint8Array)
-TYPED_ARRAY_DECL(v8::Uint8ClampedArray)
-TYPED_ARRAY_DECL(v8::Int16Array)
-TYPED_ARRAY_DECL(v8::Uint16Array)
-TYPED_ARRAY_DECL(v8::Int32Array)
-TYPED_ARRAY_DECL(v8::Uint32Array)
-
-#undef TYPED_ARRAY_DECL
+template <typename T>
+typename TypedArrayFactory<T>::return_t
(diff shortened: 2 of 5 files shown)


agnat:
Hey @zbjornson. Thanks for looking into this.

> This is a very repetitive set of functions to add. I'm open to suggestions for better methods than macros for de-duping. I struggled for a bit to try to templatize these but couldn't figure out how to specialize Factory via another level of templating.

Hm, yeah. I see... that does not feel quite right. Let me think about it.

> I'm also interested in exposing the C types for the typed array types. [...] Is there a better API to offer for this than a macro?

Yes. It's called a type trait (or meta-function if you like) and you already (almost) implemented it. It is currently named `TypedArrayFactory`. Note that `TypedArrayFactory<v8::Uint8Array>::value_type` _is_ a typedef for `uint8_t`. ;)

agnat:
Ok, I _think_ I see how to do it. I'm only staring at the diff. So there might be dragons...

#### 1. Free type trait

AFAICT, this should compile (and run) after each step.

 * stop multi-inheriting from `TypedArrayFactory`
 * rename `TypedArrayFactory` to `TypedArrayCType` adjusting the only reference
 * rename inner typedef `value_type` to `type` adjusting the only reference 
 * hoist `TypedArrayCType` into namespace `Nan`

#### 2. Reimplement `TypedArrayFactory` template

This adds a new facility and keeps the old stuff intact for now. We add a new template class

````
template <typename T>
struct TypedArrayFactory : FactoryBase<T> {
  // declare static New(...)s here
} 
````

The implementation of `New(...)` looks like this... I think:

````
template <typename T>
TypedArrayFactory<T>::return_t
TypedArrayFactory<T>::New(...) {
  // paste from macro
}
````

This should again compile and run... after fixing the syntax errors. ;)

#### 3. Use it

First, comment out the invocations of `TYPED_ARRAY_DECL(...)` breaking things for the first time. Now add full specializations of `Factory` for each array type like so:

````
template <> struct Factory<v8::Uint8Array> : TypedArrayFactory<v8::Uint8Array> {};
// etc....
```` 

I think that's it.

So, compile, fix and finally, remove all macro stuff.

**Comment by zbjornson** [comment]
Yay, much better, thanks @agnat.

Now pending the pre-v0.12 impl and docs, will do early this week.

**Comment by zbjornson** [comment]
I'm a bit stuck with the pre-0.12 impl. Before going further, will NAN v3 drop support for 0.8/0.10? If so, perhaps this would be best left for v3 without a 0.10 impl, or maybe released as a non-NAN helper header.

The problem: Pre-v0.12, there was no `v8::Int8Array`, etc., instead these were plain `v8::Object`s, and the types were differentiated by the `v8::ExternalArrayType` enum (e.g. `kExternalByteArray`). Likewise, `v8::ArrayBuffer` was a plain `v8::Object`.

* Patching 0.10 with the 0.12+ syntax `New<v8::Int8Array>(...)` means augmenting `v8::` with those classes, but `v8::Object` cannot be extended because its constructor is private. Operator overloads on a wrapper class get part way to a solution, but it doesn't seem possible to create a `Handle` (`Local`) for the wrapper class. Something like (with errors):

    ```c++
    // namespace v8
    class TypedArray {
      Local<Object> typedarray_;
      public:
        static TypedArray* New(Isolate* isolate, size_t byte_length); // can't return Local<TypedArray>
        Local<Object>* operator->() { return &typedarray_; }
        operator Local<Object>() { return typedarray_; }
        size_t ByteLength() { ... } // etc, to match 0.12+
    }
    class Uint8Array : TypedArray { ... }

    // usage
    v8::Uint8Array* ta = v8::Uint8Array::New(iso, 24);
    // have to explicitly cast because v8 has a non-compiling overload for T*
    info.GetReturnValue().Set((Local<Object>)(ta));
    ```

    (I think the same problems apply if those classes were in `Nan::` instead of `v8::`.)

* Alternatively, `New` in both <0.12 and 0.12+ could return `v8::Objects`, but users would then have to check the node module version to see if it's safe to use the newer TypedArray/ArrayBufferView methods, or the older Object methods.

* Adding `typedef Object ArrayBuffer` to the 0.10 impl is simple and allows `Local<v8::ArrayBuffer> buf = Nan::New<v8::ArrayBuffer>(isolate, byteLen)`, but has the same problem as above with methods.

Suggestions?

**Comment by kkoopa** [comment]
The plan is to drop support for 0.8 and 0.10. There is a new stable node.js module API under development and NAN 3 might well materialize on top of that in one form or another. However, this will take a while.

I do not really have any good suggestions at hand. Part of the reason I never looked further into this is because there are so many complications.

The `v8::Handle(T*)` constructor should be available before 0.12, if I am not mistaken.

The method problem might be solvable if there is only a handful of methods: make NAN wrappers for said methods that take e.g. a `Local<Nan::ArrayBuffer>`, where `Nan::ArrayBuffer` is `v8::Object` or `v8::ArrayBuffer` depending on version.

**Comment by agnat** [comment]
TBH, I'm a little irritated here. This looks like a real show-stopper. Why does this pop up now?

In #521 @kkoopa [commented](https://github.com/nodejs/nan/issues/521#issuecomment-158629456) on the approach:

>  I'd suggest planning the design before starting on the implementation, as that is the easiest way of getting something consistent and working.

To me it looks like you asked others to re-design your sketch without even looking into feasibility. I spent time and effort on this, which now seems wasted.

> [...] or maybe released as a non-NAN helper header

Right.

:-/

**Comment by zbjornson** [comment]
@agnat I had prototyped this but missed some aspects. The separate file idea was a last resort.

In any case, either (a) implementing Nan wrapper classes for these or (b) following kkoopa's suggestion and wrapping the methods both seem to work. Option (a) sort of follows the approach taken with the napi PoC.

```cpp
namespace Nan {
  class ArrayBuffer {
  public:
    static ArrayBuffer New(/*Isolate* isolate, */size_t byte_length) {
      v8::Local<v8::Object> global = v8::Context::GetCurrent()->Global();
      // TODO assert byte_length
      v8::Local<v8::Value> argv[] = { v8::Uint32::NewFromUnsigned((uint32_t)byte_length) };
      v8::Local<v8::Function> fn = global->Get(v8::String::New("ArrayBuffer")).As<v8::Function>();
      v8::Local<v8::Object> obj = fn->NewInstance(1, argv);
      ArrayBuffer ab;
      ab.obj_ = obj;
      return ab;
    }
    operator v8::Local<v8::Object>() const { return obj_; }
    inline v8::Local<v8::Object> operator -> () const { return obj_; }
    inline v8::Local<v8::Object> operator * () const { return obj_; }
    size_t ByteLength() const { return obj_->GetIndexedPropertiesExternalArrayDataLength(); }
    void* Data() const { return obj_->GetIndexedPropertiesExternalArrayData(); }
  private:
    v8::Local<v8::Object> obj_;
  };
}

// class ReturnValue
  inline void Set(ArrayBuffer &ab) {
    value_->Dispose();
    *value_ = v8::Persistent<T>::New((v8::Local<v8::Object>)ab);
  } // allows info.GetReturnValue().Set(Nan::ArrayBuffer)
```

(Dead option (c): I was wrong about not be able to extend `v8::Object` and was able to implement `Local<v8::ArrayBuffer> v8::ArrayBuffer::New()`, but casting the `v8::ArrayBuffer` into the Local handle did not appear legal/safe, [ex](https://gist.github.com/zbjornson/7310218caf6eff697220792fbdc6c384).)

agnat:
Ah, alright then. Never mind. Just ping me for the next round of review...

zbjornson:
Hello again, @agnat et al. -- The below seems to work to shim <0.12 with classes to match 0.12+. Is this an acceptable approach?

```cpp
// Usage same as 0.12+:
// v8::Local<v8::ArrayBuffer> ab = v8::ArrayBuffer::New(isolate, bytelen);

namespace v8 {
  class ArrayBuffer : public Object {
    // (accessory class, returned by GetContents)
    class Contents {
    public:
      void* Data() const { return data_; }
      size_t ByteLength() const { return byte_length_; }
    private:
      void* data_;
      size_t byte_length_;
      friend class Local<ArrayBuffer>;
    };
    // Wrapped value
    Local<Object> val_;
    ArrayBuffer();
    friend class Local<ArrayBuffer>;
  public:
    inline static Local<ArrayBuffer> New(v8::Isolate*, size_t);
  };

  template<> class Local<ArrayBuffer> {
  public:
    inline Local(ArrayBuffer* val) : arrbuf_(val) {}
    inline operator Local<Object>() const { return arrbuf_->val_; }
    // Need conversion defined all the way up to Handle<Value>:
    inline operator Handle<Value>() const { return arrbuf_->val_; }
    // Delegate through to the v8::Object wrapped within ArrayBuffer
    inline Object* operator->() const { return arrbuf_->val_.operator->(); }
    inline Object* operator*() const { return arrbuf_->val_.operator*(); }
    // Because of operator overloading, API methods need to be on this specialization.
    size_t ByteLength() const { return arrbuf_->val_->GetIndexedPropertiesExternalArrayDataLength(); }
    ArrayBuffer::Contents GetContents() const {
      ArrayBuffer::Contents contents;
      contents.data_ = arrbuf_->GetIndexedPropertiesExternalArrayData();
      contents.byte_length_ = (size_t)arrbuf_->GetIndexedPropertiesExternalArrayDataLength();
      return contents;
    }
  private:
    ArrayBuffer* arrbuf_;
  };

  Local<ArrayBuffer> ArrayBuffer::New(v8::Isolate* isolate, size_t byte_length) {
    // isolate isn't used in pre-12, but present to be consistent with 12+
    v8::Local<v8::Object> global = v8::Context::GetCurrent()->Global();
    v8::Local<v8::Value> argv[] = { v8::Uint32::NewFromUnsigned((uint32_t)byte_length) };
    v8::Local<v8::String> ctorName = v8::String::New("ArrayBuffer");
    v8::Local<v8::Function> fn = global->Get(ctorName).As<v8::Function>();
    v8::Local<v8::Object> obj = fn->NewInstance(1, argv);
    ArrayBuffer* ab;
    ab->val_ = obj;
    Local<ArrayBuffer> local_ab(ab);
    return local_ab;
  }
}
```

(Same approach applies to the TypedArray classes.)

Have three backups if that's not a good approach, but they APIs they would provide are not as nice. The above is invisible to the majority of users (i.e. 0.12+ users) and would thus be most easily forward-compatible.

Thanks!

agnat:
A couple of things that jump the eye:

````
    // Wrapped value
    Local<Object> val_;
````

I think this won't work. Having a `Local<>` data member does not really make sense.

I'm also very skeptical regarding the full specialization of `Local<ArrayBuffer>`. It has the same _syntax_ but it does not provide the same _semantics_. It does not actually root the `ArrayBuffer`. In general specializing a _foreign_ template is not a good idea.

We usually avoid polluting the v8 namespace. We prefer to re-export v8 components in namespace Nan. This kind of is an artifact of the full specialization, though.

Last, please add code as actual commits including at least one invocation. This helps with the review.

PR #623, opened 2016-11-03
Test on Node 7

addaleax:
edit: does this even make sense here? I basically wanted to duplicate the commit for Node 6, but since @nodejs doesn’t have Travis enabled, I’m not so sure anymore…

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
   - TRAVIS_NODE_VERSION="4"
   - TRAVIS_NODE_VERSION="5"
   - TRAVIS_NODE_VERSION="6"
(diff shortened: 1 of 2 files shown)


brodybits:
This repo does have Travis and Appveyor enabled with status included in the README. The test on Node 7 already passes on Linux and Windows.

addaleax:
Huh. Right. Weird, I thought that wasn’t possible for some reason (but maybe that is only about new repositories). Thanks for pointing it out to silly me!

kkoopa:
Looks good. It's the github hooks that are not allowed, which is why the CI status does not show in pull requests or elsewhere you might expect to see it.

PR #631, opened 2016-12-08
Linted code

kkoopa:
The linter was complaining about some implicit conversions.


kkoopa:
I ended up calling it "Delint code.", since i ran the lint finder and removed the lint.
e8280fe268350502fc36ee9dbaf305581271b7d1

PR #633, opened 2016-12-12
abort in delete operators that shouldn't be called

jBarz:
Section 3.2 of the C++ standard states that destructor definitions
implicitly "use" operator delete functions. Therefore, these operator
delete functions must be defined even if they are never called by
user code explicitly.
http://www.open-std.org/JTC1/SC22/WG21/docs/cwg_defects.html#261

gcc allows them to remain as empty definitions. However, not all
compilers allow this.

This pull request creates definitions which if ever called, result
in an abort.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -351,7 +351,9 @@ class HandleScope {
   HandleScope(const HandleScope &);
   void operator=(const HandleScope &);
   void *operator new(size_t size);
-  void operator delete(void *, size_t);
+  void operator delete(void *, size_t) {
+    abort();
+  }
 };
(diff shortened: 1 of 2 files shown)


jBarz:
@kkoopa 
Do you know when the next release for nan is planned?
We have ported nodeJS to a new platform (zOS). Our npm users require this fix :-)
If I can provide them a date for the next nan release, that would be great!

kkoopa:
Waiting for a review of the private accessors PR. Once it is merged I'll publish a new version.

PR #634, opened 2016-12-24
removed "HandleScope scope" from private accessors

JoshuaWise:
These `HandleScope scope;` declarations were causing crashes on some node versions, and failed tests on other node versions. Removing them fixed all my problems.

### DIFF
--- a/nan_private.h
+++ b/nan_private.h
@@ -11,7 +11,6 @@
 
 inline Maybe<bool>
 HasPrivate(v8::Local<v8::Object> object, v8::Local<v8::String> key) {
-  HandleScope scope;
 #if NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION
   v8::Isolate *isolate = v8::Isolate::GetCurrent();
   v8::Local<v8::Context> context = isolate->GetCurrentContext();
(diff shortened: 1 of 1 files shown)


kkoopa:
But they are perfectly correct. Something else is afoot. Applying this patch would be wrong.

kkoopa:
Possibly the GetPrivate one should be an escapable handle scope.

JoshuaWise:
> Possibly the GetPrivate one should be an escapable handle scope.

How would we use an EscapableHandleScope for a function that returns a MaybeLocal?

JoshuaWise:
Okay, I've updated the pull request with a new solution that also works for me. I'm guessing it's probably not correct to blindly call `.ToLocalChecked()`, but maybe you can provide insight on how to properly handle the MaybeLocales with `EscapableHandleScope#Escape`?

kkoopa:
I have been thinking about that, but have not yet come up with a definitive answer. The problem actually goes a bit deeper, since much of the MaybeLocal is affected by this. As it happens, that code does not use handle scopes (although it should).

The safest alternative would be to have a Persistent member in MaybeLocal, which would be reset in the destructor. Then the stored handle would be safe. However, this does come at a price, which is why I would like to find a better alternative.

Next week I will dig into the V8 implementation of MaybeLocal and see how it is handled there. If that does not yield anything, it will either have to be the Persistents, or not using scopes.

JoshuaWise:
Okay, thank you for looking into this.

kkoopa:
Thank you for reporting this.

PR #635, opened 2016-12-26
REPLACE_INVALID_UTF8 requires node 0.10.29

Piu130:
NodeJS introduced `REPLACE_INVALID_UTF8` in 0.10.29. See [here](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V010.md#0.10.29).

Usage [here](https://github.com/nodejs/nan/blob/master/nan.h#L324).

Would be nice to have this dependency because I tried to install some sensors that use this module on a raspberry pi with preinstalled node version from ubuntu repo (<0.10.29).

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,5 +30,8 @@
     "tap": "~0.7.1",
     "xtend": "~4.0.0"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "engines": {
+    "node": ">=0.10.29"
+  }
 }


kkoopa:
I do not understand why the minimum version should be set to that.
Firstly, it would prevent 0.8 from working. Secondly, this is not a required feature. If your version of Node 0.10 is less than 0.10.29, it will not do anything, since the feature is unavailable. If you are running an unofficial (which includes every Linux distribution that mess around with anything) Node build, all bets are off. For all I know the code may have been replaced with Tetris.

Piu130:
The problem is `npm install node-dht-sensor` throws an error because `REPLACE_INVALID_UTF8` is used in your module. This module wont work on 0.8 either because there was no `REPLACE_INVALID_UTF8` in this version.

kkoopa:
I wrote the code in question. Node 0.8 definitely supports REPLACE_INVALID_UTF8. Based on the provided information, I am guessing that the fault lies in the unofficial node package bundled with your distribution.

PR #637, opened 2016-12-30
Fix #634

kkoopa:
This ought to be the proper fix for #634.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -637,23 +637,23 @@ class TryCatch {
   }
 
 # define X(NAME)                                                               \
-    inline v8::Local<v8::Value> NAME(const char *msg) {                    \
+    inline v8::Local<v8::Value> NAME(const char *msg) {                        \
       EscapableHandleScope scope;                                              \
       return scope.Escape(v8::Exception::NAME(New(msg).ToLocalChecked()));     \
     }                                                                          \
                                                                                \
-    inline                                                                 \
+    inline                                                                     \
     v8::Local<v8::Value> NAME(v8::Local<v8::String> msg) {                     \
       return v8::Exception::NAME(msg);                                         \
     }                                                                          \
                                                                                \
-    inline void Throw ## NAME(const char *msg) {                           \
+    inline void Throw ## NAME(const char *msg) {                               \
       HandleScope scope;                                                       \
       v8::Isolate::GetCurrent()->ThrowException(                               \
           v8::Exception::NAME(New(msg).ToLocalChecked()));                     \
     }                                                                          \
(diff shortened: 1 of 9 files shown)


nan_private.h
```diff
@@ -24,16 +24,21 @@ HasPrivate(v8::Local<v8::Object> object, v8::Local<v8::String> key) {
...
 GetPrivate(v8::Local<v8::Object> object, v8::Local<v8::String> key) {
-  HandleScope scope;
 #if NODE_MODULE_VERSION >= NODE_6_0_MODULE_VERSION
   v8::Isolate *isolate = v8::Isolate::GetCurrent();
+  v8::EscapableHandleScope scope(isolate);
   v8::Local<v8::Context> context = isolate->GetCurrentContext();
   v8::Local<v8::Private> private_key = v8::Private::ForApi(isolate, key);
-  return object->GetPrivate(context, private_key);
+  v8::MaybeLocal<v8::Value> v = object->GetPrivate(context, private_key);
+  assert(!v.IsEmpty() && "v should not be empty");
```
kkoopa: This line is arguably unnecessary.

JoshuaWise:
bump

grinich:
another +1 to land this

stemcc:
This would be an extremely helpful fix.

nan.h
```diff
@@ -1122,51 +1136,53 @@ widenString(std::vector<uint16_t> *ws, const uint8_t *s, int l) {
   NewOneByteString(const uint8_t * value, int length = -1) {
     std::vector<uint16_t> wideString;  // NOLINT(build/include_what_you_use)
     imp::widenString(&wideString, value, length);
-    return imp::Factory<v8::String>::return_t(v8::String::New(
-        &wideString.front(), static_cast<int>(wideString.size())));
+    return v8::String::New(&wideString.front(),
```
bnoordhuis: Using .front() when the string is empty is UB.  You could use .data() instead, that has well-defined behavior for the .empty() case (pointer to a `\0` byte as of C++11, something I learned earlier this week.)

bnoordhuis:
Sorry for being so late to reply.  I was on holiday until last week and I had 1,000+ unread emails when I came back.  I will admit to not having read them all.

nan.h
kkoopa: Alright, will do.

kkoopa:
No problem. Thanks for the review. I'll get this sorted out now.

kkoopa:
Done.

JoshuaWise:
Thanks!

Issue #650, opened 2017-03-14
add a helper for the v8::JSON object and its methods Parse & Stringify


mkrufky:
As the V8 API for the `JSON` object has changed over the course of the versions of V8 used in node, it makes good sense to add a helper for it, especially now that the `Stringify` method has been made accessible in node 7.

`v8::JSON` has two methods, `Parse` and `Stringify` ...  For a link to the v8 docs used in node 7, see [v8::JSON Class Reference](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html)

As for the changes in `v8::JSON` in the versions of node supported by _NAN_:

`v8::JSON::Parse` was first available in node version 0.12.x - it is *not* otherwise available in versions 0.8 or 0.10.x
`v8::JSON::Stringify` was first available in node version 7 - it is *not* otherwise available in earlier versions

mkrufky:
I submitted PR #651 to add this functionality.

Anybody that is interested in using this functionality before it is reviewed / merged into _NAN_ can use the [native-json](https://www.npmjs.com/package/native-json) npm package.   Also maintained on github: [node-native-json](https://github.com/mkrufky/node-native-json)

I just think something like this makes more sense as a part of _NAN_ rather than a separate package.  Comments are welcome.

PR #651, opened 2017-03-14
nan: add support for JSON::Parse & Stringify

mkrufky:
As the V8 API for the `JSON` object has changed over the versions of V8 used in node, it makes good sense to add a helper for it, especially now that the `Stringify` method has been made available in node 7.

`v8::JSON` has two methods, `Parse` and `Stringify` ...  For a link to the v8 docs used in node 7, see [v8::JSON Class Reference](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html)

As for the changes in `v8::JSON` in the versions of node supported by _NAN_:

`v8::JSON::Parse` was first available in node version 0.12.x - it is *not* otherwise available in versions 0.8 or 0.10.x
`v8::JSON::Stringify` was first available in node version 7 - it is *not* otherwise available in earlier versions

closes https://github.com/nodejs/nan/issues/650

### DIFF
--- a/doc/json.md
+++ b/doc/json.md
@@ -47,7 +47,7 @@ Example:
 
 ```c++
 // using `v8::Local<v8::Value> val` from the `JSON::Parse` example
-v8::Local<v8::Object> obj = val->ToObject();
+v8::Local<v8::Object> obj = Nan::To<v8::Object>(val);
 
 v8::Local<v8::String> stringified = Nan::JSON::Stringify(obj).ToLocalChecked();
 ```
--- a/doc/json.md
+++ b/doc/json.md
@@ -16,7 +16,7 @@ A simple wrapper around [`v8::JSON::Parse`](https://v8docs.nodesource.com/node-7
 Definition:
 
 ```c++
-Nan::MaybeLocal<v8::Value> Nan::JSON::Parse(v8::Local<v8::String> jsonString);
+Nan::MaybeLocal<v8::Value> Nan::JSON::Parse(v8::Local<v8::String> json_string);
 ```
 
 Use `JSON.Parse(json_string)` to parse a `v8::String` into a `v8::Value`.
@@ -39,7 +39,7 @@ A simple wrapper around [`v8::JSON::Stringify`](https://v8docs.nodesource.com/no
 Definition:
 
 ```c++
-Nan::MaybeLocal<v8::String> Nan::JSON::Stringify(v8::Local<v8::Object> jsonObject, v8::Local<v8::String> gap = v8::Local<v8::String>());
+Nan::MaybeLocal<v8::String> Nan::JSON::Stringify(v8::Local<v8::Object> json_object, v8::Local<v8::String> gap = v8::Local<v8::String>());
 ```
 
 Use `JSON.Stringify(value)` to convert a `v8::Object` into a `v8::String`.
--- a/doc/json.md
+++ b/doc/json.md
@@ -1,0 +1,54 @@
+## JSON
+
+The _JSON_ object provides the c++ versions of the methods offered by the `JSON` object in javascript. V8 exposes these methods via the `v8::JSON` object.
+
+ - <a href="#api_nan_json_parse"><b><code>Nan::JSON::Parse</code></b></a>
+ - <a href="#api_nan_json_stringify"><b><code>Nan::JSON::Stringify</code></b></a>
+
+Refer to the V8 JSON object in the [V8 documentation](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html) for more information about these methods and their arguments.
+
+<a name="api_nan_json_parse"></a>
+
+### Nan::JSON::Parse
+
+A simple wrapper around [`v8::JSON::Parse`](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html#a936310d2540fb630ed37d3ee3ffe4504).
+
+Definition:
+
+```c++
+Nan::MaybeLocal<v8::Value> Nan::JSON::Parse(v8::Local<v8::String> jsonString);
+```
+
+Use `JSON::Parse(json_string)` to parse a `v8::String` into a `v8::Value`.
+
+Example:
+
+```c++
+v8::Local<v8::String> json_string = Nan::New("{ \"JSON\": \"object\" }").ToLocalChecked();
+
+v8::Local<v8::Value> val = Nan::JSON::Parse(json_string).ToLocalChecked();
+```
+
+<a name="api_nan_json_stringify"></a>
+
+### Nan::JSON::Stringify
+
+A simple wrapper around [`v8::JSON::Stringify`](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html#a44b255c3531489ce43f6110209138860).
+
+Definition:
+
+```c++
+Nan::MaybeLocal<v8::String> Nan::JSON::Stringify(v8::Local<v8::Object> jsonObject, v8::Local<v8::String> gap = v8::Local<v8::String>());
+```
+
+Use `JSON::Stringify(value)` to convert a `v8::Object` into a `v8::String`.
+
+Example:
+
+```c++
+// using `v8::Local<v8::Value> val` from the `JSON::Parse` example
+v8::Local<v8::Object> obj = val->ToObject();
+
+v8::Local<v8::String> stringified = Nan::JSON::Stringify(obj).ToLocalChecked();
+```
+
--- a/doc/json.md
+++ b/doc/json.md
@@ -2,14 +2,14 @@
 
 The _JSON_ object provides the c++ versions of the methods offered by the `JSON` object in javascript. V8 exposes these methods via the `v8::JSON` object.
 
- - <a href="#api_nan_json_parse"><b><code>Nan::JSON::Parse</code></b></a>
- - <a href="#api_nan_json_stringify"><b><code>Nan::JSON::Stringify</code></b></a>
+ - <a href="#api_nan_json_parse"><b><code>Nan::JSON.Parse</code></b></a>
+ - <a href="#api_nan_json_stringify"><b><code>Nan::JSON.Stringify</code></b></a>
 
 Refer to the V8 JSON object in the [V8 documentation](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html) for more information about these methods and their arguments.
 
 <a name="api_nan_json_parse"></a>
 
-### Nan::JSON::Parse
+### Nan::JSON.Parse
 
 A simple wrapper around [`v8::JSON::Parse`](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html#a936310d2540fb630ed37d3ee3ffe4504).
 
@@ -35,7 +35,7 @@ if (!result.IsEmpty()) {
 
 <a name="api_nan_json_stringify"></a>
 
-### Nan::JSON::Stringify
+### Nan::JSON.Stringify
 
 A simple wrapper around [`v8::JSON::Stringify`](https://v8docs.nodesource.com/node-7.4/da/d6f/classv8_1_1_j_s_o_n.html#a44b255c3531489ce43f6110209138860).
 
--- a/doc/json.md
+++ b/doc/json.md
@@ -27,7 +27,10 @@ Example:
 v8::Local<v8::String> json_string = Nan::New("{ \"JSON\": \"object\" }").ToLocalChecked();
 
 Nan::JSON NanJSON;
-v8::Local<v8::Value> val = NanJSON.Parse(json_string).ToLocalChecked();
+Nan::MaybeLocal<v8::Value> result = NanJSON.Parse(json_string);
+if (!result.IsEmpty()) {
+  v8::Local<v8::Value> val = result.ToLocalChecked();
+}
 ```
 
 <a name="api_nan_json_stringify"></a>
@@ -51,6 +54,9 @@ Example:
 v8::Local<v8::Object> obj = Nan::To<v8::Object>(val).ToLocalChecked();
 
 Nan::JSON NanJSON;
-v8::Local<v8::String> stringified = NanJSON.Stringify(obj).ToLocalChecked();
+Nan::MaybeLocal<v8::String> result = NanJSON.Stringify(obj);
+if (!result.IsEmpty()) {
+  v8::Local<v8::String> stringified = result.ToLocalChecked();
+}
 ```
 
--- a/doc/json.md
+++ b/doc/json.md
@@ -19,7 +19,7 @@ Definition:
 Nan::MaybeLocal<v8::Value> Nan::JSON::Parse(v8::Local<v8::String> json_string);
 ```
 
-Use `JSON.Parse(json_string)` to parse a `v8::String` into a `v8::Value`.
+Use `JSON.Parse(json_string)` to parse a string into a `v8::Value`.
 
 Example:
 
@@ -45,7 +45,7 @@ Definition:
 Nan::MaybeLocal<v8::String> Nan::JSON::Stringify(v8::Local<v8::Object> json_object, v8::Local<v8::String> gap = v8::Local<v8::String>());
 ```
 
-Use `JSON.Stringify(value)` to convert a `v8::Object` into a `v8::String`.
+Use `JSON.Stringify(value)` to stringify a `v8::Object`.
 
 Example:
 
--- a/doc/json.md
+++ b/doc/json.md
@@ -19,14 +19,15 @@ Definition:
 Nan::MaybeLocal<v8::Value> Nan::JSON::Parse(v8::Local<v8::String> jsonString);
 ```
 
-Use `JSON::Parse(json_string)` to parse a `v8::String` into a `v8::Value`.
+Use `JSON.Parse(json_string)` to parse a `v8::String` into a `v8::Value`.
 
 Example:
 
 ```c++
 v8::Local<v8::String> json_string = Nan::New("{ \"JSON\": \"object\" }").ToLocalChecked();
 
-v8::Local<v8::Value> val = Nan::JSON::Parse(json_string).ToLocalChecked();
+Nan::JSON NanJSON;
+v8::Local<v8::Value> val = NanJSON.Parse(json_string).ToLocalChecked();
 ```
 
 <a name="api_nan_json_stringify"></a>
@@ -41,7 +42,7 @@ Definition:
 Nan::MaybeLocal<v8::String> Nan::JSON::Stringify(v8::Local<v8::Object> jsonObject, v8::Local<v8::String> gap = v8::Local<v8::String>());
 ```
 
-Use `JSON::Stringify(value)` to convert a `v8::Object` into a `v8::String`.
+Use `JSON.Stringify(value)` to convert a `v8::Object` into a `v8::String`.
 
 Example:
 
@@ -49,6 +50,7 @@ Example:
 // using `v8::Local<v8::Value> val` from the `JSON::Parse` example
 v8::Local<v8::Object> obj = Nan::To<v8::Object>(val).ToLocalChecked();
 
-v8::Local<v8::String> stringified = Nan::JSON::Stringify(obj).ToLocalChecked();
+Nan::JSON NanJSON;
+v8::Local<v8::String> stringified = NanJSON.Stringify(obj).ToLocalChecked();
 ```
 
--- a/doc/json.md
+++ b/doc/json.md
@@ -47,7 +47,7 @@ Example:
 
 ```c++
 // using `v8::Local<v8::Value> val` from the `JSON::Parse` example
-v8::Local<v8::Object> obj = Nan::To<v8::Object>(val);
+v8::Local<v8::Object> obj = Nan::To<v8::Object>(val).ToLocalChecked();
 
 v8::Local<v8::String> stringified = Nan::JSON::Stringify(obj).ToLocalChecked();
 ```
--- a/nan_json.h
+++ b/nan_json.h
@@ -1,0 +1,135 @@
+/*********************************************************************
+ * NAN - Native Abstractions for Node.js
+ *
+ * Copyright (c) 2017 NAN contributors
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
+
+#ifndef NAN_JSON_H_
+#define NAN_JSON_H_
+
+#if ((NODE_MAJOR_VERSION == 0) && (NODE_MINOR_VERSION < 12))
+#define NAN_JSON_H_NEED_PARSE 1
+#else
+#define NAN_JSON_H_NEED_PARSE 0
+#endif
+
+#if (NODE_MAJOR_VERSION >= 7)
+#define NAN_JSON_H_NEED_STRINGIFY 0
+#else
+#define NAN_JSON_H_NEED_STRINGIFY 1
+#endif
+
+class JSON {
+ public:
+  static inline
+  Nan::MaybeLocal<v8::Value> Parse(v8::Local<v8::String> jsonString) {
+#if NAN_JSON_H_NEED_PARSE
+    return instance().parse(jsonString);
+#else
+#if (NODE_MAJOR_VERSION >= 7)
+    return v8::JSON::Parse(Nan::GetCurrentContext(), jsonString);
+#else
+    return v8::JSON::Parse(jsonString);
+#endif
+#endif
+  }
+
+  static inline
+  Nan::MaybeLocal<v8::String> Stringify(v8::Local<v8::Object> jsonObject) {
+#if NAN_JSON_H_NEED_STRINGIFY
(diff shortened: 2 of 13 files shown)


mkrufky:
Anybody that is interested in using this functionality before it is reviewed / merged into _NAN_ can use the [native-json](https://www.npmjs.com/package/native-json) npm package.   Also maintained on github: [node-native-json](https://github.com/mkrufky/node-native-json)

I just think something like this makes more sense as a part of _NAN_ rather than a separate package.  Comments are welcome.

mkrufky:
Here is a link to the passing tests on my own appveyor account: [mkrufky/nan](https://ci.appveyor.com/project/mkrufky/nan/build/1)

I haven't yet discovered a way to trigger Travis CI without tearing down and re-pushing the branch, which I'm avoiding since I've already opened the PR.

mkrufky:
Here is a link to the passing tests on my own Travis CI account: [mkrufky/nan](https://travis-ci.org/mkrufky/nan)

I haven't yet discovered a way to trigger Travis CI without tearing down and re-pushing the branch, which I'm avoiding since I've already opened the PR, so build 1 ran off my `master` branch, which is equal to this `json` branch in this pull request.

nan_json.h
```diff
@@ -0,0 +1,138 @@
...
+#if NAN_JSON_H_NEED_PARSE
+  Nan::Callback m_cb_parse;
+#endif
+#if NAN_JSON_H_NEED_STRINGIFY
+  Nan::Callback m_cb_stringify;
+#endif
+
+#if (NAN_JSON_H_NEED_PARSE + NAN_JSON_H_NEED_STRINGIFY)
+  static JSON& instance() {
+    static JSON i;
```
bnoordhuis: I hate to be a Debby Downer but I don't think this can work.

`i` has static scope, its destructor runs at program exit, when the VM has already been shut down.  That's going to have disastrous results when the `m_cb_parse` and `m_cb_stringify` destructors run.

mkrufky: Thanks for the review, @bnoordhuis .   The singleton aspect is part of what makes this solution so cool, in my opinion, but you might be right.   Another option would be to ditch the singleton, and have have the `parse` and `stringify` entry points each call into something like this:

```c++
  v8::Local<v8::Value> call(const char *method, v8::Local<v8::Value> arg)
  {
    v8::Local<v8::Object> json = Nan::New(m_persistent);

    v8::Local<v8::Value> thisMethod = json->Get(Nan::New(method).ToLocalChecked());

    if (thisMethod.IsEmpty() || !thisMethod->IsFunction()) {
      return Nan::Undefined();
    }

    v8::Local<v8::Function> methodFunction = v8::Local<v8::Function>::Cast(thisMethod);

    return methodFunction->Call(json, 1, &arg);
  }
```

... changing the entry points to look more like this:

```c++
  static inline v8::Local<v8::Value> Parse(v8::Local<v8::Value> jsonString)
  {
    return JSON::instance().call("parse", jsonString);
  }

  static inline v8::Local<v8::Value> Stringify(v8::Local<v8::Value> jsonObject)
  {
    return JSON::instance().call("stringify", jsonObject);
  }
```

What do you think?

mkrufky: *or* I could keep everything as-is and just remove the singleton bit, requiring the user to instantiate a `Nan::JSON` object before use.

mkrufky: Working on this now... will push a patch and ping again in a few

mkrufky: @bnoordhuis I got rid of the singleton and all persistence.  The current solution does *not* require the user to instantiate a `Nan::JSON` object before use, as all the methods are static.

mkrufky: @bnoordhuis another option is to require the user to instantiate the `Nan::JSON` class and allow it to persist the global JSON object and required methods.  I've already done the work if you decide that it is a better option.  

You can find that in my https://github.com/mkrufky/nan/tree/json-persistent branch.

The patch that converts the class is here: https://github.com/mkrufky/nan/commit/f949132405bdc988bdeb2f9e895465dfe257e130

It is more resource efficient than the solution in this PR, but it is slightly less elegant, as it requires the user to instantiate the `Nan::JSON` object.

Let me know what your preference is.  If you want me to pull that into this PR, I will.  Otherwise hopefully this is good enough as-is.

mkrufky: ...and yet another option is to just remove the singleton.   If that's what you prefer, then the original PR but without the singleton is here:  https://github.com/mkrufky/nan/tree/json-without-singleton  .   Just like the previous option, it requires the user to instantiate the `Nan::JSON` object before use, but any persistent storage will use `Nan::Callback` just like the original PR (rather than storing inside the single persistent object, as in the previous option)

kkoopa:
I have not looked through the code in detail, as Ben is already on it, but my overall impression is that this is very nice. Thank you for a detailed pull request with documentation, tests and nicely formatted code. I'll make sure to get this published as soon as we have ironed out the details.

kkoopa:
I believe you should add `HandleScopes` in the other entry points too, not only for the `ifdef` parts. For example, since `GetCurrentContext` returns a `Local` handle, I am under the impression that it will leak to the surrounding scope.

kkoopa:
Also, don't directly use `v8::Value::ToString` or any other such conversions, they have been marked as soon to be deprecated for two years now. Use `Nan::To<T>` for type conversion. A similar argument applies to `v8::Object::Get` and related methods, use `Nan::Get` and such, check the NAN docs for maybe helpers to get a list of what should use them.

mkrufky:
Thank you for your review, @kkoopa .

I added `HandleScope`s to all entry points now, except for the pre-node 7 case for `Parse()` which doesn't take a context.

I converted all uses of `v8::Value::ToString` and `v8::Value::ToObject` to use `Nan::To<T>` and I also converted all uses of `v8::Object::Get` to `Nan::Get`

...in this pull request and the two other branches that I've created that depict the alternative approaches that I mentioned my comments earlier.   I have a feeling we'll end up with the code in the https://github.com/mkrufky/nan/tree/json-without-singleton branch, but I'm waiting to hear back from Ben on his decision before I pull those changes into this PR.

doc/json.md
```diff
@@ -47,7 +47,7 @@ Example:
 
 ```c++
 // using `v8::Local<v8::Value> val` from the `JSON::Parse` example
-v8::Local<v8::Object> obj = val->ToObject();
+v8::Local<v8::Object> obj = Nan::To<v8::Object>(val);
```
kkoopa: `Nan::To<v8::Object>` returns a `MaybeLocal<v8::Object>`.

nan_json.h
```diff
@@ -66,18 +67,20 @@ class JSON {
...
 
-    v8::Local<v8::Object> json = globalJSON->ToObject();
+    v8::Local<v8::Object> json =
+      Nan::To<v8::Object>(globalJSON).ToLocalChecked();
 
     v8::Local<v8::Value> thisMethod =
-      json->Get(Nan::New(method).ToLocalChecked());
+      Nan::Get(json, Nan::New(method).ToLocalChecked()).ToLocalChecked();
 
     if (thisMethod.IsEmpty() || !thisMethod->IsFunction()) {
```
kkoopa: `thisMethod` cannot be empty, since it came from a call to `ToLocalChecked`, so part of this check is unnecessary.
However, you should not be calling `ToLocalChecked` here in the first place, `ToLocalChecked` on an empty `MaybeLocal` produces a hard crash. You should check if the `MaybeLocal` from `Nan::Get` is empty before proceeding.

The `globalJSON`getting above has similar issues.

doc/json.md
mkrufky: fixed.  its easy to miss such a mistake in the documentation ;-)  thanks

nan_json.h
mkrufky: fixed... I updated all three branches to always check `MaybeLocal.IsEmpty()` before proceeding to `ToLocalChecked()`     ...  is this the right way to do it?  Unfortunately, I haven't been able to find much in the _NAN_ documentation about how to check a `MaybeLocal` but I found `IsEmpty()` in the _V8_ documentation.

mkrufky:
A question:  Before issuing this pull request, I didn't have any `HandleScope` at the class entry points and things were working nicely against all versions of node.

Now that `HandleScope`s have been added, I am seeing some strange results in some versions of node but not others.  I'm pretty sure I incorrectly used `Nan::HandleScope` where I should have actually used `Nan::EscapableHandleScope` , since the methods need to return a V8 type that has been created within.

However, `Nan::EscapableHandleScope.Escape()` returns a v8::Local<T> and the methods that we are trying to wrap return `v8::MaybeLocal<T>`.

I really think that both `Parse()` and `Stringify()` should continue to return `MaybeLocal` , so the idea of checking the `MaybeLocal` and unwrapping it with `ToLocalChecked()` only to stuff it into `Nan::EscapableHandleScope.Escape()` seems wrong.

What should I do?

bnoordhuis:
>  I have a feeling we'll end up with the code in the https://github.com/mkrufky/nan/tree/json-without-singleton branch

I'd say that's the best approach.  There are just too many things that can go wrong with singletons.

As well, since nan is a header-only project, a singleton gets duplicated in every source file that includes it.

> I really think that both Parse() and Stringify() should continue to return MaybeLocal , so the idea of checking the MaybeLocal returned from the underlying V8 API and unwrapping it with ToLocalChecked() only to stuff it into Nan::EscapableHandleScope.Escape() seems wrong.

I didn't test but I think this should work:
```c++
MaybeLocal<v8::Value> F() {
  EscapableHandleScope handle_scope;
  v8::Local<v8::Value> result;
  // ...
  if (result.IsEmpty()) return MaybeLocal<v8::Value>();
  return MaybeLocal<v8::Value>(handle_scope.Escape(result));
}
```

kkoopa:
Implicit conversion from `Local` to `MaybeLocal` applies, this is sufficient:

```c++
MaybeLocal<v8::Value> F() {
  EscapableHandleScope handle_scope;
  v8::Local<v8::Value> result;
  // ...
  if (result.IsEmpty()) return v8::Local<v8::Value>();
  return handle_scope.Escape(result);
}
```

bnoordhuis:
Oh, that's true.  I actually think that's a misfeature (I've seen more than one nodejs/node pull request where people think they're calling the MaybeLocal version when they're not) but yes, implicit conversion works at the moment.

mkrufky:
@bnoordhuis Thanks - I will try your suggestion with the `EscapableHandleScope` ...

As per the singleton, just to clarify:  I have already removed the singleton from the branch in this current PR.

The branch in this PR allows the user to call `Nan::JSON::Parse()` and `Nan::JSON::Stringify()` directly:

```c++
Nan::JSON::Parse("{ \"some\": \"thing\" }");
```

The one thing I don't love about this branch is that if the version of v8 doesn't expose the api, then the class has to grab the method from the javascript global JSON object on *every single call*.

The other branch ( https://github.com/mkrufky/nan/tree/json-without-singleton ) saves the methods into a `Nan::Callback` so that the class doesn't have to fetch the methods on every call, but the drawback is that the methods can't be static anymore.  Instead, the user must first instantiate the JSON object, then call the method.  For example:

```c++
Nan::JSON NanJSON;
NanJSON.Parse("{ \"some\": \"thing\" }");
```

The branch names are a misnomer now, I'm sorry for that.


Which solution would you prefer to merge?   ...or should I add both classes, one with static methods for one-offs, and the other with `Nan::Callback`s for re-use?

bnoordhuis:
Your second approach, the RAII class, is what I had in mind.

mkrufky:
OK... I will pull those changes into this branch so I don't have to constantly update multiple branches with every fix :-)

mkrufky:
I think this is all ready to go now.  The `EscapableHandleScope` indeed solved my problem, now the behavior is consistent across all versions of node once again :-)

Are there any other changes that need to be made?

If you decide to merge this, please revert 00a98ec254af10906b95d2e21f2e91d3e8d9a898 before doing so.  I needed to use my own Travis CI account to make sure I didn't break any node version that I cant test locally, and it was sending notifications to rvagg ...     I can't revert the change in this branch myself, because it will end up notifying him again :-/   I think things would be much better if the _NAN_ project had Travis build enabled on pull requests...

kkoopa:
Travis builds are enabled, just not the github hook for showing it. They cannot be enabled. If you check Travis you will see the PR build listed.

mkrufky:
Awesome.  Now I can see all of my PR builds :-)  I went ahead and reverted the notification patch from rvagg and disabled _nan_ on my own Travis account.

Let me know if anything else needs changing.

test/cpp/json-parse.cpp
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) 2017 NAN contributors
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
+
+#include <nan.h>
+
+NAN_METHOD(Parse) {
+  info.GetReturnValue().Set(
+    (new Nan::JSON)->Parse(
```
bnoordhuis: This leaks memory unnecessarily.

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+ public:
+  JSON() {
+#if (NAN_JSON_H_NEED_PARSE + NAN_JSON_H_NEED_STRINGIFY)
+    Nan::MaybeLocal<v8::Value> maybeGlobalJSON = Nan::Get(
+      Nan::GetCurrentContext()->Global(),
+      Nan::New("JSON").ToLocalChecked()
+    );
+
+    if (!maybeGlobalJSON.IsEmpty()) {
+      v8::Local<v8::Value> valGlobalJSON = maybeGlobalJSON.ToLocalChecked();
```
mkrufky: fixed.

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+          Nan::MaybeLocal<v8::Value> maybeParseMethod = Nan::Get(
+            globalJSON, Nan::New("parse").ToLocalChecked()
+          );
+
+          if (!maybeParseMethod.IsEmpty()) {
+            v8::Local<v8::Value> parseMethod =
+              maybeParseMethod.ToLocalChecked();
+
+            if (parseMethod->IsFunction()) {
+              m_cb_parse.Reset(v8::Local<v8::Function>::Cast(parseMethod));
```
mkrufky: fixed.

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+            }
+          }
+#endif
```
mkrufky: fixed.

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+    }
+#endif
+  }
+
+  ~JSON() {
+#if NAN_JSON_H_NEED_PARSE
+    m_cb_parse.Reset();
+#endif
+#if NAN_JSON_H_NEED_STRINGIFY
+    m_cb_stringify.Reset();
```
mkrufky: fixed.

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+    return scope.Escape(result.ToLocalChecked());
+  }
+
+ private:
+  NAN_DISALLOW_ASSIGN_COPY_MOVE(JSON)
+#if NAN_JSON_H_NEED_PARSE
+  Nan::Callback m_cb_parse;
+#endif
+#if NAN_JSON_H_NEED_STRINGIFY
+  Nan::Callback m_cb_stringify;
```
mkrufky: fixed.

test/cpp/json-parse.cpp
mkrufky: fixed.

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * Copyright (c) 2017 NAN contributors
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
+
+#ifndef NAN_JSON_H_
+#define NAN_JSON_H_
+
+#if ((NODE_MAJOR_VERSION == 0) && (NODE_MINOR_VERSION < 12))
```
mkrufky: fixed.

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+#define NAN_JSON_H_NEED_PARSE 1
+#else
+#define NAN_JSON_H_NEED_PARSE 0
+#endif
+
+#if (NODE_MAJOR_VERSION >= 7)
```
mkrufky: fixed.

mkrufky:
I believe that I've addressed all of the comments so far. I just cleaned up the documentation a bit, and fixed the `MaybeLocal` handling in the test cpp code.  Ready for the next round of review...

nan_json.h
```diff
@@ -0,0 +1,171 @@
...
+    if (result.IsEmpty()) return v8::Local<v8::Value>();
+    return scope.Escape(result.ToLocalChecked());
+#else
+    return scope.Escape(v8::JSON::Parse(jsonString));
+#endif
+#endif
+  }
+
+  inline
+  Nan::MaybeLocal<v8::String> Stringify(v8::Local<v8::Object> jsonObject) {
```
mkrufky: fixed this one last week too, just forgot to reply.

mkrufky:
Something is wrong with the Travis CI configuration for OSX.  Commit f0dcc863dda4ecc137935cffa89b23c05c4c5916 should have failed on node versions 0.8, 0.10 and 0.12, but it passed on OSX.  Maybe an extra copy of more recent v8 headers are laying around somewhere on that container?

Anyway, I fixed the issue in my code.  I accidentally used a `v8::MaybeLocal` rather than a `Nan::MaybeLocal` today when I realized that there was one last unchecked `MaybeLocal` in my tests.

I think this is all ready to go now...

mkrufky:
@bnoordhuis ping?

bnoordhuis:
I don't have much time to review at the moment.  Maybe @kkoopa can take a look.

kkoopa:
next week sometime

nan_json.h
```diff
@@ -0,0 +1,160 @@
...
+
+#if NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION
+#define NAN_JSON_H_NEED_STRINGIFY 0
+#else
+#define NAN_JSON_H_NEED_STRINGIFY 1
+#endif  // NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION
+
+class JSON {
+ public:
+  JSON() {
```
kkoopa: I think there should be a `HandleScope` somewhere here, possibly one in each conditional compilation part.

nan_json.h
```diff
@@ -0,0 +1,160 @@
...
+#if NAN_JSON_H_NEED_PARSE + NAN_JSON_H_NEED_STRINGIFY
+    Nan::MaybeLocal<v8::Value> maybe_global_json = Nan::Get(
+      Nan::GetCurrentContext()->Global(),
+      Nan::New("JSON").ToLocalChecked()
+    );
+
+    if (!maybe_global_json.IsEmpty()) {
```
kkoopa: When can this be empty? Is it when a user has overwritten the `JSON` object from `JavaScript`? Should an assert or error be thrown? Is it recoverable? What about the other sanity checking in this function, should it throw an error if the premise is not met?

nan_json.h
```diff
@@ -0,0 +1,160 @@
...
+  Nan::MaybeLocal<v8::String> Stringify(v8::Local<v8::Object> json_object) {
+    Nan::EscapableHandleScope scope;
+    Nan::MaybeLocal<v8::String> result =
+#if NAN_JSON_H_NEED_STRINGIFY
+      Nan::To<v8::String>(stringify(json_object));
+#else
+      v8::JSON::Stringify(Nan::GetCurrentContext(), json_object);
+#endif  // NAN_JSON_H_NEED_STRINGIFY
+    if (result.IsEmpty()) return v8::Local<v8::String>();
+    return scope.Escape(result.ToLocalChecked());
```
kkoopa: Here, and in some other places, you could have done `return scope.Escape(result.ToLocal(v8::Local<v8::String>()));`, but leaving it as is is fine by me.

test/cpp/json-parse.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <nan.h>
+
+NAN_METHOD(Parse) {
+  Nan::JSON NanJSON;
+
+  Nan::MaybeLocal<v8::String> inp = Nan::To<v8::String>(info[0]);
+
+  if (inp.IsEmpty()) {
+    info.GetReturnValue().Set(Nan::Undefined());
```
kkoopa: The default return value is `Undefined`, so there is no need to set it. Also, prefer the "fast primitive setters" where available, e.g. `ReturnValue::SetUndefined`. This applies in other places as well. Since the test suite has become a de facto example set and documentation, it is important to show best practices here.

nan_json.h
mkrufky: At first, I didn't think it would ever be possible for this to be empty. I included the check because it was a `MaybeLocal` and thus needed to be checked, although now that you mention it, yes, a user can indeed overwrite the `JSON` object from `JavaScript`.  Part of me thinks that if the user has done this, that they deserve whatever they get as a result 3:-)  But perhaps it would be a good idea to throw an error or add an assertion.  I'm just not sure.  Without changing the code, if this does end up being empty somehow, the entry point method (`Parse` / `Stringify`) will simply return `Undefined()`.  To me, this seems the most compatible path.

nan_json.h
mkrufky: You're right... I missed this one.  I added a Nan::Handlescope in the `#if NAN_JSON_H_NEED_PARSE + NAN_JSON_H_NEED_STRINGIFY` case.  I don't think more than one is needed in the constructor.

nan_json.h
mkrufky: This was inspired by your earlier comment, https://github.com/nodejs/nan/pull/651#issuecomment-287055725 ...  I'll be happy to change it in any way you prefer.  I don't have any preference on this one, so I'll leave it unless told otherwise.

test/cpp/json-parse.cpp
mkrufky: I replaced all `ReturnValue::Set(Nan::Undefined())` with `ReturnValue::SetUndefined()` ... You mentioned that the return value is `Undefined` by default, so there is actually no need to set any of these.   Since it was in a test suite I thought it best to be explicit.   Should I remove them or leave them as is, now that they've all been converted to `ReturnValue::SetUndefined()` ?

nan_json.h
mkrufky: What would happen to `V8`'s version of `v8::JSON::Parse` or `v8::JSON::Stringify` if the user overwrote the global `JSON` object?   probably nothing - it would probably still work.  I thought it would be safer to return Undefined, but maybe we should indeed error.....   I'll add some asserts.

mkrufky: done.  Each sanity check in the constructor has been converted to an `assert` ... it just makes me wonder now if I should just remove these later checks in the method entry points:  https://github.com/nodejs/nan/pull/651/files#diff-c15ba21614ac1fc925f08856752f6968R133

mkrufky: Done.  I converted all remaining sanity checks to `assert`s

test/cpp/json-parse.cpp
kkoopa: Better to remove them, as they are not needed. We don't explicitly set the return value to `Undefined` anywhere else, so it could be confusing to suddenly see it here.

nan_json.h
kkoopa: Yes, this looks good now.

nan_json.h
kkoopa: Nah, just leave it. I only mentioned it as an aside.

test/cpp/json-parse.cpp
mkrufky: OK, I went ahead and removed the unneeded `info.GetReturnValue().SetUndefined();` since `Undefined` is the default return value anyway.

nan_json.h
mkrufky: So now, if the user deliberately screws with the global `JSON` object *and* compiles with `NDEBUG` defined, bad things will happen.    It makes me wonder if I should have just ADDED the `assert`s rather than replacing the checks with `assert`s, but that would be redundant.

What would you like to see done, here?  Should I leave these `assert`s as they are?  Should I bring back the checks that I previously had, but after the `assert`s ?   I don't think we should throw any exceptions, but maybe that is a better path?

mkrufky: So now, if the user deliberately screws with the global `JSON` object *and* compiles with `NDEBUG` defined, bad things will happen.    It makes me wonder if I should have just ADDED the `assert`s rather than replacing the checks with `assert`s, but that would be redundant.

What would you like to see done, here @kkoopa ?  Should I leave these `assert`s as they are?  Should I bring back the checks that I previously had, but only after the `assert`s ?   I don't think we should throw exceptions, but maybe that is a better path?

kkoopa:
I do not have a clear idea of what is best, but I would say that a conservative approach is better, then we can loosen the restrictions at some future point, if feasible. Since I do not see an easy way of recovery from someone messing with the JSON object, an assert or error is probably good for now. The main point is that these errors should not be silent. We depend on a usable JSON object.

nan_json.h
mkrufky: (responding to comment here https://github.com/nodejs/nan/pull/651#issuecomment-290717929 )  OK, In that case, then I've addressed all of your comments, @kkoopa and I believe that this is all ready to go! :-)  Thanks for the review!

kkoopa:
Great. I will take a final look in the weekend, merge and publish mid-next week.

PR #662, opened 2017-04-06
add node version 6.9 to the CI environment test matrices

### LINKED ISSUE #661: node version 6.9.x should be explicitly tested in CI
Apparently the current CI tests for node version `6` pull down node version `6.10.x` , but we've come to the realization last night / this morning that there is an ABI breakage between the `V8` headers used between those versions of node.

Let's try to avoid this in the future by also testing node version `6.9` explicitly on both Travis and Appveyor

mkrufky:
Apparently the current CI tests for node version `6` pull down node version `6.10.x` , but we've come to the realization last night / this morning that there is an ABI breakage between the `V8` headers used between those versions of node.

Let's try to avoid this in the future by also testing node version `6.9` explicitly on both Travis and Appveyor

closes #661

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
   - TRAVIS_NODE_VERSION="4"
   - TRAVIS_NODE_VERSION="5"
   - TRAVIS_NODE_VERSION="6"
+  - TRAVIS_NODE_VERSION="6.9"
   - TRAVIS_NODE_VERSION="7"
 notifications:
   email:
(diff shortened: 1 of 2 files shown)


kkoopa:
Seems reasonable. However, this should not have occurred in the first place. The ABI is not allowed to break in a minor update of node.js.

mkrufky:
As per the discussion in #661 , now I also see that arbitrarily testing for `node 6.9.x` doesn't win us anything. This breakage was entirely due to electron being non-compliant. Let's not merge this.

PR #673, opened 2017-06-01
Update asyncprogressworkerstream.cpp

AndyleeSharp:
size is the number of objects, rather than the number of bytes


kkoopa:
Yes, it seems you are correct, which hints at possible bugs in the implementation or a badly named parameter. I will to take a closer look.

kkoopa:
Apart from the bug you found in the test, the actual implementation only suffers from a badly named parameter. I will rename `size` to `count` in the implementation and update the documentation when merging this commit.

AndyleeSharp:
yes, _count_ is better

kkoopa:
Merged as 9ef7d4dc69e23f7f8fc52f40eced5323db201c05

PR #680, opened 2017-07-10
Add Nan::To<v8::Function>() overload.

bnoordhuis:
Converting from v8::Value to v8::Function is a laborious and somewhat
unobvious process.  Add a convenience function to make it easier.


kkoopa:
This looks good, thank you both. I'll merge and push out a new release this week.

PR #683, opened 2017-07-30
Fix an incomplete code sample in the document.

hokein:


kkoopa:
Thank you for spotting and fixing it.

Issue #676, opened 2017-06-07
Drop support for v0.10 and v0.12?


bnoordhuis:

kkoopa:
My only real concern is that doing so would entail a new major release. Major releases are interpreted as breaking changes. Even if we do nothing else and plainly state that there are no actual changes, this still suggests that the new version would have a hard time gaining traction.

So, since major releases are expected to actually be major, I would want to take the opportunity to redesign everything again with lowest support for Node 4.0, incorporating Isolates, Context-passing and a bunch of other stuff. The first problem here is that I do not really have time to do this in a reasonable time. The second is that a lot of the API surface could be reduced, since V8 versions from Node 4 to 8 have been fairly stable. However, that could likely change at some point in the fairly near future, which would require adding a bunch of stuff back in. The third is that it would be fairly pointless, as the napi C++ layer already covers most of that. Due to these problems, a major redesign is not a feasible option.

So, are you suggesting to (choose any number of options):
1. remove any mention of 0.8, 0.10, 0.12 (and io.js) in the documentation
2. remove the code related to 0.8, 0.10, 0.12 (and io.js)
3. change other code
4. something else

As a (choose one option):
1. make a major release
2. make a minor release
3. make a patch release

bnoordhuis:
In that light I'd go for 1, remove any mention of 0.8, 0.10, 0.12 (and io.js) in the documentation.

I wouldn't actually remove all references, just state somewhere where it's clearly visible that interoperability with node.js < v4.x cannot be ensured due to those release lines being out of support (and maybe a note that it's known not to work with newer compilers.  Why be coy?)

I don't think we need to bump major for that because it's outside our control.  It's not until we start ripping out the support code (option 2) that we need to bump major.

kkoopa:
Okay, that sounds reasonable. Then the procedure would be to:
1. Update the documentation in the repository.
2. Include it in the next release, whenever that is (I do not think a documentation change warrants an instant patch release).

I think it is best to still run the tests for all versions until the code under test no longer exists.

bnoordhuis:
@kgryte If you feel strongly enough to dowvote, you might also want to elucidate why.

mkrufky:
Is there any plan to add some new functionality that can't be handled in node 0.10 and 0.12?  xcode might not compile NAN against 0.12.x, but NAN builds just fine on those embedded linux sdks and slightly older linux distros out there that have old packaged versions of node.

[...]

PR #686, opened 2017-08-18
Make garbage collection on the next tick to please Node 8.4

kkoopa:
I do not know why this is necessary, nor if it is the correct fix or simply addresses a symptom. My first guess would be something related to TurboFan. Please take a look.

CI:
https://travis-ci.org/nodejs/nan/builds/265943908
https://ci.appveyor.com/project/RodVagg/nan/build/1348

### DIFF
--- a/test/js/weak-test.js
+++ b/test/js/weak-test.js
@@ -21,11 +21,16 @@ test('weak', function (t) {
     count++;
   });
 
-  // run weak callback, should dispose
-  gc();
+  var timeout = setTimeout(function () {
+    // run weak callback, should dispose
+    gc();
 
-  // do not run weak callback
-  gc();
+    // do not run weak callback
+    gc();
 
-  t.equal(count, 1);
+    if (count > 0) {
+      clearTimeout(timeout);
+      t.equal(count, 1);
+    }
+  }, 100);
 });


bnoordhuis:
Does the test pass when you start node with `--single_threaded`?

kkoopa:
No, the existing test does not pass with `--single_threaded` set.

kkoopa:
Seems 8.3 had the same problem, but 8.2.1 is fine.

bnoordhuis:
I had a look but that test does something unsafe: it calls into JS land from a weak callback.

I'd say it's not a matter of figuring out why it doesn't work with v8.4.0 but why it worked at all in older versions. :-)

test/js/weak-test.js
```diff
@@ -21,11 +21,16 @@ test('weak', function (t) {
     count++;
   });
 
-  // run weak callback, should dispose
-  gc();
+  var timeout = setTimeout(function () {
```
bnoordhuis: I think just wrapping it in a `setImmediate()` should be enough, no need to fuss around with clearing timers again.

kkoopa: Probably. I copied it from the weak2-test, which tests the internal field variety of weak callbacks. I'll try to change that to `setImmediate` as well.

kkoopa:
Calling into js ought to be fine at that point, no? All callbacks are set as two-pass callbacks and the registered callback runs in the second pass. It should not be doing anything with the handle which was freed either.

bnoordhuis:
I don't think V8 ever made the guarantee that it's safe to call into JS from the GC post-processing step.  I know we used to do it at one time in node.js and we ran into issues after a while.

kkoopa:
Alright. Thanks for looking into this. I am content with having the tests working again.

kkoopa:
Okay, it was `setTimeout` because 0.8 does not have `setImmediate`. Should we just stop testing on 0.8 and release a new minor?

bnoordhuis:
Oh, good point.  I'm way past caring about v0.8 but if we're officially going to drop support for it, it should be semver-major.  If you simply want to stop testing for it, that could be minor or patch.

kkoopa:
We have #676 as a related issue, but it is still seems to be hanging in the air. I think I'll just change this back to `setTimeout` for now and deal with #676 at some later point.

bnoordhuis:
Sounds good.

PR #688, opened 2017-09-01
support z/OS ebcdic string types

jBarz:
This introduces a `NativeString` type which basically is the same as `Utf8String` on all platforms except z/OS where it is `EbcdicString`. 

Previously,
``` 
v8::Handle<v8::String> s;
Nan::Utf8String str(s);
system_library_call(*s);
```

This wouldn't work because the z/OS system doesn't use Utf8 string encoding but Ebcdic encoding.

But now, this will work for all platforms.
```
v8::Handle<v8::String> s
Nan::NativeString str(s);
system_library_call(*s);
```

This is a preliminary attempt to solve this problem but I am open to any other ways of achieving this.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -62,6 +62,9 @@
 #else
 # include <string>
 # include <vector>
+# ifdef __MVS__
+#  include <unistd.h>
+# endif
 #endif
 
 // uv helpers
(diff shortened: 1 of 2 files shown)


nan_implementation_12_inl.h
```diff
@@ -278,8 +278,18 @@ Factory<v8::String>::New(const char * value, int length) {
 Factory<v8::String>::return_t
 Factory<v8::String>::New(std::string const& value) {
   assert(value.size() <= INT_MAX && "string too long");
+#ifdef __MVS__
+  std::string str(value);
+  std::transform(str.begin(), str.end(), str.begin(), [](char c) -> char {
```
kkoopa: Lambdas are not allowed here, only C++98 features.

nan.h
```diff
@@ -940,11 +943,31 @@ class Utf8String {
...
   NAN_DISALLOW_ASSIGN_COPY_MOVE(Utf8String)
 
+ protected:
   int length_;
   char *str_;
   char str_st_[1024];
 };
 
+# ifdef __MVS__
+class EbcdicString : public Utf8String {
```
kkoopa: It seems wrong to have this extend `Utf8String`. It should be a sibling based on `WriteOneByteString`.

nan.h
```diff
@@ -940,11 +943,31 @@ class Utf8String {
...
+ public:
+  inline explicit EbcdicString(v8::Local<v8::Value> from) :
+      Utf8String(from) {
+    /* Convert buffer now */
+    if (length_ > 0)
+      __a2e_s(str_);
```
kkoopa: UTF-8 is a superset of ASCII.

jBarz:
thanks @kkoopa . I modified the source based on your suggestions.

nan.h
```diff
@@ -945,6 +948,56 @@ class Utf8String {
...
+      v8::Local<v8::String> string = from->ToString();
+      if (!string.IsEmpty()) {
+        size_t len = string->Length() + 1;
+        assert(len <= INT_MAX);
+        if (len > sizeof (str_st_)) {
+          str_ = static_cast<char*>(malloc(len));
+          assert(str_ != 0);
+        }
+        const int flags = v8::String::NO_NULL_TERMINATION;
+        length_ = string->WriteOneByte(reinterpret_cast<unsigned char*>(str_),
```
kkoopa: `static_cast`

jBarz: I will change the underlying storage type to a an unsigned char* (because ebcdic is based on unsigned char)
But I will then have to to move the reinterpret_cast to the cast operators.

 inline char* operator*() { return reinterpret_cast<char*>(str_); }

kkoopa:
Ah, yes. It does require a reinterpret_cast somewhere. I do not know where it is strictly better to do the cast.

bnoordhuis:
I don't know if I agree nan is the best place to solve this.  Nan abstracts over V8 and Node.js differences, not platform differences.

jBarz:
We could probably have the NativeString type declared in Node and expose that via the Node API?

jBarz:
or better yet, via libuv that is supposed to abstract over platform differences.

bnoordhuis:
I've thought about it but it seems like a flawed abstraction: not all platforms use UTF-8, and UTF-8 <-> EBCDIC conversion is not lossless.

kkoopa:
What about allowing strings to have an optional "translation" hook? This would be more generic, and allow someone to do a translation from Klingon to Romulan, as well as between anything else.

jBarz:
I have modified it to not involve the Utf8String class at all.
Now, there is just one class `NativeString` that uses `String::WriteOneByte` from v8.

jBarz:
I will look at implementing this outside `nan`.
Thanks for looking into it anyway :-)

PR #689, opened 2017-09-01
Prevent downcasting in json-stringify test

kkoopa:


mkrufky:
The change looks good to me and should be merged, of course... I'm just wondering, was there a warning generated prior to this change?  In what environment under what conditions does that warning get generated?  (I want to be sure to not generate more warnings in my next PR :-P)

kkoopa:
Yes, MSVC usually generates these warnings. Maybe with 32 bit builds. They were in all the AppVeyor logs.

PR #692, opened 2017-09-05
add AsyncProgressQueueWorker to queue and deliver all events

mkrufky:
add `AsyncProgressQueueWorker`: wake the main thread like AsyncProgressWorkerBase, except all events are queued and delivered

There are often cases where we want to handle events initiated by threads other than node threads.  Developers are often told to look at `AsyncProgressWorker` , which allows us to send progress updates from external threads, but this is not necessarily enough.  `AsyncProgressWorker` does what it needs to do, but if it is invoked many times before it wakes the main thread, only the last event will be delivered.  This is great for sending "progress" style events, but sometimes we need to make sure that all events get delivered.  `AsyncProgressWorker` doesn't do the job in those cases.

This PR refactors `AsyncProgressWorkerBase` to separate common elements and adds a new structure, `AsyncProgressQueueWorker` that behaves like `AsyncProgressWorkerBase` except all events are queued and delivered.

* The first patch, `AsyncProgressWorkerBase inherits AsyncBareProgressWorker` is a no-op change, that simply separates `AsyncBareProgressWorker` from `AsyncProgressWorkerBase`.

* The second patch adds the new functionality: `AsyncProgressQueueWorker`

* The third patch adds a test that shows how `AsyncProgressWorker` can (and will) miss some events.

* The fourth patch adds tests that show how `AsyncProgressQueueWorker` does not miss events.

* The fifth patch adds adds documentation for the new  `AsyncProgressQueueWorker` class.

* The sixth patch ensures that any possible remaining events in the queue are drained within `WorkComplete()`

The API of `AsyncProgressQueueWorker` is exactly the same as the API of `AsyncProgressWorkerBase` -- it's a drop-in replacement.  The only difference is that events will not be missed as a result of multiple wake requests being made before the `uv_async_t` wakes the main thread.  For more info, see the warning at the bottom of the [uv_async_t documentation](http://docs.libuv.org/en/v1.x/async.html):

> libuv will coalesce calls to uv_async_send(), that is, not every call to it will yield an execution of the callback. For example: if uv_async_send() is called 5 times in a row before the callback is called, the callback will only be called once. If uv_async_send() is called again after the callback was called, it will be called again.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1614,13 +1614,12 @@ template<class T>
  public:
   explicit AsyncBareProgressWorker(Callback *callback_)
       : AsyncWorker(callback_) {
-    async = new uv_async_t;
     uv_async_init(
         uv_default_loop()
-      , async
+      , &async
       , AsyncProgress_
     );
-    async->data = this;
+    async.data = this;
   }
 
   virtual ~AsyncBareProgressWorker() {
@@ -1632,7 +1631,7 @@ template<class T>
     friend class AsyncBareProgressWorker;
    public:
     void Signal() const {
-        uv_async_send(that_->async);
+        uv_async_send(&that_->async);
     }
 
     void Send(const T* data, size_t count) const {
@@ -1649,7 +1648,7 @@ template<class T>
   virtual void HandleProgressCallback(const T *data, size_t size) = 0;
 
   virtual void Destroy() {
-      uv_close(reinterpret_cast<uv_handle_t*>(async), AsyncClose_);
+      uv_close(reinterpret_cast<uv_handle_t*>(&async), AsyncClose_);
   }
 
  private:
@@ -1669,12 +1668,11 @@ template<class T>
   inline static void AsyncClose_(uv_handle_t* handle) {
     AsyncBareProgressWorker *worker =
             static_cast<AsyncBareProgressWorker*>(handle->data);
-    delete reinterpret_cast<uv_async_t*>(handle);
     delete worker;
   }
 
  protected:
-  uv_async_t *async;
+  uv_async_t async;
 };
 
 template<class T>
@@ -1721,7 +1719,7 @@ class AsyncProgressWorkerBase : public AsyncBareProgressWorker<T> {
     uv_mutex_unlock(&async_lock);
 
     delete[] old_data;
-    uv_async_send(this->async);
+    uv_async_send(&this->async);
   }
 
   uv_mutex_t async_lock;
@@ -1804,7 +1802,7 @@ class AsyncProgressQueueWorker : public AsyncBareProgressWorker<T> {
     asyncdata_.push(datapair);
     uv_mutex_unlock(&async_lock);
 
-    uv_async_send(this->async);
+    uv_async_send(&this->async);
   }
 
   uv_mutex_t async_lock;
--- a/nan.h
+++ b/nan.h
@@ -1744,13 +1744,12 @@ class AsyncProgressQueueWorker : public AsyncBareProgressWorker<T> {
     uv_mutex_lock(&async_lock);
 
     while (!asyncdata_.empty()) {
-      std::pair<T*, size_t> *datapair = asyncdata_.front();
-      T *data = datapair->first;
+      std::pair<T*, size_t> &datapair = asyncdata_.front();
+      T *data = datapair.first;
 
       asyncdata_.pop();
 
       delete[] data;
-      delete datapair;
     }
 
     uv_mutex_unlock(&async_lock);
@@ -1766,12 +1765,13 @@ class AsyncProgressQueueWorker : public AsyncBareProgressWorker<T> {
     uv_mutex_lock(&async_lock);
 
     while (!asyncdata_.empty()) {
-      std::pair<T*, size_t> *datapair = asyncdata_.front();
+      std::pair<T*, size_t> &datapair = asyncdata_.front();
+
+      T *data = datapair.first;
+      size_t size = datapair.second;
+
       asyncdata_.pop();
       uv_mutex_unlock(&async_lock);
-
-      T *data = datapair->first;
-      size_t size = datapair->second;
 
       // Don't send progress events after we've already completed.
       if (this->callback) {
@@ -1779,7 +1779,6 @@ class AsyncProgressQueueWorker : public AsyncBareProgressWorker<T> {
       }
 
       delete[] data;
-      delete datapair;
 
       uv_mutex_lock(&async_lock);
     }
@@ -1795,18 +1794,15 @@ class AsyncProgressQueueWorker : public AsyncBareProgressWorker<T> {
       std::copy(data, data + count, it);
     }
 
-    std::pair<T*, size_t> *datapair =
-        new std::pair<T*, size_t>(new_data, count);
-
     uv_mutex_lock(&async_lock);
-    asyncdata_.push(datapair);
+    asyncdata_.push(std::pair<T*, size_t>(new_data, count));
     uv_mutex_unlock(&async_lock);
 
     uv_async_send(&this->async);
   }
 
   uv_mutex_t async_lock;
-  std::queue<std::pair<T*, size_t>*> asyncdata_;
+  std::queue<std::pair<T*, size_t> > asyncdata_;
 };
 
 inline void AsyncExecute (uv_work_t* req) {
--- a/nan.h
+++ b/nan.h
@@ -53,13 +53,16 @@
 #include <cstring>
 #include <climits>
 #include <cstdlib>
+#include <utility>
 #if defined(_MSC_VER)
 # pragma warning( push )
 # pragma warning( disable : 4530 )
+# include <queue>
 # include <string>
 # include <vector>
 # pragma warning( pop )
 #else
+# include <queue>
 # include <string>
 # include <vector>
 #endif
@@ -1729,6 +1732,79 @@ class AsyncProgressWorkerBase : public AsyncBareProgressWorker<T> {
 // This ensures compatibility to the previous un-templated AsyncProgressWorker
 // class definition.
 typedef AsyncProgressWorkerBase<char> AsyncProgressWorker;
+
+template<class T>
+/* abstract */
+class AsyncProgressQueueWorker : public AsyncBareProgressWorker<T> {
+ public:
+  explicit AsyncProgressQueueWorker(Callback *callback_)
+      : AsyncBareProgressWorker<T>(callback_) {
+    uv_mutex_init(&async_lock);
+  }
+
+  virtual ~AsyncProgressQueueWorker() {
+    uv_mutex_lock(&async_lock);
+
+    while (!asyncdata_.empty()) {
+      std::pair<T*, size_t> *datapair = asyncdata_.front();
+      T *data = datapair->first;
+
+      asyncdata_.pop();
+
+      delete[] data;
+      delete datapair;
+    }
+
+    uv_mutex_unlock(&async_lock);
+    uv_mutex_destroy(&async_lock);
+  }
+
+  void WorkProgress() {
+    uv_mutex_lock(&async_lock);
+
+    while (!asyncdata_.empty()) {
+      std::pair<T*, size_t> *datapair = asyncdata_.front();
+      asyncdata_.pop();
+      uv_mutex_unlock(&async_lock);
+
+      T *data = datapair->first;
+      size_t size = datapair->second;
+
+      // Don't send progress events after we've already completed.
+      if (this->callback) {
+          this->HandleProgressCallback(data, size);
+      }
+
+      delete[] data;
+      delete datapair;
+
+      uv_mutex_lock(&async_lock);
+    }
+
+    uv_mutex_unlock(&async_lock);
+  }
+
+ private:
+  void SendProgress_(const T *data, size_t count) {
+    T *new_data = new T[count];
+    {
+      T *it = new_data;
+      std::copy(data, data + count, it);
+    }
+
+    std::pair<T*, size_t> *datapair =
+      new std::pair<T*, size_t>(new_data, count);
+
+    uv_mutex_lock(&async_lock);
+    asyncdata_.push(datapair);
+    uv_mutex_unlock(&async_lock);
+
+    uv_async_send(this->async);
+  }
+
+  uv_mutex_t async_lock;
+  std::queue<std::pair<T*, size_t>*> asyncdata_;
+};
 
 inline void AsyncExecute (uv_work_t* req) {
   AsyncWorker *worker = static_cast<AsyncWorker*>(req->data);
--- a/nan.h
+++ b/nan.h
@@ -1606,12 +1606,11 @@ class Callback {
   char *errmsg_;
 };
 
-
 template<class T>
-/* abstract */ class AsyncProgressWorkerBase : public AsyncWorker {
+/* abstract */ class AsyncBareProgressWorker : public AsyncWorker {
  public:
-  explicit AsyncProgressWorkerBase(Callback *callback_)
-      : AsyncWorker(callback_), asyncdata_(NULL), asyncsize_(0) {
+  explicit AsyncBareProgressWorker(Callback *callback_)
+      : AsyncWorker(callback_) {
     async = new uv_async_t;
     uv_async_init(
         uv_default_loop()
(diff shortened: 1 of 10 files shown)


kkoopa:
Overall it looks fine to me. This particular issue has been brought up on several occasions, so it would be good to have a correct example of an implementation with a queue.

mkrufky:
Thanks for the feedback!  Tests are passing on appveyor and on travis / linux, but I had to work around some different OSX behavior where the destructor of the test/cpp/asyncwakerequestor class got called before the main thread had a chance to call the first callback....  It's all good now (tested locally on my own mac) but I may have tied up Travis CI for the next few hours with older commits.

mkrufky:
After staring at these templates so much these past few days to make the tests pass, I've decided that `AsyncProgressQueueWorker` would be a much better name instead of `AsyncWakeRequestorBase`

I'm trying to separate the `AsyncWorker` from the `AsyncWakeRequestorBase`, if I manage to get that to work properly, the resulting class will have a `uv_async_t` but not have a `uv_work_t` and THAT class should be called `AsyncWakeRequestor` ...

I'll be away for the next few days so this will have to wait until I get back.

mkrufky:
ping @kkoopa ...  I know github's notifications aren't the best so here's a bump.

kkoopa:
I'll take a look tomorrow and see if I spot anything.

kkoopa:
As far as I can tell, this looks good. I would however like to ask @bnoordhuis to please take a look regarding the locking and thread safety.

nan.h
```diff
@@ -58,10 +58,14 @@
 # pragma warning( disable : 4530 )
 # include <string>
 # include <vector>
+# include <queue>
+# include <utility>
```
kkoopa: Do any of these cause warning 4530, or could they be moved to the unconditional include section?

mkrufky: I moved `<utility>` to the unconditional include section, but `<queue>` actually does generate warning 4530

kkoopa:
Alright, nevermind then

mkrufky:
I just added a patch that makes sure that `AsyncProgressQueueWorker::WorkProgress()` gets called before `AsyncWorker::WorkComplete()`, which otherwise could be risky without at least 1 ms sleep, so I also added a test for 0 ms sleep.

d403e6364aa76998f1a56e7f79f185b1a1828083

This might be useful for `AsyncProgressWorker` as well, but this PR aims to leave the behavior of `AsyncProgressWorker` untouched.  If we decide to "fix" `AsyncProgressWorker`, then we can just move `void WorkProgress()` to the `AsyncBareProgressWorker`

Making sure that `WorkProgress()` gets called before `WorkComplete()` is important to `AsyncProgressQueueWorker`, since it's aim is to deliver every event that is generated from `progress.Send()`

mkrufky:
@bnoordhuis , can you please take a look regarding the locking and thread safety?

bnoordhuis:
I'll add it to my TBR (To Be Reviewed) list but at +341 −43 it's a bit too big to review right now.

mkrufky:
The change isn't actually as big as the numbers imply -- it's mostly just breaking one class into smaller classes for re-use, then adding a new class, tests and documentation.

I re-spun the patch series and folded the fix-ups together so now this pull request is down to six small patches, each described in the OP, if that makes things any easier to review.

nan.h
```diff
@@ -1713,6 +1733,84 @@ template<class T>
...
+
+    uv_mutex_lock(&async_lock);
+    asyncdata_.push(datapair);
+    uv_mutex_unlock(&async_lock);
+
+    uv_async_send(this->async);
+  }
+
+  uv_mutex_t async_lock;
+  std::queue<std::pair<T*, size_t>*> asyncdata_;
```
bnoordhuis: This doesn't need to be a pointer to a `std::pair<...>`, does it?

test/cpp/asyncprogressqueueworker.cpp
```diff
@@ -0,0 +1,69 @@
...
+      progress.Send(reinterpret_cast<const char*>(&i), sizeof(int));
+      Sleep(milliseconds);
+    }
+  }
+
+  void HandleProgressCallback(const char *data, size_t count) {
+    HandleScope scope;
+
+    v8::Local<v8::Value> argv[] = {
+        New<v8::Integer>(*reinterpret_cast<int*>(const_cast<char*>(data)))
```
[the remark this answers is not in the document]
mkrufky: hmmm.  This file is an exact copy of `test/cpp/asyncprogressworker.cpp`  If this is a problem here, then it's a pre-existing problem in the original test as well.  Do we need to really change anything here?

nan.h
mkrufky: Is something wrong with a pointer to a std::pair<> ?  I wanted this new class template to be a drop-in replacement for `AsyncProgressWorkerBase`, which means it must pass both a `T*` and a `size_t` , I thought this was the natural way to do it.  Would you prefer a `std::pair<T*, size_t>` instead of a pointer?

nan.h
```diff
@@ -1713,6 +1733,84 @@ template<class T>
...
+ private:
+  void SendProgress_(const T *data, size_t count) {
+    T *new_data = new T[count];
+    {
+      T *it = new_data;
+      std::copy(data, data + count, it);
+    }
+
+    std::pair<T*, size_t> *datapair =
+      new std::pair<T*, size_t>(new_data, count);
```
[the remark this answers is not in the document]
mkrufky: done, in 3a209382bf6d09442e531a067851f6fd67535660

nan.h
```diff
@@ -1673,6 +1658,55 @@ template<class T>
...
+    AsyncBareProgressWorker *worker =
+            static_cast<AsyncBareProgressWorker*>(async->data);
+    worker->WorkProgress();
+  }
+
+  inline static void AsyncClose_(uv_handle_t* handle) {
+    AsyncBareProgressWorker *worker =
+            static_cast<AsyncBareProgressWorker*>(handle->data);
+    delete reinterpret_cast<uv_async_t*>(handle);
+    delete worker;
```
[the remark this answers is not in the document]
mkrufky: Great suggestion - thank you :-)   Done in b2cf2fee563cd37b9fcdd8a8081e85ff15e4294c

nan.h
bnoordhuis: Yes, that's what I mean, the extra level of indirection is not needed.  Sorry, I should have been more explicit.

test/cpp/asyncprogressqueueworker.cpp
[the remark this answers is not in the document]
bnoordhuis: Okay, let's leave it this way for now then.

nan.h
mkrufky: no problem. done in 02ccbd5c279df612bd552765fb5da1fe1da1d234

mkrufky: no problem.  done in 8fb5ba91d05e2631523d97365c7c8333ad79e25b

kkoopa:
Awesome! Thank you both for all your efforts.

maxbrunsfeld:
Since upgrading to nan 2.8.0, my [module](https://github.com/atom/superstring) which [uses `AsyncProgressWorkerBase`](https://github.com/atom/superstring/blob/8fd8d4ae3bec251007191849f4f8214ee09a2143/src/bindings/text-buffer-wrapper.cc#L677) has started to intermittently crash. I don't yet have a minimal repro for the bug, but I will try to produce one now.

kkoopa:
For starters, you seem to be missing `HandleScope`s in all `HandleOkCallbacks`.
The `Finish()` method returning a pair of `Local`s is not really proper, as it does not use an `EscapableHandleScope` (which would make returning a pair of `Local`s impossible), but it should be alright as it is currently used. Same goes for the `Finish` method returning a single `Local`.

The most likely reason for crashes is the missing `HandleScope`s in `HandleOKCallback`. Whenever execution goes from libuv to node, a `HandleScope` must be set up if any handles are created.

mkrufky:
The only change that comes to mind that affects `AsyncProgressWorkerBase` in this PR is this: b2cf2fee563cd37b9fcdd8a8081e85ff15e4294c .  Otherwise, the functionality of `AsyncProgressWorkerBase` remains exactly the same as it always was.  kkoopa's answer is likely the problem, but before he posted, I was scrambling for a fix for you to try, and I wonder how the following patch would affect things:
```patch
diff --git a/nan.h b/nan.h
index 7c7699f..f1e5338 100644
--- a/nan.h
+++ b/nan.h
@@ -1626,6 +1626,11 @@ class Callback {
   virtual ~AsyncBareProgressWorkerBase() {
   }
 
+  void WorkComplete() {
+    WorkProgress();
+    AsyncWorker::WorkComplete();
+  }
+
   virtual void WorkProgress() = 0;
 
   virtual void Destroy() {
@@ -1807,11 +1812,6 @@ class AsyncProgressQueueWorker : public AsyncBareProgressQueueWorker<T> {
     uv_mutex_destroy(&async_lock);
   }
 
-  void WorkComplete() {
-    WorkProgress();
-    AsyncWorker::WorkComplete();
-  }
-
   void WorkProgress() {
     uv_mutex_lock(&async_lock);
 
```
If that fixes things, don't rely on it -- I can't promise we'll merge it  (although it would be nice to hear if it does)  Regardless, please fix your missing `HandleScope` in `HandleOKCallback` and hopefully that will eliminate the problem altogether.

maxbrunsfeld:
I am guessing that my problem is related to the fact that I construct an instance of a `AsyncProgressWorkerBase` subclass on the stack, as a convenient way of implementing a synchronous variant of an async method. The crash seems to occur sometime after calling my `loadSync` method. Is there a reason why instances of `AsyncProgressWorkerBase ` should need to be heap-allocated?

maxbrunsfeld:
I noticed that the `async` field is no longer heap allocated in https://github.com/nodejs/nan/commit/b2cf2fee563cd37b9fcdd8a8081e85ff15e4294c. Prior to this commit, what code was responsible for *deleting* that object?

agnat:
The post-close delete in [1672](https://github.com/nodejs/nan/commit/b2cf2fee563cd37b9fcdd8a8081e85ff15e4294c#diff-558c37be49a47656e3ff41806c24bd33L1672) I believe...

kkoopa:
For the asynchronous case it almost surely needs to be allocated elsewhere than the stack, otherwise it would likely not survive until the worker thread is done. As for using it synchronously, nothing comes straight to mind, but I have never thought of doing that.

mkrufky:
If you don't queue the worker using `AsyncQueueWorker` then nobody ever calls `worker->Destroy()`

agnat:
Ok, I *think* I understand what happens...

Your stack allocated worker goes out of scope. This triggers an async close. The async close handler, that runs *later*, accesses something that is already gone... *boom*.

agnat:
Ok, I don't get it ... @mkrufky is more qualified to spot this anyway ...

maxbrunsfeld:
Ok, I'm just going to avoid using my `AsyncWorkerBase` subclass synchronously. I'll have to pull out its guts into a separate object. I didn't realize that the constructor of that class did non-trivial stuff. 

It might be nice if the `uv_async_init` happened in a separate initialization method that `AsyncQueueWorker` would call so that you could construct these objects on the stack if desired.

mkrufky:
I think, change ` TextBufferWrapper::load_sync()` as follows:
```c++
LoadWorker *worker = new LoadWorker(...)
```
and obviously, fix all the `worker` pointers, and then call
```cpp
  worker->Destroy();
```
at the end of the function.

maxbrunsfeld:
@mkrufky Ha that's a much quicker fix, great suggestion!

kkoopa:
And the `HandleScope`s I mentioned in `HandleOkCallback` should not be necessary after all, since `AsyncWorker::WorkComplete` makes a `HandleScope` before calling `HandleOKCallback` (unless `AsyncWorker::WorkComplete` has been overridden).

mkrufky:
`AsyncWorker::WorkComplete` never gets called in this case, and he's calling `HandleOkCallback` directly.  Better off to add the `HandleScope` because of the weird usage.

maxbrunsfeld:
Ah, thanks for the follow-up @kkoopa.

I actually don't call `HandleOkCallback` in the synchronous case. I call an underlying method called `Finish`.

maxbrunsfeld:
In general, is this an ok way of creating synchronous variants of async methods? In other words, is creating `AsyncProgressWorker` instances and calling `Destroy` on them manually without ever using `Nan::AsyncQueueWorker` a supported use case that I can rely on?

kkoopa:
Please, no. This is the first time I have ever seen this. The `AsyncWorker` classes are rather finicky (for a number of reasons, the main one being legacy). The standard way is to take the body of `AsyncWorker::Execute`, put it in a function `f`, call `f` from `AsyncWorker::Execute`. Export two functions `FooSync` and `FooAsync`, do the async setup in `FooAsync`, while `FooSync` simply calls `f`.

maxbrunsfeld:
Yeah, in this case there is a bunch of intermediate state that both `Execute` and `HandleOkCallback` need to read and write, but I see what you're saying. Thanks for all of the quick help and clarification.

maxbrunsfeld:
Maybe `Destroy` shouldn't be public.

kkoopa:
It is a virtual method serving part of the role of a destructor, which are public. I am not sure if it could be made protected, but that would constitute a breaking change so it will not happen in the foreseeable future. Again, legacy.

mkrufky:
I think we can safely do it without breaking anything, unless there's other code out there that calls these functions `worker->Destroy()`, `AsyncExecute()` & `AsyncExecuteComplete()` directly, and there shouldn't be, since I believe the only supported invocation is made via `AsyncQueueWorker`:
```patch
diff --git a/nan.h b/nan.h
index 7c7699f..9141df1 100644
--- a/nan.h
+++ b/nan.h
@@ -1502,6 +1502,7 @@ class Callback {
 };
 
 /* abstract */ class AsyncWorker {
+ friend class AsyncQueueWorker;
  public:
   explicit AsyncWorker(Callback *callback_)
       : callback(callback_), errmsg_(NULL) {
@@ -1571,14 +1572,14 @@ class Callback {
 
   uv_work_t request;
 
-  virtual void Destroy() {
-      delete this;
-  }
-
  protected:
   Persistent<v8::Object> persistentHandle;
   Callback *callback;
 
+  virtual void Destroy() {
+      delete this;
+  }
+
   virtual void HandleOKCallback() {
     HandleScope scope;
 
@@ -1628,11 +1629,11 @@ class Callback {
 
   virtual void WorkProgress() = 0;
 
+ private:
   virtual void Destroy() {
       uv_close(reinterpret_cast<uv_handle_t*>(&async), AsyncClose_);
   }
 
- private:
   inline static NAUV_WORK_CB(AsyncProgress_) {
     AsyncBareProgressWorkerBase *worker =
             static_cast<AsyncBareProgressWorkerBase*>(async->data);
@@ -1856,25 +1857,29 @@ class AsyncProgressQueueWorker : public AsyncBareProgressQueueWorker<T> {
   std::queue<std::pair<T*, size_t> > asyncdata_;
 };
 
-inline void AsyncExecute (uv_work_t* req) {
-  AsyncWorker *worker = static_cast<AsyncWorker*>(req->data);
-  worker->Execute();
-}
+class AsyncQueueWorker {
+ public:
+  explicit AsyncQueueWorker(AsyncWorker* worker) {
+    uv_queue_work(
+        uv_default_loop()
+      , &worker->request
+      , AsyncExecute
+      , reinterpret_cast<uv_after_work_cb>(AsyncExecuteComplete)
+    );
+  }
 
-inline void AsyncExecuteComplete (uv_work_t* req) {
-  AsyncWorker* worker = static_cast<AsyncWorker*>(req->data);
-  worker->WorkComplete();
-  worker->Destroy();
-}
+ private:
+  inline static void AsyncExecute (uv_work_t* req) {
+    AsyncWorker *worker = static_cast<AsyncWorker*>(req->data);
+    worker->Execute();
+  }
 
-inline void AsyncQueueWorker (AsyncWorker* worker) {
-  uv_queue_work(
-      uv_default_loop()
-    , &worker->request
-    , AsyncExecute
-    , reinterpret_cast<uv_after_work_cb>(AsyncExecuteComplete)
-  );
-}
+  inline static void AsyncExecuteComplete (uv_work_t* req) {
+    AsyncWorker* worker = static_cast<AsyncWorker*>(req->data);
+    worker->WorkComplete();
+    worker->Destroy();
+  }
+};
 
 namespace imp {
 
```

...passes all the tests

PR #695, opened 2017-09-20
Fix broken link

abetomo:
The link of `io.js-3.0` was 404, so I modified it to ~~`node-8.0`~~ `io.js-3.3`

### DIFF
--- a/doc/errors.md
+++ b/doc/errors.md
@@ -24,7 +24,7 @@ Also consult the V8 Embedders Guide section on [Exceptions](https://developers.g
 <a name="api_nan_error"></a>
 ### Nan::Error()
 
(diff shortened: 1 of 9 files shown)


kkoopa:
This will not work as the anchor links do not match. It seems io.js-3.3 docs are still up and there at least the majority of anchor tags seem to work.

abetomo:
Should I rewrite it to `io.js-3.3`?

kkoopa:
Unless you want to update all the link anchors to match those in the node 8 documentation, yes. This should involve less effort.

kkoopa:
Thank you for noticing and fixing this.

PR #697, opened 2017-09-24
Fix version of request to allow for installation on old versions

kkoopa:


kkoopa:
https://github.com/request/request/commit/479143d546d6dff378da0e5fac5801dd3bf01f15

PR #700, opened 2017-10-06
AsyncBareProgressWorker: allow reuse of  uv_async_t and AsyncWorker components

mkrufky:
`AsyncBareProgressWorker`: move `uv_async_t` and `AsyncWorker` components to non-template class `AsyncBareProgressWorkerBase`

Now that the larger change is merged, I'd like to break `AsyncBareProgressWorker` down into smaller classes so that the base can be reused in yet another worker.  Unfortunately, since this new worker uses a variadic template, it requires c++11.  Without an alternative for older compilers, I don't think this additional new worker actually belongs in _NAN_ until after we drop support for older node versions.

**update:  I was just reading over #676 and it looks like kkoopa and bnoordhuis already agreed that it's OK to use c++11 for new code.  So, after receiving some feedback, I can add the class below to this pull request if desired.

Meanwhile, if we merge this pull request, it opens the door to do awesome things like this:

```c++
#if NAN_HAS_CPLUSPLUS_11

template<class T, typename... Targs>
/* abstract */ class AsyncBareFactoryWorker : public AsyncBareProgressWorkerBase {
 public:
  explicit AsyncBareFactoryWorker(Callback *callback_)
      : AsyncBareProgressWorkerBase(callback_) {
  }

  virtual ~AsyncBareFactoryWorker() {
  }

  class ExecutionProgress {
    friend class AsyncBareFactoryWorker;
   public:
    void Construct(Targs... Fargs) const {
        that_->ConstructProgress_(Fargs...);
    }

   private:
    explicit ExecutionProgress(AsyncBareFactoryWorker *that) : that_(that) {}
    NAN_DISALLOW_ASSIGN_COPY_MOVE(ExecutionProgress)
    AsyncBareFactoryWorker* const that_;
  };

  virtual void Execute(const ExecutionProgress& progress) = 0;
  virtual void HandleProgressCallback(const T *data, size_t size) = 0;

 private:
  void Execute() /*final override*/ {
      ExecutionProgress progress(this);
      Execute(progress);
  }

  virtual void ConstructProgress_(Targs... Fargs) = 0;
};

template<class T, typename... Targs>
/* abstract */
class AsyncFactoryWorker : public AsyncBareFactoryWorker<T, Targs...> {
 public:
  explicit AsyncFactoryWorker(Callback *callback_)
      : AsyncBareFactoryWorker<T, Targs...>(callback_) {
    uv_mutex_init(&async_lock);
  }

  virtual ~AsyncFactoryWorker() {
    uv_mutex_lock(&async_lock);

    while (!asyncdata_.empty()) {
      T *data = asyncdata_.front();

      asyncdata_.pop();

      delete data;
    }

    uv_mutex_unlock(&async_lock);
    uv_mutex_destroy(&async_lock);
  }

  void WorkComplete() {
    WorkProgress();
    AsyncWorker::WorkComplete();
  }

  void WorkProgress() {
    uv_mutex_lock(&async_lock);

    while (!asyncdata_.empty()) {
      T *data = asyncdata_.front();
      asyncdata_.pop();
      uv_mutex_unlock(&async_lock);

      // Don't send progress events after we've already completed.
      if (this->callback) {
          this->HandleProgressCallback(data, 1);
      }

      delete data;

      uv_mutex_lock(&async_lock);
    }

    uv_mutex_unlock(&async_lock);
  }

 private:
  void ConstructProgress_(Targs... Fargs) {
    T *data = new T(Fargs...);

    uv_mutex_lock(&async_lock);
    asyncdata_.push(data);
    uv_mutex_unlock(&async_lock);

    uv_async_send(&this->async);
  }

  uv_mutex_t async_lock;
  std::queue<T*> asyncdata_;
};

#endif  // NAN_HAS_CPLUSPLUS_11
```

This works just like the new `AsyncProgressQueueWorker`, but instead of `std::copy`ing the data, the `progress.Construct` function passes along the object constructor arguments, constructing and delivering the object to the main thread.

It allows us to Async Construct objects even if they lack a public copy constructor, although it also does support copy constructors.

Sure, the constructor arguments end up getting copied, but not the entire object, which only gets constructed once.  Depending on the code being built around this worker, this could make things much more efficient.

Since `AsyncFactoryWorker` uses a variadic template, one can use it in many different ways.  Pass no constructor arguments to use the default constructor.  One can even use `&&T` as the constructor arguments and the worker will use the `move` constructor!!

Meanwhile, I can go on and on about describing this new class, but all this pull request does is factor out the `AsyncBareProgressWorker` so that the `uv_async_t` and `AsyncWorker` components can be reused.

How does this PR alone benefit users today?  If an addon module defines multiple AsyncProgress[Queue]Workers based on `AsyncProgress[Queue]Worker<T>` where `T` is actually used for more than one type, then the compiler will have to generate entire _NAN_ base classes for each worker.  However, with this patch applied, the `AsyncBareProgressWorkerBase` will be a non-templated common base class to all AsyncProgress[Queue]Workers, and will only be defined once (per code segment).  The generated template classes will be smaller.

Also, if there is any interest in a pull request that adds the above `AsyncFactoryWorker` classes (only when c++11 support is present), please let me know.  I'll likely publish a separate npm package for it otherwise.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1609,10 +1609,9 @@ class Callback {
   char *errmsg_;
 };
 
-template<class T>
-/* abstract */ class AsyncBareProgressWorker : public AsyncWorker {
+/* abstract */ class AsyncBareProgressWorkerBase : public AsyncWorker {
  public:
-  explicit AsyncBareProgressWorker(Callback *callback_)
+  explicit AsyncBareProgressWorkerBase(Callback *callback_)
       : AsyncWorker(callback_) {
     uv_async_init(
         uv_default_loop()
@@ -1622,10 +1621,42 @@ template<class T>
     async.data = this;
   }
 
-  virtual ~AsyncBareProgressWorker() {
+  virtual ~AsyncBareProgressWorkerBase() {
   }
 
   virtual void WorkProgress() = 0;
+
+  virtual void Destroy() {
+      uv_close(reinterpret_cast<uv_handle_t*>(&async), AsyncClose_);
+  }
+
+ private:
+  inline static NAUV_WORK_CB(AsyncProgress_) {
+    AsyncBareProgressWorkerBase *worker =
+            static_cast<AsyncBareProgressWorkerBase*>(async->data);
+    worker->WorkProgress();
+  }
+
+  inline static void AsyncClose_(uv_handle_t* handle) {
+    AsyncBareProgressWorkerBase *worker =
+            static_cast<AsyncBareProgressWorkerBase*>(handle->data);
+    delete worker;
+  }
+
+ protected:
+  uv_async_t async;
+};
+
+template<class T>
+/* abstract */
+class AsyncBareProgressWorker : public AsyncBareProgressWorkerBase {
+ public:
+  explicit AsyncBareProgressWorker(Callback *callback_)
+      : AsyncBareProgressWorkerBase(callback_) {
+  }
+
+  virtual ~AsyncBareProgressWorker() {
+  }
 
   class ExecutionProgress {
     friend class AsyncBareProgressWorker;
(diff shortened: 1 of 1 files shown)


kkoopa:
This PR in itself looks good. I am not sure as to what to do about C++11, but it would need a major version bump.

mkrufky:
@kkoopa so if you're willing to merge this one as-is, then i'll create another pull request later on with the c++11 class in the OP, and I'll expect that one might hang open for a while...

kkoopa:
Alright, that sounds good.

PR #701, opened 2017-10-07
AsyncProgressQueueWorker doesnt support ExecutionProgress.Signal()

mkrufky:
`AsyncProgressQueueWorker` doesnt support `ExecutionProgress.Signal()`

If we leave things as-is, `AsyncProgressQueueWorker::ExecutionProgress.Signal()` is a no-op.  We can try to fix that, but there isn't any point.  If somebody wants that functionality, they should just use an `AsyncProgressWorkerBase<T>`  Let's just remove the entry point by swapping out the `AsyncBareProgressWorker<T>` with a specialized `AsyncBareProgressQueueWorker<T>` that simply doesn't have an `ExecutionProgress.Signal()`

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1662,11 +1662,11 @@ class AsyncBareProgressWorker : public AsyncBareProgressWorkerBase {
     friend class AsyncBareProgressWorker;
    public:
     void Signal() const {
-        uv_async_send(&that_->async);
+      uv_async_send(&that_->async);
     }
 
     void Send(const T* data, size_t count) const {
-        that_->SendProgress_(data, count);
+      that_->SendProgress_(data, count);
     }
 
    private:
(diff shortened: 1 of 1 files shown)


nan.h
```diff
@@ -1745,10 +1745,47 @@ typedef AsyncProgressWorkerBase<char> AsyncProgressWorker;
...
+  }
+
+  virtual ~AsyncBareProgressQueueWorker() {
+  }
+
+  class ExecutionProgress {
+    friend class AsyncBareProgressQueueWorker;
+   public:
+    void Send(const T* data, size_t count) const {
+        that_->SendProgress_(data, count);
```
bnoordhuis: Indent error here and below.

kkoopa:
The inheritance changes should not make any difference, as the queue workers were only added now and have not been released yet.

nan.h
mkrufky: fixed in 76bee55889ed4ec3df42ba49c3e16acbbc7078d1  ...

PR #702, opened 2017-10-07
add AsyncProgressQueueWorker test for streams

mkrufky:
`AsyncProgressQueueWorker<T>` is a better choice over `AsyncProgressWorker<T>` when it comes to streaming, in most cases.  As such, we should include an example of how to use it.

This pull request adds an `AsyncProgressQueueWorker` test for streams, similar to the test for `AsyncProgressWorker`.   So similar, in fact, that this test is copied exactly from the `AsyncProgressWorker` test, but with all sleeps removed.

* The first patch copies the old test to a new test.

* The second patch shows the changes that I made to accommodate `AsyncProgressQueueWorker<T>`

### DIFF
--- a/test/cpp/asyncprogressqueueworkerstream.cpp
+++ b/test/cpp/asyncprogressqueueworkerstream.cpp
@@ -6,10 +6,6 @@
  * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
  ********************************************************************/
 
-#ifndef _WIN32
-#include <unistd.h>
-#define Sleep(x) usleep((x)*1000)
-#endif
 #include <nan.h>
 
 using namespace Nan;  // NOLINT(build/namespaces)
@@ -21,17 +17,15 @@ struct data_t {
   int data;
 };
 
-// Unlike test/cpp/ayncprogressworker.cpp this test is explicitly templated.
 template<typename T>
 class ProgressQueueWorker : public AsyncProgressQueueWorker<T> {
  public:
   ProgressQueueWorker(
       Callback *callback
     , Callback *progress
-    , int milliseconds
     , int iters)
     : AsyncProgressQueueWorker<T>(callback), progress(progress)
-    , milliseconds(milliseconds), iters(iters) {}
+    , iters(iters) {}
 
   ~ProgressQueueWorker() {
     delete progress;
@@ -44,7 +38,6 @@ class ProgressQueueWorker : public AsyncProgressQueueWorker<T> {
       data.index = i;
       data.data = i * 2;
       progress.Send(&data, 1);
-      Sleep(milliseconds);
     }
   }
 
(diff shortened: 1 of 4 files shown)


test/cpp/asyncprogressqueueworkerstream.cpp
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+ private:
+  Callback *progress;
+  int iters;
+};
+
+NAN_METHOD(DoProgress) {
+  Callback *progress = new Callback(info[1].As<v8::Function>());
+  Callback *callback = new Callback(info[2].As<v8::Function>());
```
bnoordhuis: Future improvement: update all instances of `val.As<v8::Function>()` to `To<v8::Function>(val).ToLocalChecked()` in the test suite.

test/cpp/asyncprogressqueueworkerstream.cpp
```diff
@@ -0,0 +1,80 @@
...
+  AsyncQueueWorker(new ProgressQueueWorker<data_t>(
+      callback
+    , progress
+    , To<uint32_t>(info[0]).FromJust()));
+}
+
+NAN_MODULE_INIT(Init) {
+  Set(target
+    , New<v8::String>("a").ToLocalChecked()
```
bnoordhuis: `"doProgress"`?

mkrufky: Again, since it's copied from the existing tests, I don't think we should change it in this PR.   But wouldn't be a bad cleanup afterwards.   I'd make the change now if it wasn't already in all the existing tests (...or if you ask me to do so)   ...   If we change it here, we'll have to change it everywhere else, too:
```
mk@vujade:~/git/nan$ git grep "\"a\"" test/cpp/*
test/cpp/asyncprogressqueueworker.cpp:    , New<v8::String>("a").ToLocalChecked()
test/cpp/asyncprogressworker.cpp:    , New<v8::String>("a").ToLocalChecked()
test/cpp/asyncprogressworkersignal.cpp:    , New<v8::String>("a").ToLocalChecked()
test/cpp/asyncprogressworkerstream.cpp:    , New<v8::String>("a").ToLocalChecked()
test/cpp/asyncworker.cpp:    , New<v8::String>("a").ToLocalChecked()
test/cpp/asyncworkererror.cpp:    , New("a").ToLocalChecked()
test/cpp/bufferworkerpersistent.cpp:    , New<v8::String>("a").ToLocalChecked()
test/cpp/setcallhandler.cpp:    , New("a").ToLocalChecked()
```

test/cpp/asyncprogressqueueworkerstream.cpp
mkrufky: should we turn this (and the `vector<T>.swap()` idea from pr 692: https://github.com/nodejs/nan/pull/692#discussion_r142992578 ) into github issues so that we dont forget, and maybe one of us (or maybe a noob looking to get involved) will have a place to find ideas for improvements?

test/cpp/asyncprogressqueueworkerstream.cpp
mkrufky: As much as "doProgress" would make sense here, I think uniformity amongst all of the tests is most important, and it's kinda nice how all of the tests export the addon function using the same name, `a` , but I'll agree that if I were writing new code today I would use a meaningful name.   If you want me to change it, I will... but I think it should remain, at least for this pull request.   If we want to make them meaningful names then lets chaneg them all at once after this is merged...  but I repeat, I am happy to comply with whatever you'd prefer.

mkrufky: On a second look, I realize that yes, there are some tests that call the addon `"a"` while others do have meaningful names.  So, I've changed my mind.  Renamed `"a"` to  `"doProgress"` in 0c0248ab1e62cd2c9b0d900a7113a194d1bf1ed7  ...  We should probably clean up those others at some point as well.

mkrufky:
The failure in the tests is unrelated to this pull request.  I fixed that in a separate PR: https://github.com/nodejs/nan/pull/703

mkrufky:
@kkoopa ping?  can we merge this along with 703 & 704 before the next release?  These three pull requests clean up the tests and serve as better examples of how to use the new worker template.

kkoopa:
I think so. Will look into it in a week.

Issue #670, opened 2017-05-17
nan_json.h(89): error C2660: 'v8::JSON::Parse' : function does not take 2 arguments


YLivay:
Hi, 
since around a month ago, we've been having issues building native modules for nwjs.
The problematic dependency is lzma-native which depends on nan@^2.3.2.

The error logs I'm seeing seems to be related to recent breaking changes with json parsing:
```
c:\projects\gamejolt\build\client\prod\source\node_modules\nan\nan_json.h(89): error C2660: 'v8::JSON::Parse' : function does not take 2 arguments [C:\projects\gamejolt\build\client\prod\source\node_modules\lzma-native\build\lzma_native.vcxproj]
```
Is this related in some way to https://github.com/nodejs/nan/issues/658?

bnoordhuis:
In a way.  I suspect it's because nw.js announces (through the NODE_MODULE_VERSION macro) that it's compatible with node.js v6 when in fact it ships a newer version of V8 than ships with the real node.js v6.

I don't know of a way to reliably distinguish between node.js and nw.js so this might be unfixable in nan.  It wouldn't be an issue if nw.js had its own NODE_MODULE_VERSION but it doesn't, as far as I can tell.

mkrufky:
It might be worth the test of forcing usage of older nan version 2.6.1, before `Fix v8::JSON::Parse() deprecation warning.` got applied.  Before that commit, we used a single argument version of `v8::JSON::Parse`, which will generate warnings against some node versions but should continue to build successfully.

If this works for you, it will be good for us to know, but like Ben said, I'm not sure of the best way to fix it moving forward.

mkrufky:
I believe this issue is resolved in _nan_ version `2.8.0` ... It would be really nice if somebody can confirm that and then we can close this issue.....

YLivay:
Hey @mkrufky thanks a ton for looking into it, and sorry for not replying earlier. In the frenzy of trying to find a workaround i forgot to check back on the issue :(

Our build no longer depends on native modules and setting up the old environment is too time consuming atm, but i'll get to it and update back here as soon as I can. You can close the issue if you think it's fixed, i'll recreate it if the issue persists.

PR #706, opened 2017-10-24
Fix compiling on io.js 1 & 2

abalabahaha:
This fixes a compilation error on io.js 1 and 2. V8 4.2.77 (io.js 2.5.0) still uses `Parse(Local<String> json_string)`, and V8 4.4.63 (io.js 3.3.1) uses `Parse(Isolate *isolate, Local<String> json_string)`

### DIFF
--- a/nan_json.h
+++ b/nan_json.h
@@ -88,7 +88,8 @@ class JSON {
     v8::Isolate* context_or_isolate = v8::Isolate::GetCurrent();
 #endif  // NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION
     result = v8::JSON::Parse(context_or_isolate, json_string);
-#endif  // NODE_MODULE_VERSION >= NODE_0_12_MODULE_VERSION && NODE_MODULE_VERSION <= IOJS_2_0_MODULE_VERSION
+#endif  // NODE_MODULE_VERSION >= NODE_0_12_MODULE_VERSION &&
+        // NODE_MODULE_VERSION <= IOJS_2_0_MODULE_VERSION
     if (result.IsEmpty()) return v8::Local<v8::Value>();
     return scope.Escape(result.ToLocalChecked());
 #endif  // NAN_JSON_H_NEED_PARSE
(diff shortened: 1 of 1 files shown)


nan_json.h
```diff
@@ -78,7 +78,7 @@ class JSON {
     return scope.Escape(parse(json_string));
 #else
     Nan::MaybeLocal<v8::Value> result;
-#if NODE_MODULE_VERSION == NODE_0_12_MODULE_VERSION
+#if NODE_MODULE_VERSION >= NODE_0_12_MODULE_VERSION && NODE_MODULE_VERSION <= IOJS_2_0_MODULE_VERSION
```
bnoordhuis: Long line.  Can you put a `\` and line break after the `&&`?

mkrufky:
Nice.  Thank you for this.  I think this might also resolve #670

kkoopa:
I'll be happy to merge once the nit is resolved.

nan_json.h
```diff
@@ -87,7 +87,7 @@ class JSON {
     v8::Isolate* context_or_isolate = v8::Isolate::GetCurrent();
 #endif  // NODE_MODULE_VERSION > NODE_6_0_MODULE_VERSION
     result = v8::JSON::Parse(context_or_isolate, json_string);
-#endif  // NODE_MODULE_VERSION == NODE_0_12_MODULE_VERSION
+#endif  // NODE_MODULE_VERSION >= NODE_0_12_MODULE_VERSION && NODE_MODULE_VERSION <= IOJS_2_0_MODULE_VERSION
```
abalabahaha: Should this comment also be split into 2 lines?

abalabahaha:
@bnoordhuis I actually only came across this when doing some test builds for a project. I did notice that the CI configs no longer test against io.js, but the README and several places in the code still reference these versions, and the fix seemed small enough that it wouldn’t be a huge hassle.

nan_json.h
kkoopa: Depends on if the linter is smart enough to comprehend it. Try `make lint` and see what it says.
I am guessing it will complain about too long line or missing comment...

Issue #707, opened 2017-10-31
Deprecated warning with nan v2.7.0 and Node.js 9.0.0


fivdi:
When a native module using nan v2.7.0 is installed with Node.js 9.0.0 there is a deprecated warning. For example, here's the warning shown when installing the epoll module:

```
pi@raspberrypi:~/epoll $ npm i epoll

> epoll@1.0.0 install /home/pi/epoll/node_modules/epoll
> node-gyp rebuild

make: Entering directory '/home/pi/epoll/node_modules/epoll/build'
  CXX(target) Release/obj.target/epoll/src/epoll.o
In file included from ../../nan/nan.h:192:0,
                 from ../src/epoll.cc:15:
../../nan/nan_maybe_43_inl.h: In function ‘Nan::Maybe<bool> Nan::ForceSet(v8::Local<v8::Object>, v8::Local<v8::Value>, v8::Local<v8::Value>, v8::PropertyAttribute)’:
../../nan/nan_maybe_43_inl.h:112:73: warning: ‘v8::Maybe<bool> v8::Object::ForceSet(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>, v8::PropertyAttribute)’ is deprecated: Use CreateDataProperty / DefineOwnProperty [-Wdeprecated-declarations]
   return obj->ForceSet(isolate->GetCurrentContext(), key, value, attribs);
                                                                         ^
In file included from /home/pi/.node-gyp/9.0.0/include/node/v8.h:26:0,
                 from ../src/epoll.cc:11:
/home/pi/.node-gyp/9.0.0/include/node/v8.h:3115:29: note: declared here
                 Maybe<bool> ForceSet(Local<Context> context, Local<Value> key,
                             ^
/home/pi/.node-gyp/9.0.0/include/node/v8config.h:318:3: note: in definition of macro ‘V8_DEPRECATED’
   declarator __attribute__((deprecated(message)))
   ^~~~~~~~~~
  SOLINK_MODULE(target) Release/obj.target/epoll.node
  COPY Release/epoll.node
make: Leaving directory '/home/pi/epoll/node_modules/epoll/build'
npm WARN saveError ENOENT: no such file or directory, open '/home/pi/epoll/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/home/pi/epoll/package.json'
npm WARN epoll No description
npm WARN epoll No repository field.
[...]

Issue #504, opened 2015-10-28
Stuff V8 will be breaking soon


kkoopa:
Skimmed through v8.h@4.8.168, these will be deprecated soon:
- ForceSet -> DefineOwnProperty (prefer CreateDataProperty)
- Function::New -> return MaybeLocal
- Object::SetAccessor -> return Maybe

I don't think it's worth breaking compatibility with a new major for this drivel yet. I'd like to keep NAN 2 working for at least a year (given no major overhauls of V8), so I propose aliasing `Nan::ForceSet` to `v8::DefineOwnProperty`, ignoring `v8::CreateDataProperty`, and un-maybefying `v8::Function::New` and `v8::Object::SetAccessor`.

rvagg:
> and un-maybefying v8::Function::New and v8::Object::SetAccessor.

will be a breaking change no?

kkoopa:
No, the idea is to keep them working the current way, without maybes.

kkoopa:
However, what is the actual Node release policy now after 5 has been released? My understanding is that it will be around for half a year or so, but won't be an LTS release, which 4 is. Will there be a 6, 7, 8, and 9 (given V8's six week cycle) in those six months, or are future major versions (and thereby V8 upgrades) off the table until six months have passed?

If Node 6 won't be out for six months, the best option is probably to wait for more breaking V8 changes to accumulate in the meantime and then to break the API if more changes have accumulated, or to keep compatibility if these are the only offenders or they can be reasonably covered with only additive changes.

As mentioned elsewhere, there will be a major overhaul the day we can finally stop supporting 0.8 and 0.10, either voluntarily, or because it simply is not feasible to keep it working any longer.

fivdi:
@kkoopa the Node.js LTS release schedule at the [bottom of this page](https://github.com/nodejs/LTS/) should provide some of the information you're looking for.

kkoopa:
Alright, if that schedule will be held, it seems like we could even drop 0.8, 0.10, 0.12 and all of io.js in a year from now, moving on to the regular schedule of supporting two LTSs after that. However, it is still hard to determine such things in advance. It should at least not be necessary to do a new major of NAN until April next year.

trevnorris:
> so I propose aliasing `Nan::ForceSet` to `v8::DefineOwnProperty`

From `includes.v8.h`:
```cpp
  // Defines a configurable, writable, enumerable property with the given value
  // on the object unless the property already exists and is not configurable
[...]

PR #708, opened 2017-10-31
Use DefineOwnProperty as ForceSet for Node 9

kkoopa:
Main question is if this truly is a breaking change or not. Allegedly, cases where `ForceSet` work and `DefineOwnProperty` do not break V8's object model and are undefined behavior.
Is there a better fix? Are enough edge cases covered? Can we do more?
#707 #504


nan_maybe_43_inl.h
```diff
@@ -109,7 +109,12 @@ inline Maybe<bool> ForceSet(
   , v8::PropertyAttribute attribs = v8::None) {
   v8::Isolate *isolate = v8::Isolate::GetCurrent();
   v8::HandleScope scope(isolate);
+#if NODE_MODULE_VERSION >= NODE_9_0_MODULE_VERSION
+  return obj->DefineOwnProperty(isolate->GetCurrentContext(),
+                                key.As<v8::Name>(), value, attribs);
```
bnoordhuis: Needs an `key->IsName()` check.  The cast is unsafe, not every Value is a Name.

kkoopa:
Leave what be? Nan::ForceSet? Yes, it dawned on me that this does not break backwards compatibility as is. Although it would be more consistent to limit old ForceSet versions too. 

What about adding a Nan::DefineOwnProperty, marking Nan::ForceSet as deprecated and leaving it at that?

bnoordhuis:
Sounds like a plan.

kkoopa:
This is what I had in mind.

kkoopa:
Thanks for a swift response. However, I noticed that I had negated the configurable test and that the standard says a non-configurable property may still be made read-only. This last commit should fix that.

nan_maybe_pre_43_inl.h
```diff
@@ -153,7 +153,10 @@ inline Maybe<bool> DefineOwnProperty(
   , v8::Handle<v8::String> key
   , v8::Handle<v8::Value> value
   , v8::PropertyAttribute attribs = v8::None) {
-  return obj->GetPropertyAttributes(key) == v8::DontDelete
+  v8::PropertyAttribute current = obj->GetPropertyAttributes(key);
+  return !(current & v8::DontDelete) ||
+                 !(current & v8::ReadOnly) &&
+                     !((attribs ^ current) & ~v8::ReadOnly)
              ? Just<bool>(obj->ForceSet(key, value, attribs))
              : Nothing<bool>();
```
bnoordhuis: I'd argue that this is easier to read written in guard style:
```c++
if (current & v8::DontDelete) return Nothing<bool>();
if (current & v8::ReadOnly) return Nothing<bool>();
// etc.
```

kkoopa: I disagree, but since you are doing the reviewing, I changed it. The proposed style does not work, and now it looks messier to me. The way I read the spec it should return `Nothing` iff the property is non-configurable, read-only and flags other than read-only disagree.

nan_maybe_pre_43_inl.h
```diff
@@ -148,7 +148,22 @@ inline Maybe<bool> Set(
...
+    v8::Handle<v8::Object> obj
+  , v8::Handle<v8::String> key
+  , v8::Handle<v8::Value> value
+  , v8::PropertyAttribute attribs = v8::None) {
+  v8::PropertyAttribute current = obj->GetPropertyAttributes(key);
+  if (current & v8::DontDelete) {
+    if (!(current & v8::ReadOnly) && !((attribs ^ current) & ~v8::ReadOnly)) {
+      return Just<bool>(obj->ForceSet(key, value, attribs));
+    }
+      return Nothing<bool>();
```
mkrufky: whitespace breakage, should remove 2 spaces

bnoordhuis:
Okay, I agree the new code is not much of an improvement but the old code took me 30 seconds to decipher and that's assuming I got it right.

I'd say that makes it too clever by half so at least add a comment explaining what it does. :-)

kkoopa:
Will do. I agree it is cryptic, but I do not see an easy way of making it elegant. I will make a new release soon so Node 9 does not suffer deprecation warnings.

I am still not sure if everything is entirely correct in regards to emulating DefineOwnProperty, but I believe this is as close as it gets and it only applies to ancient versions which really should not be used. I could try adding some tests, but I suspect that it is not possible to get fully correct behavior. If someone else submits a better patch I will merge it.

kkoopa:
Added comments.

PR #711, opened 2017-11-06
remove my email address from travis

rvagg:
does someone else want to replace me on this? I don't read them and don't react to them


kkoopa:
I went ahead and merged it straight away so you would not get more notifications. I doubt anyone would read the notifications, I check the status page manually whenever I want to know the build status.

PR #714, opened 2017-11-12
convert uses of `val.As<Foo>()` to `To<Foo>(val).ToLocalChecked()`

mkrufky:
Since a new release seems to be looming around the corner, better to fix these now...

Expanding on bnoordhuis' suggestion here: https://github.com/nodejs/nan/pull/702#discussion_r143352671

The changes are mostly in `test/cpp`, `doc` and `examples` but there were two callers of `To<v8::Function>`  in `nan_json.h` that got converted here as well.

### DIFF
--- a/doc/v8_internals.md
+++ b/doc/v8_internals.md
@@ -33,7 +33,7 @@ NAN_GC_CALLBACK(gcPrologueCallback) {
 }
 
 NAN_METHOD(Hook) {
-  callback.Reset(args[0].As<Function>());
+  callback.Reset(To<Function>(args[0]).ToLocalChecked());
   Nan::AddGCPrologueCallback(gcPrologueCallback);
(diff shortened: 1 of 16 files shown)


bnoordhuis:
`handle.As<T>()` is legal when it's known the handle points to a T.

Granted, we don't properly check that everywhere, but where we do, `To<T>(handle).ToLocalChecked()` is just a slower way of doing `handle.As<T>()`.

mkrufky:
OK, so I went ahead and reverted the change to `nan_json.h`, leaving this PR only to touch `test/` , `doc/` and `examples/`

mkrufky:
(squashed)

kkoopa:
The tests make the assumption that sanitation and validation is handled at  the JavaScript side, but I do not mind these changes. Everything seems to be in order and all tests pass CI. You can go ahead and merge.

PR #718, opened 2017-11-16
nan_maybe_pre_43_inl.h: fix warnings in DefineOwnProperty()

### LINKED ISSUE #717: DefineOwnProperty: warning: suggest parentheses around ‘&&’ within ‘||’
_nan_ v2.8.0 generates the following warning under node versions `0.10.x` & `0.12.x`:
```
In file included from ../node_modules/nan/nan.h:199:0,
                 from ../src/dvbtee-parser.cc:11:
../node_modules/nan/nan_maybe_pre_43_inl.h: In function ‘Nan::Maybe<bool> Nan::DefineOwnProperty(v8::Handle<v8::Object>, v8::Handle<v8::String>, v8::Handle<v8::Value>, v8::PropertyAttribute)’:
../node_modules/nan/nan_maybe_pre_43_inl.h:158:44: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
                  !(current & v8::ReadOnly) &&               // writable AND
                                            ^
```
It looks like it should be an easy fix and I'd submit a patch myself, but I'm not exactly sure of the correct order of operations, there.

...and to whomever fixes this, while you're at it, there's a whitespace / linter issue in the same code block:
```
~/git/nan$ make
[...]
nan_maybe_pre_43_inl.h:159:  At least two spaces is best between code and comments  [whitespace/comments] [2]
[...]
```

here is the code in question:
```c++
inline Maybe<bool> DefineOwnProperty(
    v8::Handle<v8::Object> obj
  , v8::Handle<v8::String> key
  , v8::Handle<v8::Value> value
  , v8::PropertyAttribute attribs = v8::None) {
  v8::PropertyAttribute current = obj->GetPropertyAttributes(key);
  return !(current & v8::DontDelete) ||                     // configurable OR
                 !(current & v8::ReadOnly) &&               // writable AND
                     !((attribs ^ current) & ~v8::ReadOnly) // same excluding RO
             ? Just<bool>(obj->ForceSet(key, value, attribs))
             : Nothing<bool>();
}
```

The linter just wants another space before `// same excluding RO`, but I'm not exactly sure where the parenthesis go to silence the build warning.

mkrufky:
I think I agree with the compiler. The function was difficult to read. I think it's easier (and prettier) to just do as it says.

closes #717

### DIFF
--- a/nan_maybe_pre_43_inl.h
+++ b/nan_maybe_pre_43_inl.h
@@ -156,7 +156,7 @@ inline Maybe<bool> DefineOwnProperty(
   v8::PropertyAttribute current = obj->GetPropertyAttributes(key);
   return !(current & v8::DontDelete) ||                     // configurable OR
                  !(current & v8::ReadOnly) &&               // writable AND
-                     !((attribs ^ current) & ~v8::ReadOnly) // same excluding RO
+                 !((attribs ^ current) & ~v8::ReadOnly)     // same excluding RO
              ? Just<bool>(obj->ForceSet(key, value, attribs))
              : Nothing<bool>();
 }
--- a/nan_maybe_pre_43_inl.h
+++ b/nan_maybe_pre_43_inl.h
@@ -155,8 +155,8 @@ inline Maybe<bool> DefineOwnProperty(
   , v8::PropertyAttribute attribs = v8::None) {
   v8::PropertyAttribute current = obj->GetPropertyAttributes(key);
   return !(current & v8::DontDelete) ||                     // configurable OR
-                 !(current & v8::ReadOnly) &&               // writable AND
-                 !((attribs ^ current) & ~v8::ReadOnly)     // same excluding RO
+                  (!(current & v8::ReadOnly) &&             // writable AND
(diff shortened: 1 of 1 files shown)


nan_maybe_pre_43_inl.h
```diff
@@ -155,8 +155,8 @@ inline Maybe<bool> DefineOwnProperty(
   , v8::PropertyAttribute attribs = v8::None) {
   v8::PropertyAttribute current = obj->GetPropertyAttributes(key);
   return !(current & v8::DontDelete) ||                     // configurable OR
-                 !(current & v8::ReadOnly) &&               // writable AND
-                     !((attribs ^ current) & ~v8::ReadOnly) // same excluding RO
+                ( !(current & v8::ReadOnly) &&              // writable AND
+                  !((attribs ^ current) & ~v8::ReadOnly) )  // same excluding RO
```
kkoopa: No spaces inside parentheses, please.

mkrufky: fixed

PR #728, opened 2018-01-30
Fix compilation on io.js 3

### LINKED ISSUE #725: error: ‘class v8::Object’ has no member named ‘DefineOwnProperty’
Looks like io.js 3.2 don't have `DefineOwnProperty` on `v8::Object`.

https://travis-ci.org/TooTallNate/ref/jobs/331158677

```
> weak@1.0.1 install /home/travis/build/TooTallNate/ref/node_modules/weak
> node-gyp rebuild
make: Entering directory `/home/travis/build/TooTallNate/ref/node_modules/weak/build'
  CXX(target) Release/obj.target/weakref/src/weakref.o
In file included from ../../nan/nan.h:197:0,
                 from ../src/weakref.cc:18:
../../nan/nan_maybe_43_inl.h: In function ‘Nan::Maybe<bool> Nan::DefineOwnProperty(v8::Local<v8::Object>, v8::Local<v8::String>, v8::Local<v8::Value>, v8::PropertyAttribute)’:
../../nan/nan_maybe_43_inl.h:112:15: error: ‘class v8::Object’ has no member named ‘DefineOwnProperty’
   return obj->DefineOwnProperty(isolate->GetCurrentContext(), key, value,
               ^
make: *** [Release/obj.target/weakref/src/weakref.o] Error 1
make: Leaving directory `/home/travis/build/TooTallNate/ref/node_modules/weak/build'
```

kkoopa:
Fixes #725

### DIFF
--- a/nan_maybe_43_inl.h
+++ b/nan_maybe_43_inl.h
@@ -109,8 +109,22 @@ inline Maybe<bool> DefineOwnProperty(
   , v8::PropertyAttribute attribs = v8::None) {
   v8::Isolate *isolate = v8::Isolate::GetCurrent();
   v8::HandleScope scope(isolate);
+#if NODE_MODULE_VERSION >= NODE_4_0_MODULE_VERSION
   return obj->DefineOwnProperty(isolate->GetCurrentContext(), key, value,
                                 attribs);
+#else
+  Maybe<v8::PropertyAttribute> maybeCurrent =
+      obj->GetPropertyAttributes(isolate->GetCurrentContext(), key);
+  if (maybeCurrent.IsNothing()) {
+    return Nothing<bool>();
+  }
+  v8::PropertyAttribute current = maybeCurrent.FromJust();
+  return !(current & v8::DontDelete) ||                     // configurable OR
+                  (!(current & v8::ReadOnly) &&             // writable AND
+                   !((attribs ^ current) & ~v8::ReadOnly))  // same excluding RO
+             ? Just<bool>(obj->ForceSet(key, value, attribs))
+             : Nothing<bool>();
+#endif
 }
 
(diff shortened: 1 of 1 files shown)


nan_maybe_43_inl.h
```diff
@@ -109,8 +109,22 @@ inline Maybe<bool> DefineOwnProperty(
...
+    return Nothing<bool>();
+  }
+  v8::PropertyAttribute current = maybeCurrent.FromJust();
+  return !(current & v8::DontDelete) ||                     // configurable OR
+                  (!(current & v8::ReadOnly) &&             // writable AND
+                   !((attribs ^ current) & ~v8::ReadOnly))  // same excluding RO
+             ? Just<bool>(obj->ForceSet(key, value, attribs))
+             : Nothing<bool>();
+#endif
 }
```
mkrufky: I realize that it is by design, but this code is identical to the block in the `nan_maybe_pre_43_inl.h` version: https://github.com/nodejs/nan/blob/master/nan_maybe_pre_43_inl.h#L151

How about we add a new file called, `nan_maybe.h` or `nan_maybe_inl.h` and move the logic of deciding which header to use from `nan.h` to this new file, along with a block of common code (such as this duplicated block)

I'd hate to see future maintenance improve one block and possibly forget to include the other.

kkoopa: I thought of that while writing it, but it is not fully identical. There is some `Maybe`-stuff going on before the `return` statement, which makes deduplication harder.

mkrufky: ...but not impossible.   Meanwhile, my desire for deduplication shouldn't hold up this PR.  We can always take a stab at that later.

kkoopa:
Thanks for reviewing. I will merge now.

PR #730, opened 2018-02-07
Drop osx builds with Node 0.12

ofrobots:
As per https://github.com/nodejs/nan/pull/729#issuecomment-363618436, stop testing against Node.js 0.12 on OSX.


ofrobots:
This is ready for review. @kkoopa PTAL.

PR #731, opened 2018-02-09
Add context awareness to Nan::AsyncWorker and Nan::Callback

ofrobots:
There are three commits here. Each can be individually reviewed.

Background: https://github.com/nodejs/node/issues/13254.

Continuing on from #729, this adds async context support to Nan::Callback and Nan::AsyncWorker (and subclasses). 

I have not added any deprecations to the legacy `Nan::Callback::Call`s versions yet. I propose to do that in the next PR.

### DIFF
--- a/doc/asyncworker.md
+++ b/doc/asyncworker.md
@@ -15,7 +15,7 @@
 This class internally handles the details of creating an [`AsyncResource`][AsyncResource], and running the callback in the
 correct async context. To be able to identify the async resources created by this class in async-hooks, provide a
 `resource_name` to the constructor. It is recommended that the module name be used as a prefix to the `resource_name` to avoid
-collisions in the names. For more details see [`AsyncResource`][AsyncResource] documentation.  The `resource_name` needs to stay valid for the lifetime of worker instance.
+collisions in the names. For more details see [`AsyncResource`][AsyncResource] documentation.  The `resource_name` needs to stay valid for the lifetime of the worker instance.
 
 Definition:
 
--- a/doc/asyncworker.md
+++ b/doc/asyncworker.md
@@ -12,12 +12,17 @@
 
 `Nan::AsyncWorker` is an _abstract_ class that you can subclass to have much of the annoying asynchronous queuing and handling taken care of for you. It can even store arbitrary V8 objects for you and have them persist while the asynchronous work is in progress.
 
+This class internally handles the details of creating an [`AsyncResource`][AsyncResource], and running the callback in the
+correct async context. To be able to identify the async resources created by this class in async-hooks, provide a
+`resource_name` to the constructor. It is recommended that the module name be used as a prefix to the `resource_name` to avoid
+collisions in the names. For more details see [`AsyncResource`][AsyncResource] documentation.  The `resource_name` needs to stay valid for the lifetime of worker instance.
+
 Definition:
 
 ```c++
 class AsyncWorker {
  public:
-  explicit AsyncWorker(Callback *callback_);
+  explicit AsyncWorker(Callback *callback_, const char* resource_name = "nan:AsyncWorker");
 
   virtual ~AsyncWorker();
 
@@ -73,9 +78,9 @@ Definition:
 template<class T>
 class AsyncProgressWorkerBase<T> : public AsyncWorker {
  public:
-  explicit AsyncProgressWorker(Callback *callback_);
+  explicit AsyncProgressWorkerBase(Callback *callback_, const char* resource_name = ...);
 
-  virtual ~AsyncProgressWorker();
+  virtual ~AsyncProgressWorkerBase();
 
   void WorkProgress();
 
@@ -108,7 +113,7 @@ Definition:
 template<class T>
 class AsyncProgressQueueWorker<T> : public AsyncWorker {
  public:
-  explicit AsyncProgressQueueWorker(Callback *callback_);
+  explicit AsyncProgressQueueWorker(Callback *callback_, const char* resource_name = "nan:AsyncProgressQueueWorker");
 
   virtual ~AsyncProgressQueueWorker();
 
@@ -137,3 +142,5 @@ Definition:
 ```c++
 void AsyncQueueWorker(AsyncWorker *);
 ```
+
+[AsyncResource]: "node_misc.html#api_nan_asyncresource"
--- a/doc/callback.md
+++ b/doc/callback.md
@@ -22,14 +22,14 @@ class Callback {
 
   v8::Local<v8::Function> operator*() const;
 
-  v8::Local<v8::Value> operator()(AsyncResource* async_resource,
-                                  v8::Local<v8::Object> target,
-                                  int argc = 0,
-                                  v8::Local<v8::Value> argv[] = 0) const;
+  MaybeLocal<v8::Value> operator()(AsyncResource* async_resource,
+                                   v8::Local<v8::Object> target,
+                                   int argc = 0,
+                                   v8::Local<v8::Value> argv[] = 0) const;
 
-  v8::Local<v8::Value> operator()(AsyncResource* async_resource,
-                                  int argc = 0,
-                                  v8::Local<v8::Value> argv[] = 0) const;
+  MaybeLocal<v8::Value> operator()(AsyncResource* async_resource,
+                                   int argc = 0,
+                                   v8::Local<v8::Value> argv[] = 0) const;
 
   void SetFunction(const v8::Local<v8::Function> &fn);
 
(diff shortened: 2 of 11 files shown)


nan.h
```diff
@@ -1552,10 +1567,62 @@ class Callback {
...
+  v8::Local<v8::Value> Call_(v8::Isolate *isolate
+                           , v8::Local<v8::Object> target
+                           , int argc
+                           , v8::Local<v8::Value> argv[]
+                           , AsyncResource* resource) const {
+    EscapableHandleScope scope;
+
+    v8::Local<v8::Function> func = New(handle_);
+    return scope.Escape(resource->runInAsyncScope(target, func, argc, argv)
+                                  .ToLocalChecked());
```
ofrobots: Admittedly, I am a little bit unsure here about the `ToLocalChecked`. I think we ought to change the prototype here to return a `MaybeLocal` rather than a `Local`. I need to go figure out `Escape` and `MaybeLocal` are supposed to interact.

kkoopa: It is a legacy remnant from the time when `v8::Function::Call` returned a `Local<T>`. I think everything dealing with `v8::Function` works this way. It's been planned to change in the next major release for years, but that release has not had to happen so far.

As for `MaybeLocal` and `Escape`, there is a non-explicit conversion constructor `MaybeLocal(Local<T>)`, so just escape the `Local` and it will be wrapped on return.

nan.h
```diff
@@ -1668,13 +1668,15 @@ class Callback {
...
-  explicit AsyncWorker(Callback *callback_)
+  explicit AsyncWorker(Callback *callback_,
+                       const char* resource_name = "nan:AsyncWorker")
       : callback(callback_), errmsg_(NULL) {
     request.data = this;
 
     HandleScope scope;
     v8::Local<v8::Object> obj = New<v8::Object>();
     persistentHandle.Reset(obj);
+    async_resource = new AsyncResource(obj, resource_name);
```
kkoopa: Could this not  be a direct member of the class or am I missing something?

ofrobots: `AsyncResource` doesn't have a default constructor. It needs the resource object which is not constructed until later in `AsyncWorker` constructor here.

.gitignore
```diff
@@ -1,2 +1,3 @@
 build/
-node_modules/
\ No newline at end of file
+node_modules/
+.vscode/
```
bnoordhuis: Global .gitignore?

doc/asyncworker.md
```diff
@@ -73,7 +78,7 @@ Definition:
 template<class T>
 class AsyncProgressWorkerBase<T> : public AsyncWorker {
  public:
-  explicit AsyncProgressWorker(Callback *callback_);
+  explicit AsyncProgressWorker(Callback *callback_, const char* resource_name = ...);
```
bnoordhuis: Since you're here: the constructor name is wrong, it should be AsyncProgressWorkerBase.  Ditto the destructor.

nan.h
```diff
@@ -1362,13 +1346,15 @@ class Utf8String {
...
-          static_cast<node::async_context>(asyncContext));
+          v8::Isolate::GetCurrent(), target, method, argc, argv, asyncContext);
 #endif
     }
 
    private:
-    async_context asyncContext;
+    NAN_DISALLOW_ASSIGN_COPY_MOVE(AsyncResource)
+#if NODE_MODULE_VERSION >= NODE_8_0_MODULE_VERSION
+    node::async_context asyncContext;
```
bnoordhuis: That gratuitous camelCasing when everything else uses snake_case... it's giving me facial tics.

nan.h
```diff
@@ -1552,10 +1553,62 @@ class Callback {
 #endif
   }
 
+  inline v8::Local<v8::Value>
+  Call(v8::Local<v8::Object> target
+     , int argc
+     , v8::Local<v8::Value> argv[]
+     , AsyncResource* resource) const {
+#if (NODE_MODULE_VERSION >= NODE_8_0_MODULE_VERSION)
```
bnoordhuis: Superfluous parens here and below.

test/cpp/callbackcontext.cpp
```diff
@@ -0,0 +1,66 @@
+/*********************************************************************
+ * NAN - Native Abstractions for Node.js
+ *
+ * Copyright (c) 2018 NAN contributors
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
+
+#include <nan.h>
+#include <unistd.h>
```
bnoordhuis: Needs special-casing for windows.  We have 5 repetitions of that in test/cpp already so it's about time it gets factored out into a header file.

test/js/callbackcontext-test.js
```diff
@@ -0,0 +1,70 @@
...
+ * NAN - Native Abstractions for Node.js
+ *
+ * Copyright (c) 2018 NAN contributors
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
+
+try {
+  require('async_hooks');
+} catch (e) {
```
bnoordhuis: Can you print a skip message here?

doc/asyncworker.md
```diff
@@ -12,12 +12,17 @@
 
 `Nan::AsyncWorker` is an _abstract_ class that you can subclass to have much of the annoying asynchronous queuing and handling taken care of for you. It can even store arbitrary V8 objects for you and have them persist while the asynchronous work is in progress.
 
+This class internally handles the details of creating an [`AsyncResource`][AsyncResource], and running the callback in the
+correct async context. To be able to identify the async resources created by this class in async-hooks, provide a
+`resource_name` to the constructor. It is recommended that the module name be used as a prefix to the `resource_name` to avoid
+collisions in the names. For more details see [`AsyncResource`][AsyncResource] documentation.
```
bnoordhuis: Can you document that `resource_name` needs to stay valid for the lifetime of the instance?

test/cpp/callbackcontext.cpp
kkoopa: Yes, it should be factored out. Windows CI runs are here https://ci.appveyor.com/project/RodVagg/nan

doc/callback.md
```diff
@@ -39,6 +41,23 @@ class Callback {
 
   void Reset();
 
+  v8::Local<v8::Value> Call(v8::Local<v8::Object> target,
```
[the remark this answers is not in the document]
ofrobots: People can, and very commonly do, use a single `Nan::Callback` to repeatedly call the same function multiple times, potentially in difference asynchronous contexts. I think it would be dangerous, i.e. prone to misuse, if we associated an async resource with a `Callback` instance – given how this class is used / intended to be used.

From a convenience point of view, `AsyncWorker` provides a significantly more convenient API that handles async resource management automatically. I don't think we should make this class more convenient at the expense of making it dangerous.

.gitignore
ofrobots: Will do.

nan.h
ofrobots: Sorry for the tics. This simply follows the clash of styles that happened in `node.h`. Apologies if I made the tics worse now. The `async_context` class was defined by the n-api project which, being a C-only, project used a very different style than rest of Node.

I will avoid camel case here.

doc/callback.md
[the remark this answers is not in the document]
ofrobots: Some previous discussion on this topic: https://github.com/nodejs/node/issues/13254#issuecomment-357392244

doc/asyncworker.md
ofrobots: Done.

nan.h
ofrobots: Done.

test/cpp/callbackcontext.cpp
ofrobots: I've refactored `Sleep` into `sleep.h`. More aggressive refactoring of `asyncresource.cpp` and `callbackcontext.cpp` is possible, but IMO it would be keep those as is to serve as examples.

test/js/callbackcontext-test.js
ofrobots: Done.

doc/asyncworker.md
ofrobots: Done.

nan.h
ofrobots: Done.

nan.h
ofrobots: Thanks! Done.

doc/asyncworker.md
```diff
@@ -12,12 +12,17 @@
 
 `Nan::AsyncWorker` is an _abstract_ class that you can subclass to have much of the annoying asynchronous queuing and handling taken care of for you. It can even store arbitrary V8 objects for you and have them persist while the asynchronous work is in progress.
 
+This class internally handles the details of creating an [`AsyncResource`][AsyncResource], and running the callback in the
+correct async context. To be able to identify the async resources created by this class in async-hooks, provide a
+`resource_name` to the constructor. It is recommended that the module name be used as a prefix to the `resource_name` to avoid
+collisions in the names. For more details see [`AsyncResource`][AsyncResource] documentation.  The `resource_name` needs to stay valid for the lifetime of worker instance.
```
bnoordhuis: s/of/of the/

test/cpp/sleep.h
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
+ ********************************************************************/
+
+#ifndef _WIN32
+#include <unistd.h>
+#define Sleep(x) usleep((x)*1000)
+#endif
+
+
```
bnoordhuis: Can you remove the extra blank lines?

doc/asyncworker.md
mkrufky: documentation bug - my bad.  nice catch!

kkoopa:
Is this ready to merge now?

ofrobots:
All good from my side.

kkoopa:
Thanks everybody.

PR #732, opened 2018-02-13
AsyncResource: Local param instead of MaybeLocal

ofrobots:
No need to use a `MaybeLocal`. It is possible to pass an empty `Local`
to indicated that an object should be allocated automatically.


doc/node_misc.md
```diff
@@ -21,8 +21,8 @@ Definition:
 ```c++
 class AsyncResource {
  public:
-  AsyncResource(MaybeLocal<v8::Object> maybe_resource, v8::Local<v8::String> name);
-  AsyncResource(MaybeLocal<v8::Object> maybe_resource, const char* name);
+  AsyncResource(v8::Local<v8::Object> resource, v8::Local<v8::String> name);
+  AsyncResource(v8::Local<v8::Object> resource, const char* name);
```
bnoordhuis: Programmer economics: make `resource` the last argument and give it a `= v8::Local<v8::Object>()` default value.

Good idea, bad idea?

ofrobots: Good idea! Thanks. Done.

kkoopa:
Windows builds are currently failing due to a couple (unrelated) things in two new tests from the previous PRs: https://ci.appveyor.com/project/RodVagg/nan/build/job/8vx1q4hhlnetncsi

test\cpp\callbackcontext.cpp(38): error C2466: cannot allocate an array of constant size 0
test\cpp\asyncresource.cpp(41): error C2466: cannot allocate an array of constant size 0

Please pass null instead. You can add the commit here if you like.

ofrobots:
Fixed the windows build.

nan.h
```diff
@@ -1278,26 +1278,29 @@ class Utf8String {
...
+      , v8::Local<v8::Object> resource = New<v8::Object>()) {
 #if NODE_MODULE_VERSION >= NODE_8_0_MODULE_VERSION
       v8::Isolate* isolate = v8::Isolate::GetCurrent();
 
-      v8::Local<v8::Object> resource =
-          maybe_resource.IsEmpty() ? New<v8::Object>()
-                                  : maybe_resource.ToLocalChecked();
+      if (resource.IsEmpty()) {
+        resource = New<v8::Object>();
+      }
```
kkoopa: Is this not a tautology?  If `resource` is empty, it is the result of `New<v8::Object>`. Thus, it need not be reinitialized as such. The same would apply to the other constructor.

ofrobots: An empty handle is not the same as a handle pointing to a newly allocated object.

```c++
Local<v8::Object> handle1;
Local<v8::Object> handle2 = New<v8::Object>();

assert(handle1.IsEmpty() == true);
assert(handle2.IsEmpty() == false);
```

kkoopa:
Ah, nevermind then. All tests seem to pass on all platforms now. I will merge this later today.

PR #734, opened 2018-02-22
Deprecation warnings

kkoopa:
I was about to publish 2.9.0 when I noticed spurious deprecation warnings. This should fix the problem. Since I have already tagged and pushed 2.9.0, I guess the best course of action is to skip publishing 2.9.0 to npm and jump directly to 2.9.1.

### DIFF
--- a/nan_json.h
+++ b/nan_json.h
@@ -134,14 +134,17 @@ class JSON {
 #if NAN_JSON_H_NEED_PARSE
   inline v8::Local<v8::Value> parse(v8::Local<v8::Value> arg) {
     assert(!parse_cb_.IsEmpty() && "parse_cb_ is empty");
-    return parse_cb_.Call(1, &arg);
+    AsyncResource resource("nan:JSON.parse");
+    return parse_cb_.Call(1, &arg, &resource).FromMaybe(v8::Local<v8::Value>());
   }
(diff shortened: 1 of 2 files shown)


nan_json.h
```diff
@@ -134,14 +134,18 @@ class JSON {
...
   }
 #endif  // NAN_JSON_H_NEED_PARSE
 
 #if NAN_JSON_H_NEED_STRINGIFY
   inline v8::Local<v8::Value> stringify(v8::Local<v8::Value> arg) {
     assert(!stringify_cb_.IsEmpty() && "stringify_cb_ is empty");
-    return stringify_cb_.Call(1, &arg);
+    AsyncResource resource("nan:JSON.stringify");
+    Nan::MaybeLocal<v8::Value> result = stringify_cb_.Call(1, &arg, &resource);
+    return result.IsEmpty() ? v8::Local<v8::Value>() : result.ToLocalChecked();
```
bnoordhuis: `return result.FromMaybe(v8::Local<v8::Value>());`?  Same on line 162.

kkoopa: Funny you should mention it. The thought crossed my mind, but I dismissed it as probably not working (for some misremembered reason). I'll fix it and merge.

Any input on not publishing 2.9.0 to npm?

Issue #735, opened 2018-02-22
Error: no viable conversion from 'v8::Local<v8::String>' to 'const char *'


bmathews:
:wave: I'm seeing a couple new electron-rebuild issues with the latest version of nan. Any immediate ideas?

```
An unhandled error occurred inside electron-rebuild
make: Entering directory '/opt/atlassian/pipelines/agent/build/src/node_modules/sharp/build'
  TOUCH Release/obj.target/libvips-cpp.stamp
  CXX(target) Release/obj.target/sharp/src/common.o
In file included from ../src/common.cc:24:0:
../../nan/nan.h: In constructor 'Nan::AsyncResource::AsyncResource(v8::Local<v8::String>, v8::Local<v8::Object>)':
../../nan/nan.h:1290:60: error: cannot convert 'v8::Local<v8::String>' to 'const char*' for argument '3' to 'node::async_context node::EmitAsyncInit__New(v8::Isolate*, v8::Local<v8::Object>, const char*, node::async_id)'
       context = node::EmitAsyncInit(isolate, resource, name);
                                                            ^
../../nan/nan.h: In constructor 'Nan::AsyncResource::AsyncResource(const char*, v8::Local<v8::Object>)':
../../nan/nan.h:1306:67: error: cannot convert 'v8::Local<v8::String>' to 'const char*' for argument '3' to 'node::async_context node::EmitAsyncInit__New(v8::Isolate*, v8::Local<v8::Object>, const char*, node::async_id)'
       context = node::EmitAsyncInit(isolate, resource, name_string);
```

```
An unhandled error occurred inside electron-rebuild
  SOLINK_MODULE(target) Release/.node
  CXX(target) Release/obj.target/fse/fsevents.o
In file included from ../fsevents.cc:6:
../../nan/nan.h:1290:56: error: no viable conversion from 'v8::Local<v8::String>' to 'const char *'
      context = node::EmitAsyncInit(isolate, resource, name);
                                                       ^~~~
/Users/formidable/.electron-gyp/.node-gyp/iojs-1.8.2/src/node.h:578:53: note: passing argument to parameter 'name' here
                                        const char* name,
                                                    ^
In file included from ../fsevents.cc:6:
../../nan/nan.h:1306:56: error: no viable conversion from 'v8::Local<v8::String>' to 'const char *'
      context = node::EmitAsyncInit(isolate, resource, name_string);
                                                       ^~~~~~~~~~~
[...]

PR #738, opened 2018-02-22
Bandaid for async hooks

kkoopa:
Temporary fix disabling async_hooks for Node 8 below 8.6. Having some non-obvious issues with getting the older `async_hooks` API working. Addresses #735.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1280,7 +1280,7 @@ class Utf8String {
     AsyncResource(
         v8::Local<v8::String> name
       , v8::Local<v8::Object> resource = New<v8::Object>()) {
-#if NODE_MODULE_VERSION >= NODE_8_0_MODULE_VERSION
+#if NODE_MODULE_VERSION >= NODE_9_0_MODULE_VERSION
       v8::Isolate* isolate = v8::Isolate::GetCurrent();
 
       if (resource.IsEmpty()) {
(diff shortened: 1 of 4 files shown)


kkoopa:
To follow up on the issue I mentioned: The seemingly obvious fix of using `node::EmitAsyncInit(isolate, resource, name)` with the version of `node::EmitAsyncInit` for 8.5.0 results in this assertion:
`../src/node.cc:1379:v8::MaybeLocal<v8::Value> node::MakeCallback(node::Environment*, v8::Local<v8::Value>, v8::Local<v8::Function>, int, v8::Local<v8::Value>*, node::async_context): Assertion '(env->current_async_id()) == (asyncContext.async_id)' failed.`

Flarna:
Not sure here but I fear this would result in issues if an addon is built with NodeJs >=8.6 but later used with NodeJs < 8.6 which is usually an allowed setup.

There are several projects which distribute pre-compiled binaries and they rely on one binary per major version.

Muserk:
I think it's time to merge this one in!

kkoopa:
> Not sure here but I fear this would result in issues if an addon is built with NodeJs >=8.6 but later used with NodeJs < 8.6 which is usually an allowed setup.
You may be right. I changed it to ignore versions below Node 9.0.

PR #741, opened 2018-02-23
deduplication of DefineOwnProperty() wrapper

mkrufky:
This adds a new file called `nan_define_own_property_helper.h` that contains a deduplicated code block taken from very similar code between `nan_maybe_43_inl.h` and `nan_maybe_pre_43_inl.h `

To anyone looking and scratching their head as to how the code block in  `nan_maybe_43_inl.h` ever gets reached at all, this apparently happens in io.js 3 (see #725)

I wasn't sure what to name the common function, so I went with `DefineOwnPropertyHelper()`, but I'm open to other suggestions...

As discussed: https://github.com/nodejs/nan/pull/728#discussion_r167025257

### DIFF
--- a/nan_define_own_property_helper.h
+++ b/nan_define_own_property_helper.h
@@ -9,6 +9,8 @@
 #ifndef NAN_DEFINE_OWN_PROPERTY_HELPER_H_
 #define NAN_DEFINE_OWN_PROPERTY_HELPER_H_
 
+namespace imp {
+
 inline Maybe<bool> DefineOwnPropertyHelper(
     v8::PropertyAttribute current
   , v8::Handle<v8::Object> obj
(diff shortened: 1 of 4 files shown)


kkoopa:
Will try to take a look during the weekend.

nan_define_own_property_helper.h
```diff
@@ -0,0 +1,25 @@
...
+  , v8::Handle<v8::Object> obj
+  , v8::Handle<v8::String> key
+  , v8::Handle<v8::Value> value
+  , v8::PropertyAttribute attribs = v8::None) {
+  return !(current & v8::DontDelete) ||                     // configurable OR
+                  (!(current & v8::ReadOnly) &&             // writable AND
+                   !((attribs ^ current) & ~v8::ReadOnly))  // same excluding RO
+             ? Just<bool>(obj->ForceSet(key, value, attribs))
+             : Nothing<bool>();
+}
```
bnoordhuis: Suggestion: wrap this in a `namespace imp` so it doesn't leak into the public Nan namespace.

kkoopa:
Sorry, I must have forgotten about this PR when I went through the others on the weekend. LGTM.

PR #742, opened 2018-02-23
Linter cleanups

mkrufky:
* Do not indent within a namespace  [runtime/indentation_namespace] [4]
* No #ifndef header guard found, suggested CPP variable is: TEST_CPP_SLEEP_H_  [build/header_guard] [5]
* Include the directory when naming .h files  [build/include] [4]
* Found C system header after other header. Should be: asyncworker.h, c system, c++ system, other.  [build/include_order] [4]

### DIFF
--- a/test/cpp/sleep.h
+++ b/test/cpp/sleep.h
@@ -8,8 +8,10 @@
 
 #ifndef TEST_CPP_SLEEP_H_
 #define TEST_CPP_SLEEP_H_
+
 #ifndef _WIN32
 #include <unistd.h>
 #define Sleep(x) usleep((x)*1000)
 #endif
+
 #endif  // TEST_CPP_SLEEP_H_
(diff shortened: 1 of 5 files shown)


kkoopa:
That was supposed to be a review only for commit f42a633

test/cpp/sleep.h
```diff
@@ -6,7 +6,10 @@
  * MIT License <https://github.com/nodejs/nan/blob/master/LICENSE.md>
  ********************************************************************/
 
+#ifndef TEST_CPP_SLEEP_H_
+#define TEST_CPP_SLEEP_H_
```
kkoopa: newline after include guard

kkoopa:
But the rest looks okay. I wonder if clang-format could  easily be configured to follow the existing style.

PR #744, opened 2018-02-24
add Nan::Call overload

ofrobots:
While working on adapting node-sass to the new async context preserving mechanism to call, I noticed that it would be handy to have another overload.

This commit adds a non-receiver overload for Nan::Call to make it more comparable to Nan::Callback::Call.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1679,6 +1679,18 @@ inline MaybeLocal<v8::Value> Call(
   return Call(*callback, recv, argc, argv);
 }
 
+inline MaybeLocal<v8::Value> Call(
+    const Nan::Callback& callback
+  , int argc
+  , v8::Local<v8::Value> argv[]) {
+#if NODE_MODULE_VERSION > NODE_0_10_MODULE_VERSION
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  return Call(*callback, isolate->GetCurrentContext()->Global(), argc, argv);
+#else
+  return Call(*callback, v8::Context::GetCurrent()->Global(), argc, argv);
+#endif
+}
+
 /* abstract */ class AsyncWorker {
  public:
   explicit AsyncWorker(Callback *callback_,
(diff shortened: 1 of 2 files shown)


nan.h
```diff
@@ -1679,6 +1679,18 @@ inline MaybeLocal<v8::Value> Call(
   return Call(*callback, recv, argc, argv);
 }
 
+inline MaybeLocal<v8::Value> Call(
+    const Nan::Callback& callback
+  , int argc
+  , v8::Local<v8::Value> argv[]) {
```
kkoopa: I think this needs handle scopes to avoid leaking the context handle and the global handle.

kkoopa:
LGTM with one concern regarding handle scopes

ofrobots:
Added handle scope. I am not too familiar with the pre-0.12 V8 API, so PTAL if this is reasonable.

nan.h
```diff
@@ -1679,6 +1679,25 @@ inline MaybeLocal<v8::Value> Call(
...
+#if NODE_MODULE_VERSION > NODE_0_10_MODULE_VERSION
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  v8::EscapableHandleScope scope(isolate);
+  return scope.Escape(
+      Call(*callback, isolate->GetCurrentContext()->Global(), argc, argv);
+  )
```
kkoopa: The semicolon from the previous line should be moved here.

kkoopa:
The pre-0.12 seems correct, it compiles with passing tests. You used `Nan::EscapableHandleScope`, which abstracts away the handle scope API differences.

nan.h
```diff
@@ -1679,6 +1679,25 @@ inline MaybeLocal<v8::Value> Call(
...
+#else
+  EscapableHandleScope scope;
+  return scope.Escape(Call(*callback,
+                           v8::Context::GetCurrent()->Global(),
+                           argc,
+                           argv).ToLocalChecked());
```
kkoopa: Replace the `ToLocalChecked()` with `FromMaybe(v8::Local<v8::Value>())`

nan.h
ofrobots: Sorry about that. Fixed.

nan.h
ofrobots: Thanks! Done.

ofrobots:
Can I request a release with this? I want to use this from `node-sass`?

kkoopa:
Yes, I will cut a new release as soon as I find time.

PR #747, opened 2018-02-27
Fix deprecation warnings for Node 10

### LINKED ISSUE #746: Deprecated warnings with Node.js v10.0.0-nightly
If [v10.0.0-nightly20180226dbe645f114](https://nodejs.org/download/nightly/v10.0.0-nightly20180226dbe645f114/) is installed and used to build the nan test there are deprecated warnings.

The nan tests were built with the commands:
```
git clone https://github.com/nodejs/nan.git
cd nan/
npm i
npm run-script rebuild-tests
```
The output of the build is [here](https://gist.github.com/fivdi/97cfe9543853b1f5d589f28b68f08a07).

It looks like an `AsyncResource` is needed to avoid some if not all of the warnings.

kkoopa:
Fixes #746

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -460,6 +460,106 @@ class TryCatch {
   }
 };
 
+v8::Local<v8::Value> MakeCallback(v8::Local<v8::Object> target,
+                                  v8::Local<v8::Function> func,
+                                  int argc,
+                                  v8::Local<v8::Value>* argv);
+v8::Local<v8::Value> MakeCallback(v8::Local<v8::Object> target,
+                                  v8::Local<v8::String> symbol,
+                                  int argc,
+                                  v8::Local<v8::Value>* argv);
+v8::Local<v8::Value> MakeCallback(v8::Local<v8::Object> target,
+                                  const char* method,
+                                  int argc,
+                                  v8::Local<v8::Value>* argv);
+
+// === AsyncResource ===========================================================
+
+class AsyncResource {
+ public:
+  AsyncResource(
+      v8::Local<v8::String> name
+    , v8::Local<v8::Object> resource = New<v8::Object>()) {
+#if NODE_MODULE_VERSION >= NODE_9_0_MODULE_VERSION
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+
+    if (resource.IsEmpty()) {
+      resource = New<v8::Object>();
+    }
+
+    context = node::EmitAsyncInit(isolate, resource, name);
+#endif
+  }
+
+  AsyncResource(
+      const char* name
+    , v8::Local<v8::Object> resource = New<v8::Object>()) {
+#if NODE_MODULE_VERSION >= NODE_9_0_MODULE_VERSION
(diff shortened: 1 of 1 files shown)


fivdi:
It compiles without deprecation warnings for me on Node 10 and my [favourite test](https://github.com/fivdi/epoll/blob/master/test/brute-force-leak-check.js) doesn't detect any leaks.

nan.h
```diff
@@ -836,9 +936,15 @@ class TryCatch {
     return scope.Escape(New(node::MakeCallback(
         v8::Isolate::GetCurrent(), target, func, argc, argv)));
 #else
+# if NODE_MODULE_VERSION >= NODE_9_0_MODULE_VERSION
+    AsyncResource res("nan:makeCallback");
+    return res.runInAsyncScope(target, func, argc, argv)
```
ofrobots: I am afraid that this defeats the purpose behind the deprecation of certain `MakeCallbacks` in Node.js 10.

The idea is that the `AsyncResource` should be constructed when a callback is 'stored' for later invocation. That way the async context can be restored at the callback invocation time. Creating the resource here would not restore the correct async context – it would be a no-op.

It would indeed get rid of the deprecation warning in nan itself though.

We perhaps need to deprecate these particular `MakeCallback` APIs in Nan, so that consumer native modules get notified?

ofrobots: IOW, let's silence the node warning, but give warning to `Nan` users if they use these functions by deprecating these `MakeCallback`s.

kkoopa:
While they should be deprecated, deprecating them won't help getting rid of the deprecation warnings from their inclusion in NAN.

ofrobots:
I meant, take your existing patch (which silences the Node warning) and additionally add a deprecation for `Nan::Makecallback`.

In reality it is a fair bit more work as we do use `Nan::Makecallback` internally as well. We'll have to go update all those locations too. If making sure that we don't have any warnings ourselves is important, then this sounds worthwhile.

kkoopa:
Alright, that should be doable. The usual approach is to make the old function internal and add a new (deprecated) facade.

ofrobots:
Wrote https://github.com/nodejs/nan/pull/751 to help this effort.

ofrobots:
Further: #752.

> The usual approach is to make the old function internal and add a new (deprecated) facade.

With the above two PRs in place, this should no longer be necessary. There aren't any internal uses anymore.

kkoopa:
Thanks for the pull requests. Do you think this is good to merge as is now?

nan.h
```diff
@@ -1554,7 +1598,13 @@ class Callback {
      , v8::Local<v8::Value> argv[]) const {
 #if NODE_MODULE_VERSION > NODE_0_10_MODULE_VERSION
     v8::Isolate *isolate = v8::Isolate::GetCurrent();
+# if NODE_MODULE_VERSION >= NODE_9_0_MODULE_VERSION
+    AsyncResource async("nan:Callback:operator()");
```
ofrobots: nit: `nan:Callback:Call` would be more accurate a label; here and below.

PR #751, opened 2018-03-10
add more Nan::Call overloads

ofrobots:
Presently we use MakeCallback throughout the codebase even in cases
where it doesn't make sense (sync calls). Add some more convenience
overloads for Nan::Call to allow such uses to be migrated over to
Nan::Call instead.

In lieu of coverage, change the Tap reverse bindings to use Nan::Call
to perform synchornous calls. This change exposes a timing issue in
one of the tests – fix it by properly calling Tap's t.end().

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1703,10 +1703,10 @@ inline MaybeLocal<v8::Value> Call(
   , int argc
   , v8::Local<v8::Value> argv[]) {
   EscapableHandleScope scope;
-  v8::Local<v8::Value> fn_v = recv->Get(symbol);
+  v8::Local<v8::Value> fn_v =
+      Get(recv, symbol).FromMaybe(v8::Local<v8::Value>());
   if (fn_v.IsEmpty() || !fn_v->IsFunction()) return v8::Local<v8::Value>();
   v8::Local<v8::Function> fn = fn_v.As<v8::Function>();
-
   return scope.Escape(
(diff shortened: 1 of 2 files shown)


nan.h
```diff
@@ -1697,6 +1697,31 @@ inline MaybeLocal<v8::Value> Call(
 #endif
 }
 
+inline MaybeLocal<v8::Value> Call(
+    v8::Local<v8::String> symbol
+  , v8::Local<v8::Object> recv
+  , int argc
+  , v8::Local<v8::Value> argv[]) {
+  EscapableHandleScope scope;
+  v8::Local<v8::Value> fn_v = recv->Get(symbol);
```
kkoopa: Should use the `Nan::Get` wrapper

ofrobots: Done.

kkoopa:
Is not ea4c6226e3fe6036d69dac11c533c02d0541d122 supposed to be in the other PR instead? 
I can merge this by hand once I know if its inclusion here is intended or not.

ofrobots:
Apologies. Got some wires crossed. I've dropped https://github.com/nodejs/nan/commit/ea4c6226e3fe6036d69dac11c533c02d0541d122 from here.

kkoopa:
Great, I have squashed and merged.

ofrobots:
Gentle ping; when you get a chance, it would be good to have these changes in a release.

kkoopa:
Yes, I will try to get a release done by Friday. My plate is too full until then.

PR #752, opened 2018-03-10
Deprecate makecallback

ofrobots:
This builds on top of #751. I will rebase when that lands.

This PR deprecates MakeCallback and removes remaining internal uses.

~~TODO: update the docs.~~


ofrobots:
Rebased and squashed. @kkoopa PTAL.

PR #754, opened 2018-03-12
avoid unneeded call to Isolate::GetCurrent() in converters

Flarna:
I have seen an unneeded call to Isolate::GetCurrent() in converters during stepping around. Even the overhead of this is small it's unneeded.


kkoopa:
Thank you. Seems nobody has noticed this in over a year.

PR #757, opened 2018-03-21
fix: AsyncResource needs a HandleScope

ofrobots:


ofrobots:
Actually, this isn't quite right. Let me iterate on this a bit.

ofrobots:
I don't think this is needed and/or necessary. The default parameter needs a handle scope in the caller anyway, so it would be expect a scope from callers.

kkoopa:
I agree.

PR #760, opened 2018-04-16
Fix typo

rivertam:


rivertam:
Np! While I'm here... do you know any resources that can actually help me use AsyncResource? I can see I can just make an `AsyncResource("some name")` and use that, but I'm confused about what the name actually means. I'm sure I'm missing some obvious documentation.

mathiask88:
https://github.com/nodejs/nan/blob/master/doc/node_misc.md#api_nan_asyncresource describes the name purpose.

PR #761, opened 2018-04-22
fix example deprecation

DanielRamosAcosta:
When compiling `info.GetReturnValue().Set(cons->NewInstance(argc, argv));` it gives the warning `'v8::Function::NewInstance': was declared deprecated`. This PR fixes it.


kkoopa:
Thanks.

PR #778, opened 2018-05-25
use Buffer.from('text') rather than new Buffer('text') when available

mkrufky:
fix (node:11826) [DEP0005] DeprecationWarning:
Buffer() is deprecated due to security and usability issues.
Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

Also, includes a check for whether Buffer.from is actually available :-)

### DIFF
--- a/test/js/strings-test.js
+++ b/test/js/strings-test.js
@@ -23,10 +23,13 @@ test('FromV8String', function (t) {
   t.equal(a('an utf8 strïng'), 'an utf8 strïng');
   t.equal(b('an utf8 strïng'), 'an utf8 strïng');
 
+  var buf;
+
   if (typeof(Buffer.from) === "function") {
-    t.equal(bindings.encodeHex(), Buffer.from('hello').toString('hex'));
+    buf = Buffer.from('hello');
   } else {
-    t.equal(bindings.encodeHex(), new Buffer('hello').toString('hex'));
+    buf = new Buffer('hello');
   }
+  t.equal(bindings.encodeHex(), buf.toString('hex'));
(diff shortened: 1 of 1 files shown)


test/js/strings-test.js
```diff
@@ -23,6 +23,13 @@ test('FromV8String', function (t) {
   t.equal(a('an utf8 strïng'), 'an utf8 strïng');
   t.equal(b('an utf8 strïng'), 'an utf8 strïng');
 
-  t.equal(bindings.encodeHex(), new Buffer('hello').toString('hex'));
+  var buf;
+
+  if (typeof(Buffer.from) === "function") {
```
Flarna: In old Node versions (at least some 4.x) this will be true because the base class `Uint8Array` exposes `from` which is not the same as `Buffer.from` in actual Node versions.
But you can check on `Buffer.alloc` instead.

mkrufky: Thanks for that.   I was just scratching my head because there was an unexpected Travis CI failure against node 0.10 on macos, only.

mkrufky:
node 10 is failing due to broken named interceptor test (unrelated to this PR)

kkoopa:
The test is probably broken for the same reason as this: 2b45552c97490a2d95cc2a303a799d9e609de9ab

mkrufky:
I tried to fix that (in a separate branch) but ran into other issues.  I'll push it so we can all take a look.  Meanwhile, let's approve & merge this one.

kkoopa:
Sure, squash and merge.

PR #779, opened 2018-05-25
test/cpp: replace `v8::String::Utf8Value` with `Nan::Utf8String`

mkrufky:
fix warning: ‘v8::String::Utf8Value::Utf8Value(v8::Local<v8::Value>)’ is deprecated: Use Isolate version [-Wdeprecated-declarations]


kkoopa:
LGTM

PR #780, opened 2018-05-26
appveyor.yml: allow failures against nodejs 0.8

mkrufky:
...at least until we figure out how to get past the SSL error

This will allow builds to go green if the only failure is `Environment: nodejs_version=0.8`

note:  this build still fails against nodejs version 10 due to the broken named interceptor test

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,10 @@ environment:
     - nodejs_version: "9"
     - nodejs_version: "10"
 
+matrix:
+  allow_failures:
+    - nodejs_version: "0.8"
+
 # Install scripts. (runs after repo cloning)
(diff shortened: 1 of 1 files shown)


kkoopa:
Probably good. Waiting for the results from the AppVeyor run.

mkrufky:
@kkoopa You won't be able to see the desired effect from the main appveyor run, due to the failure against nodejs version 10 due to the broken named interceptor test, but in my own repository I tested with allowed failures for both 0.8 and 10 and the build is green:

https://ci.appveyor.com/project/mkrufky/nan/build/184

kkoopa:
Good to merge then.

PR #781, opened 2018-05-26
[HELP WANTED] Fix bugs in named interceptor test.

mkrufky:
Copied, more or less, from commit 946377f194e3aaf2aeb5141fa38768f5ca56734a

Returning a string in the enumerator was never allowed but wasn't
enforced until recently.  In Node.js 10, it hits a CHECK in V8.

This builds, but the tests fail:
```
#
# Fatal error in , line 0
# Check failed: element->IsName().
#
#
#
#FailureMessage Object: 0x7ffd987f3690not ok test/js/namedinterceptors-test.js ................ 0/1
    Command: "/home/mk/.nvm/versions/node/v10.2.1/bin/node --expose-gc namedinterceptors-test.js"
    TAP version 13
    not ok 1 test/js/namedinterceptors-test.js
      ---
        exit:    ~
        signal:  SIGILL
        stderr:  |
          #
          # Fatal error in , line 0
          # Check failed: element->IsName().
          #
          #
          #
          #FailureMessage Object: 0x7ffd987f3690
        command: "/home/mk/.nvm/versions/node/v10.2.1/bin/node --expose-gc namedinterceptors-test.js"
      ...
    
    1..1
    # tests 1
    # fail  1
    
```

I'm not sure what else is needed but this is at least a starting point for fixing the Node 10 CI failure.

### DIFF
--- a/test/cpp/namedinterceptors.cpp
+++ b/test/cpp/namedinterceptors.cpp
@@ -96,7 +96,7 @@ NAN_PROPERTY_SETTER(NamedInterceptor::PropertySetter) {
 
 NAN_PROPERTY_ENUMERATOR(NamedInterceptor::PropertyEnumerator) {
   v8::Local<v8::Array> arr = Nan::New<v8::Array>();
-  Set(arr, 0, Nan::New("value").ToLocalChecked());
+  Set(arr, Nan::New("zero").ToLocalChecked(), Nan::New("value").ToLocalChecked());
   info.GetReturnValue().Set(arr);
 }
 
--- a/test/cpp/namedinterceptors.cpp
+++ b/test/cpp/namedinterceptors.cpp
@@ -96,7 +96,7 @@ NAN_PROPERTY_SETTER(NamedInterceptor::PropertySetter) {
 
 NAN_PROPERTY_ENUMERATOR(NamedInterceptor::PropertyEnumerator) {
   v8::Local<v8::Array> arr = Nan::New<v8::Array>();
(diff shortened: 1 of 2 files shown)


kkoopa:
Seems like this was not the right fix, the array elements must be strings (or symbols) https://github.com/nodejs/node/blob/master/deps/v8/include/v8.h#L5134

mkrufky:
Well, it wasn't as simple as 946377f194e3aaf2aeb5141fa38768f5ca56734a :-P

I tried, but I'm closing this.

bnoordhuis:
I suppose it must be some recent Node.js change because it wasn't failing for me two weeks ago when I fixed the other test.  I'll have a look.

bnoordhuis:
#782.  Note it didn't hit a CHECK for me but the `Object.keys(interceptor)[0] === 'value'` check was failing.

kkoopa:
> I suppose it must be some recent Node.js change because it wasn't failing for me two weeks ago when I fixed the other test. I'll have a look.

Yes, I remember everything passing after the indexed interceptor test fix, which confused me now. Semver, schmemver... I am happy it is sorted out now.

bnoordhuis:
I'm reasonably sure it's caused by https://github.com/nodejs/node/pull/20350, which I would say is a legitimate bug fix (but I signed off on it so of course I would say that :-))

mkrufky:
I don't think it was caused by a new regression - the named interceptor test was failing the moment we added node 10 to the matrix:
https://travis-ci.org/nodejs/nan/builds/377939805

...but it's passing now.  Thanks :-)

PR #782, opened 2018-05-26
Fix bugs in named interceptor test.

bnoordhuis:
The PropertyQuery callback needs to return _something_ in order for V8
to consider the property present in Node.js 10.

Fixes failing test test/js/namedinterceptors-test.js


kkoopa:
Thank you. Merged.

PR #783, opened 2018-05-27
 Fix SSL Error: CERT_UNTRUSTED on Appveyor against nodejs version 0.8 

mkrufky:
* Fix SSL Error: CERT_UNTRUSTED on Appveyor against nodejs version 0.8
* Revert "appveyor.yml: allow failures against nodejs 0.8 (#780)"

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ matrix:
 install:
   # Get the latest stable version of Node 0.STABLE.latest
   - ps: Install-Product node $env:nodejs_version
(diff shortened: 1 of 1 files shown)


kkoopa:
Looks good, but I wonder why the original string comparison did not succeed.

mkrufky:
Maybe `if %nodejs_version% == "0.8"` would have worked (note the lowercase `if`), but I saw other projects doing `IF %nodejs_version% EQU 0.8` and that did work, and looks more uniform, so I went with it.

PR #784, opened 2018-05-27
Fix typedarrays.cpp: Missing spaces around < [whitespace/operators] [3] 

mkrufky:
* Add `test/cpp/typedarrays.cpp` to `Makefile` `LINT_SOURCES`
* typedarrays.cpp: Missing spaces around < [whitespace/operators] [3]


kkoopa:
Good catch, that has been unlinted for years.

PR #786, opened 2018-06-10
Update dependency version

realdennis:
Reduce vulnerability, also update tap version and test script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url": "git://github.com/nodejs/nan.git"
   },
   "scripts": {
-    "test": "tap --gc test/js/*-test.js",
+    "test": "tap --gc --stderr test/js/*-test.js",
     "rebuild-tests": "node-gyp rebuild --msvs_version=2015 --directory test",
     "docs": "doc/.build.sh"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,10 +26,10 @@
     "bindings": "~1.2.1",
     "commander": "^2.8.1",
     "glob": "^5.0.14",
-    "request": "=2.81.0",
-    "node-gyp": "~3.6.2",
+    "node-gyp": "^3.7.0",
     "readable-stream": "^2.1.4",
-    "tap": "~0.7.1",
+    "request": "^2.87.0",
+    "tap": "^12.0.1",
```
mkrufky: I believe we stuck to `"tap": "~0.7.1"` for compatibility with builds against older versions of Node.js 

Either that, or nobody tried to update it until now.  CI failed :-(

Some of the test failures have nothing to do with the Node.js version, though, which makes things interesting...

I'd love to dig in and see why those tests are failing, but it's getting late and I have to be at the office in the morning.  I'll take a closer look in a few days if somebody else doesn't beat me to it.

Either way, I'll have to nack the PR for now as it breaks CI.

realdennis:
Sorry for waste your time...my mistake. 
But, I thought other dependency need to update if CI build pass, let dependency update progressively..

PR #787, opened 2018-06-11
Update dependency version except tap

realdennis:
As title,
By the way, I thought that `tap` version is too old to update to latest version for reducing vulnerability, but if we can find a version for lowest node version and less vulnerability, it may be a good trade-off?

and CI build pass. https://travis-ci.org/realdennis/nan

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "test": "tap --gc --stderr test/js/*-test.js",
+    "rebuild-tests": "node-gyp rebuild --msvs_version=2015 --directory test",
     "docs": "doc/.build.sh"
   },
   "contributors": [
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   "devDependencies": {
     "bindings": "~1.2.1",
     "node-gyp": "^3.7.0",
+    "readable-stream": "^2.3.6",
     "tap": "^0.7.1"
   },
   "license": "MIT"
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url": "git://github.com/nodejs/nan.git"
   },
   "scripts": {
-    "test": "tap --gc test/js/*-test.js",
+    "test": "tap --gc --stderr test/js/*-test.js",
     "rebuild-tests": "node-gyp rebuild --msvs_version=2015 --directory test",
     "docs": "doc/.build.sh"
   },
@@ -29,7 +29,7 @@
     "node-gyp": "^3.7.0",
     "readable-stream": "^2.1.4",
     "request": "^2.87.0",
-    "tap": "^12.0.1",
+    "tap": "^0.7.1",
     "xtend": "~4.0.0"
   },
   "license": "MIT"
--- a/package.json
+++ b/package.json
@@ -24,13 +24,8 @@
   ],
   "devDependencies": {
     "bindings": "~1.2.1",
-    "commander": "^2.8.1",
-    "glob": "^5.0.14",
     "node-gyp": "^3.7.0",
-    "readable-stream": "^2.1.4",
-    "request": "^2.87.0",
-    "tap": "^0.7.1",
-    "xtend": "~4.0.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,10 +26,10 @@
     "bindings": "~1.2.1",
     "commander": "^2.8.1",
     "glob": "^5.0.14",
-    "request": "=2.81.0",
-    "node-gyp": "~3.6.2",
+    "node-gyp": "^3.7.0",
     "readable-stream": "^2.1.4",
-    "tap": "~0.7.1",
+    "request": "^2.87.0",
```
bnoordhuis: FWIW, I have no idea why we depend on `request`.  We don't use it anywhere.

node-gyp is another dependency that should probably go.  I recommend (I also maintain node-gyp) to use the version of node-gyp that comes with node/npm, not install your own.

kkoopa:
Tests seem to pass, but please try the things Ben pointed out.

mkrufky:
I'm still interested in the way that some of the tests failed when you upgraded `tap` to the latest in your previous PR:  https://github.com/nodejs/nan/pull/786 Definitely worth looking at the CI failures there - some may reveal some actual issues in the code, although I still haven't looked any closer at them - will dig deeper in a few days.

I don't know if we really need to stick with `"tap": "~0.7.1"` - Is there a newer release that doesn't break builds against older Node.js versions?  ...or is ``"~0.7.1"` the latest that works without issue?

kkoopa:
It is the last that works, as far as I can remember.

package.json
realdennis: Hey I found we have some redundant dependency...
"node-gyp is another dependency that should probably go ... " -> I do not know what you meaning about ?

realdennis:
@bnoordhuis Hey, below I list that some dependency I found  we do not use.
`request` `commander` `glob` `xtend`, and  `readable-stream` should be using by node v0.8.

By the way, should I need to pull a new request, or just edit the title in this situation?

realdennis:
@mkrufky 
Except the oldest node version could not use newest tap (const declare in tap)
(I don't know how to using the higher version of tap but taking care of the old version?)

Build failed(v4-v10) cause some test case did not pass, it maybe reveal some actual issues .

kkoopa:
If I remember correctly, the problem with TAP is that later versions throw a hissy fit for anything that is logged to stderr (or was it stdout).

package.json
```diff
@@ -24,13 +24,9 @@
...
-    "commander": "^2.8.1",
-    "glob": "^5.0.14",
-    "request": "=2.81.0",
-    "node-gyp": "~3.6.2",
-    "readable-stream": "^2.1.4",
-    "tap": "~0.7.1",
-    "xtend": "~4.0.0"
+    "node-gyp": "^3.7.0",
+    "readable-stream": "^2.3.6",
+    "tap": "^0.7.1"
```
mkrufky: did you intentionally change this to `^0.7.1` or did you mean to leave it unchanged at `~0.7.1`?

realdennis: Not intentionally, sorry. 
I mean to leave it unchanged at ~0.7.1 (before we find that root cause about CI result)
I'll change it back as soon.

realdennis:
@mkrufky Refer to [npm's bundle node gyp ](https://github.com/nodejs/node-gyp/wiki/Updating-npm%27s-bundled-node-gyp), I remove the node-gyp devDependency, and rewrote .travis.yml(add `npm install -g node-gyp` in before_install, using npm script instead of older version).
Please ignore eb503bb commit, I add it back in next commit.

mkrufky:
I've been away, and this PR got a bit more complicated than its original intent.   How about we go back, for now, to whatever it takes for us not to have to modify `.travis.yml`   If that was a result of removing the `node-gyp` then let's not remove it just yet.  Let's keep this change simple and handle the rest in a separate PR.

bnoordhuis:
@realdennis This PR has been in limbo for some time now. Do you want to follow up or should we close it? Cheers.

PR #789, opened 2018-06-14
remove FunctionCallbackInfo::Callee for nodejs > 10

### LINKED ISSUE #788: error: FunctionCallbackInfo has no member Callee
Using nan@2.10, node@10.2.1
```
/home/romgrk/.node-gyp/10.2.1/include/node/v8.h: In instantiation of ‘void v8::PersistentBase<T>::SetWeak(P*, typename v8::WeakCallbackInfo<P>::Callback, v8::WeakCallbackType) [with P = Nan::ObjectWrap; T = v8::Object; typename v8::WeakCallbackInfo<P>::Callback = void (*)(const v8::WeakCallbackInfo<Nan::ObjectWrap>&)]’:
../node_modules/nan/nan_object_wrap.h:66:61:   required from here
/home/romgrk/.node-gyp/10.2.1/include/node/v8.h:9238:16: warning: cast between incompatible function types from ‘v8::WeakCallbackInfo<Nan::ObjectWrap>::Callback’ {aka ‘void (*)(const v8::WeakCallbackInfo<Nan::ObjectWrap>&)’} to ‘Callback’ {aka ‘void (*)(const v8::WeakCallbackInfo<void>&)’} [-Wcast-function-type]
In file included from ../node_modules/nan/nan_callbacks.h:83,
                 from ../node_modules/nan/nan.h:153,
                 from ../src/boxed.h:4,
                 from ../src/function.cc:2:
../node_modules/nan/nan_callbacks_12_inl.h: In instantiation of ‘v8::Local<v8::Function> Nan::FunctionCallbackInfo<T>::Callee() const [with T = v8::Value]’:
../src/function.cc:71:27:   required from here
../node_modules/nan/nan_callbacks_12_inl.h:108:64: error: ‘const class v8::FunctionCallbackInfo<v8::Value>’ has no member named ‘Callee’
   inline v8::Local<v8::Function> Callee() const { return info_.Callee(); }
                                                          ~~~~~~^~~~~~
```
https://github.com/nodejs/nan/blob/v2.10.0/nan_callbacks_12_inl.h#L108
https://github.com/nodejs/node/blob/v10.2.1/deps/v8/include/v8.h#L3666-L3721

romgrk:
Fixes #788 

Removed from V8 in https://github.com/v8/v8/commit/bc8c97c72a4f5a72015f06ea60cc78e23bd5d300
Which maps to nodejs 10.0.0 ([table](https://nodejs.org/en/download/releases/))


bnoordhuis:
I'd just add a note to the docs.  You don't even to touch any code because thanks to the magic of templates it only fails to compile when you use the function but not otherwise.

romgrk:
Really? I had no idea templates were able to do that. 
So should I delete the commit that touches the code? I don't mind doing so, but it kinda adds a nice self-documenting-code touch. Also it makes the error more user-friendly, because if the user inadvertently uses `info.Callee()` it doesn't point to an error in nan code.

bnoordhuis:
> So should I delete the commit that touches the code?

No strong opinion.  I'll let one of the other maintainers make the call.

kkoopa:
While I would prefer reimplementing the sought behavior, I do not currently have time to do it.  I think the best option with least effort is to deprecate `Callee` regardless of Node version and having it conditionally removed for Node 10+ like now, to reduce the number of spurious bug reports that come in.

I will see if I can find some time later this week to look into reimplementing it by inspecting the call stack in addition to a deprecation, since I think consistency of observable behavior is important.

bnoordhuis:
@kkoopa Keep in mind that needs V8 changes first (method to get the v8::Function for a stack frame) so the first 10.x releases will remain broken, whatever happens next.

kkoopa:
Thanks for pointing that out. I missed that part while skimming the related thread.
In that case, the previously mentioned way is probably the best.

PR #791, opened 2018-06-23
Introduce a GetCurrentEventLoop function to get the current uv event loop

rpetrich:
This is necessary now that node is supporting multiple v8 isolates and uv event loops behind the `--experimental-worker` flag in node v10.5

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -561,9 +561,9 @@ class AsyncResource {
 };
 
 inline uv_loop_t* GetCurrentEventLoop() {
-#if NODE_MAJOR_VERSION >= 10 \
-  || NODE_MAJOR_VERSION == 9 && NODE_MINOR_VERSION >= 3 \
-  || NODE_MAJOR_VERSION == 8 && NODE_MINOR_VERSION >= 10
+#if NODE_MAJOR_VERSION >= 10 || \
+  NODE_MAJOR_VERSION == 9 && NODE_MINOR_VERSION >= 3 || \
+  NODE_MAJOR_VERSION == 8 && NODE_MINOR_VERSION >= 10
     return node::GetCurrentEventLoop(v8::Isolate::GetCurrent());
 #else
     return uv_default_loop();
--- a/nan.h
+++ b/nan.h
@@ -561,7 +561,7 @@ class AsyncResource {
 };
 
 inline uv_loop_t* GetCurrentEventLoop() {
-#if NODE_MODULE_VERSION > NODE_9_0_MODULE_VERSION
+#if NODE_MAJOR_VERSION >= 10 || (NODE_MAJOR_VERSION == 9 && NODE_MINOR_VERSION >= 3) || (NODE_MAJOR_VERSION == 8 && NODE_MINOR_VERSION >= 10)
     return node::GetCurrentEventLoop(v8::Isolate::GetCurrent());
 #else
     return uv_default_loop();
--- a/nan.h
+++ b/nan.h
@@ -560,6 +560,14 @@ class AsyncResource {
 #endif
 };
 
+inline uv_loop_t* GetCurrentEventLoop() {
+#if NODE_MODULE_VERSION > NODE_9_0_MODULE_VERSION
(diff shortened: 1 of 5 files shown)


nan.h
```diff
@@ -560,6 +560,14 @@ class AsyncResource {
 #endif
 };
 
+inline uv_loop_t* GetCurrentEventLoop() {
+#if NODE_MODULE_VERSION > NODE_9_0_MODULE_VERSION
```
bnoordhuis: This needs to take the minor version into account.  `node::GetCurrentEventLoop()` was introduced in v9.3.0.

test/tap-as-worker.js
```diff
@@ -0,0 +1,12 @@
...
+const worker_threads = require("worker_threads");
+const path = require("path");
+
+if (worker_threads.isMainThread) {
+	const worker = new worker_threads.Worker(__filename, { workerData: process.argv });
+	worker.on("error", console.error);
+} else {
+	process.argv = worker_threads.workerData;
+	require(path.resolve(require.resolve("tap"), "../../bin/tap.js"));
+}
```
bnoordhuis: Can you use single quotes and fix the indentation in this file?  Should be two spaces.

nan.h
rpetrich: I've updated the check to allow versions >= 10, 9 with minor versions >= 3, and 8 with minor versions >= 10. Seems node v6.14.2 also supports `GetCurrentEventLoop` as part of the node_api_backport.h header, but it may not be worth the effort to support it given node 6's age.

Flarna: I think this would result in addons built with e.g. 8.10 don't load with Node 8.9 which is usually not expected.

rpetrich: @Flarna hmm... perhaps returning to the original `NODE_MODULE_VERSION > NODE_9_0_MODULE_VERSION` is best then? (which would use `node::GetCurrentEventLoop ` only on node 10)

bnoordhuis: >  think this would result in addons built with e.g. 8.10 don't load with Node 8.9 which is usually not expected.

That's not nan's problem to solve.  Nan provides build-time compatibility, not load-time compatibility.  If we can get both that's nice but it's not a requirement.

test/tap-as-worker.js
```diff
@@ -0,0 +1,12 @@
+// Run the tap tests in a worker
+// Not a clean approach, but is sufficient to verify correctness
+const worker_threads = require('worker_threads');
+const path = require('path');
+
+if (worker_threads.isMainThread) {
+  const worker = new worker_threads.Worker(__filename, { workerData: process.argv });
```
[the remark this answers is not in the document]
rpetrich: Wrapped.

nan.h
```diff
@@ -560,6 +560,14 @@ class AsyncResource {
 #endif
 };
 
+inline uv_loop_t* GetCurrentEventLoop() {
+#if NODE_MAJOR_VERSION >= 10 || (NODE_MAJOR_VERSION == 9 && NODE_MINOR_VERSION >= 3) || (NODE_MAJOR_VERSION == 8 && NODE_MINOR_VERSION >= 10)
```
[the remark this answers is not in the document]
rpetrich: Fixed.

test/tap-as-worker.js
```diff
@@ -0,0 +1,12 @@
...
+  worker.on('error', console.error);
```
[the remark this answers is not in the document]
rpetrich: I've updated to handle the error properly by letting the worker print the error, but exiting the process with non-zero.

nan.h
```diff
@@ -560,6 +560,16 @@ class AsyncResource {
 #endif
 };
 
+inline uv_loop_t* GetCurrentEventLoop() {
+#if NODE_MAJOR_VERSION >= 10 \
+  || NODE_MAJOR_VERSION == 9 && NODE_MINOR_VERSION >= 3 \
+  || NODE_MAJOR_VERSION == 8 && NODE_MINOR_VERSION >= 10
```
bnoordhuis: Can you put the `||` operator on the previous line?  That's more in line with other #if statements in this file.

PR #794, opened 2018-07-08
Fix AsyncProgressWorkerBase::SendProgress_ sends invalid data

tossy310:
Current ProgressWorker implementation can send invalid progress data (described below) in `AsyncProgressWorkerBase::HandleProgressCallback`.

This PR fix it by extending the range to apply mutex.

The issue can occur in the following case:

1. ProgressWorker requests to [SendProgress_](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L2038) and the event is registered to the uv queue by [uv_async_send](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L2052) (but not awaken now)

2. ProgressWorker requests to send progress again, but just after called [mutex_unlock](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L2049) (this is before calling [uv_async_send](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L2052)), the running thread switches to another one, and [HandleProgressCallback](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L2032) is called.
In this case, HandleProgressCallback sends `data` and `count` which were updated in the 2nd call of `SendProgress`. And then, `asyncdata_` set to null and `asyncsize_` is kept old value.

3. Running thread back to the `ProgressWorker::SendProgress_`, and [uv_async_send](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L2052) is called (though the corresponding data is already callbacked.)

4. [HandleProgressCallback](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L2032) is called again, but `data` is null and `count` is old value.
This behavior seems invalid.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -2046,10 +2046,10 @@ class AsyncProgressWorkerBase : public AsyncBareProgressWorker<T> {
     T *old_data = asyncdata_;
     asyncdata_ = new_data;
     asyncsize_ = count;
+    uv_async_send(&this->async);
     uv_mutex_unlock(&async_lock);
 
     delete[] old_data;
-    uv_async_send(&this->async);
   }
 
   uv_mutex_t async_lock;


kkoopa:
Thank you. I requested Ben to take a look, since he knows libuv well (to say the least). It might take some weeks due to vacations and nice weather.

mkrufky:
This is interesting.  I'm just curious, are you able to reproduce this same bad behavior if you use an `AsyncProgressQueueWorker<T>` instead of an `AsyncProgressWorker<T>`?

Assuming yes, this patch might also need to include something along the lines of:
```patch
diff --git a/nan.h b/nan.h
index 29f084e..af7ec2c 100644
--- a/nan.h
+++ b/nan.h
@@ -2139,9 +2139,8 @@ class AsyncProgressQueueWorker : public AsyncBareProgressQueueWorker<T> {
 
     uv_mutex_lock(&async_lock);
     asyncdata_.push(std::pair<T*, size_t>(new_data, count));
-    uv_mutex_unlock(&async_lock);
-
     uv_async_send(&this->async);
+    uv_mutex_unlock(&async_lock);
   }
 
   uv_mutex_t async_lock;
```

Flarna:
In my opinion the critical section is correct but `WorkProgress()` should be adapted to call `HandleProgressCallback()` only if `data != nullptr` which happens here.

To my understanding both can happen:

- several calls to `uv_async_send()` before `WorkProgress()` gets invoked - intermediate data is lost which is fine if just the actual state is needed
- `WorkProgress()` may be invoked but data is already consumed by previous call.

`AsyncProgressQueueWorker` is already prepared for this as it has a loop in `WorkProgress()` - if it get's invoked with all `asyncdata_` elements already consumed it simply does nothing.

tossy310:
Thank you for your comments.

@mkrufky 

> This is interesting. I'm just curious, are you able to reproduce this same bad behavior if you use an AsyncProgressQueueWorker<T> instead of an AsyncProgressWorker<T>?
> 
> Assuming yes, this patch might also need to include something along the lines of:
> 

The answer is no. (maybe a part of this comment overlaps with one from Flarna)
In `AsyncProgressQueueWorker<T>`, whole of the function `WorkProgress()` is wrapped by uv_mutex. `HandleProgressCallback()` is called only when there is unsent data in its queue.

In my opition, the critical section in **Queue** Worker is already safe.
So, I didn't include it in this PR.


@Flarna 

> In my opinion the critical section is correct but WorkProgress() should be adapted to call HandleProgressCallback() only if data != nullptr which happens here.

Actually, my first attempt of this PR was ,as you pointed out, including nullptr check befor calling `HandleProgressCallback()`.
But there is a [Signal](https://github.com/nodejs/nan/blob/b058fb047d18a58250e66ae831444441c1f2ac7a/nan.h#L1979) functionarity that invokes callback without sending any data, in which case `HandleProgressCallback()` is called with the data == nullptr.

More specifically, if the nullptr check is included, the test [asyncprogressworkersignal](https://github.com/nodejs/nan/blob/master/test/js/asyncprogressworkersignal-test.js) fails.

Flarna:
Right, `Signal()` is broken then... haven't seen this. Not sure about the semantics of `Signal()`; in `AsyncProgressQueueWorker` it's not included for some reason so this one can't be used as drop in replacement.

I think we should at least clear `asyncsize_` along with `asyncdata_`.

Even if the critical section in `SendProgress_()` is changed the same race is still there between calls to `Signal()` and `Send()` so it would be needed to use the mutex also in `Signal()`.

tossy310:
Thanks. Updated.

PR #801, opened 2018-08-21
Added missing multifile2.h to solve Makefile error

stefanos1316:
A missing header file was giving an error upon make command, I solved it by including it.

Cheers,

Stefanos

### DIFF
--- a/test/cpp/multifile2.cpp
+++ b/test/cpp/multifile2.cpp
@@ -7,6 +7,7 @@
  ********************************************************************/
 
 #include <nan.h>
+#include <multifile2.h>
 
 using namespace Nan;  // NOLINT(build/namespaces)
 


test/cpp/multifile2.cpp
```diff
@@ -7,6 +7,7 @@
  ********************************************************************/
 
 #include <nan.h>
+#include <multifile2.h>
```
mkrufky: It breaks every build on every version of every os :-/
```
../cpp/multifile2.cpp:10:24: fatal error: multifile2.h: No such file or directory
 #include <multifile2.h>
                        ^
```

mkrufky: should be fine if you change it to
```#include "multifile2.h"```
(change the `<` & `>` to `"`)

stefanos1316:
On my Fedora system, it worked, but I was only running the make command. 
Either way, I applied the requested change now and Travis passed all the test cases. :)

stefanos1316:
You welcome!

PR #808, opened 2018-09-28
deps: adapt to V8 7.0

refack:
Fixes: https://github.com/nodejs/node/issues/23122

~Also updates dependencies~

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1060,7 +1060,11 @@ class Utf8String {
       length_(0), str_(str_st_) {
     HandleScope scope;
     if (!from.IsEmpty()) {
+#if V8_MAJOR_VERSION >= 7
+      v8::Local<v8::String> string = from->ToString(v8::Isolate::GetCurrent());
+#else
       v8::Local<v8::String> string = from->ToString();
+#endif
       if (!string.IsEmpty()) {
         size_t len = 3 * string->Length() + 1;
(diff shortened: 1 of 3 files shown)


addaleax:
@rvagg @kkoopa Can we get a release out with this asap? I think that would be a really good step forward for un-breaking CITGM…

chjj:
Still receiving deprecation warnings with 2.11.1:

```
../../nan/nan.h:1064:78: warning: v8::Local<v8::String> v8::Value::ToString(v8::Isolate*) const is deprecated: Use maybe version [-Wdeprecated-declarations]
       v8::Local<v8::String> string = from->ToString(v8::Isolate::GetCurrent());
```

It appears that:

``` cc
v8::Local<v8::String> v8::Value::ToString(v8::Isolate*) const
``` 

is also deprecated.

Shouldn't it be doing something like:

``` cc
v8::Local<v8::String> string = from->ToString(v8::Isolate::GetCurrent()->GetCurrentContext()).FromMaybe(v8::Local<v8::String>());
```

Not sure.

cc @addaleax

Flarna:
@chjj I think #811 addresses these warnings

chjj:
@Flarna, ah, thanks. Didn't see that.

All this v8 stuff is driving me nuts. I should really switch to NAPI.

PR #816, opened 2018-10-19
[doc] fix links and a few copy paste errors

Flarna:
Corrected links to v8 docs (used 8.11 where possible) as I got a 404 for the io-js-3.3 links.

Corrected a few copy/paste errors in docs.


Flarna:
Hmm, seems request@2.81.0 pulls in sshpk@1.15.1 now (was 1.14.2 a week or two before) which causes Node 0.8 builds to fail.

kkoopa:
Seems we need to pin some dependency. Thanks.

PR #818, opened 2018-10-22
add nan-*.tgz to .npmignore

Flarna:
The 2.11.1 package at NPM includes a file nan-2.11.1.tgz - most likely npm pack was executed before npm publish.

I added nan-*.tgz to .npmignore to avoid this in future.


kkoopa:
My bad.

Issue #810, opened 2018-10-11
New deprecation warnings in node 10.12


jkrems:
Some new deprecations have been introduced that may be worth addressing:

* `StringObject::New`
* non-Maybe `ToString`
* `WriteUtf8`

nolanmar511:
So happy the deprecation warnings have been fixed!
Could a version nan with these fixes be released?

kkoopa:
Sure, I shall get to it this week.

PR #825, opened 2018-11-13
Fix Node.js v10.12.0 deprecation warnings.

bnoordhuis:
v10.12.0 turns on a number of V8 deprecation warnings. This commit fixes
them in NAN.

Fixes: #810
Refs: #811

### DIFF
--- a/nan_implementation_12_inl.h
+++ b/nan_implementation_12_inl.h
@@ -100,9 +100,17 @@ Factory<v8::Function>::New( FunctionCallback callback
     obj->SetInternalField(imp::kDataIndex, val);
   }
 
-  return scope.Escape(v8::Function::New( isolate
-                          , imp::FunctionCallbackWrapper
-                          , obj));
+#if NODE_MAJOR_VERSION >= 10
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::Function> function =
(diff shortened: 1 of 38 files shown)


nan_implementation_12_inl.h
```diff
@@ -337,7 +345,23 @@ Factory<v8::StringObject>::New(v8::Local<v8::String> value) {
 #if V8_MAJOR_VERSION >= 7
```
Flarna: At the other locations this `#if` was replaced by `#if NODE_MAJOR_VERSION >= 10`. Is it kept intentionally here?

bnoordhuis: Yes. Should I add a comment?

Flarna: Up to you.
Personally I find the #ifs on V8 version not nice as it indicates that it breaks the API/ABI stability within a major nodejs version.

bnoordhuis:
Rebased onto master. The CI should be passing now that #826 is merged.

nan_implementation_12_inl.h
bnoordhuis: Yeah, I agree. I've pushed a new commit that removes it, PTAL.

bnoordhuis:
@agnat @kkoopa Can I persuade one of you to take a look? I'll trade you a review in exchange!

kkoopa:
Nice job! LGTM.

Issue #829, opened 2018-12-15
Support for Node 12/master


mcollina:
It seems this module has not been updated to support Node 11, and I've a fear it might not be working on Node 12/master. It's not even in CIGTM.

Because of the critical importance for older packages, can somebody expert on this check it out?

kkoopa:
I just published a long overdue update, but it does not address Node 12.

PR #831, opened 2018-12-17
add support for node master

Flarna:
This PR allows to use NAN with current Node on master and should fix several fails in CITGM which use NAN.

Not sure if this should be really included in NAN at this time as Node plans to do further updates of V8 till v12 is released (see https://github.com/nodejs/node/issues/25060#issuecomment-447748949).

There is still a self deprecation warning in v8 left but I think this should be addressed inside v8 or in the patched v8 variant included in node.

Refs: https://github.com/nodejs/node/issues/25060
Fixes: #829

### DIFF
--- a/nan_converters_43_inl.h
+++ b/nan_converters_43_inl.h
@@ -19,13 +19,22 @@ imp::ToFactory<v8::TYPE>::convert(v8::Local<v8::Value> val) {                  \
           .FromMaybe(v8::Local<v8::TYPE>()));                                  \
 }
 
-X(Boolean)
 X(Number)
 X(String)
 X(Object)
 X(Integer)
 X(Uint32)
 X(Int32)
+#if NODE_MODULE_VERSION <= NODE_11_0_MODULE_VERSION
+X(Boolean)
+#else
+imp::ToFactory<v8::Boolean>::return_t                                          \
+imp::ToFactory<v8::Boolean>::convert(v8::Local<v8::Value> val) {               \
+  v8::Isolate *isolate = v8::Isolate::GetCurrent();                            \
+  v8::EscapableHandleScope scope(isolate);                                     \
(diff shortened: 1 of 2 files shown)


kkoopa:
Looking good. Not too many breaking changes so far. Will hold off until Node 12 gets closer to release. Thank you.

Flarna:
Rebased to 2.12.1.

I'm fine with waiting.

I think there are quite some people out there which would be happy with a nan release working with node master. But it's up to them to convince you to take this maintenance burden 😏.

Maybe someone (or even myself) finds some time to port nan to use napi. This would help a lot in future but not sure if it is that easy. It would be for sure a good test for napi completeness.

kkoopa:
> I think there are quite some people out there which would be happy with a nan release working with node master. But it's up to them to convince you to take this maintenance burden .

Maintenance burden indeed. Unfortunately, I do not have time to follow a volatile moving target with actual releases, since every release comes with a risk of something unintentionally breaking. However, merging to master would not hurt. Whoever wants to be on the bleeding edge can easily do so then.

> Maybe someone (or even myself) finds some time to port nan to use napi. This would help a lot in future but not sure if it is that easy. It would be for sure a good test for napi completeness.

I had originally thought of doing precisely that, but shelved the plans due to lack of time and motivation. A common feature set could likely be mapped onto NAPI, but AFAIK there are be some (rare) V8-specific features which arguably do not fit in napi, e.g. the garbage collection API.

fivdi:
@kkoopa do you think it's time to release a new version of nan with the modifications from this PR? Node 12 is scheduled to be released in April (https://github.com/nodejs/release.)

kkoopa:
Yes, it is indeed time. Thank you for reminding me.

jwulf:
It's out now.

Issue #832, opened 2018-12-17
WriteUtf8  does not take 5 arguments


Sneezry:
Hi, I have no idea if I made some mistake. But I cannot rebuild nan with electron 3.0.10 (with node 10.2.0).

With command below to rebuild:

```
node-gyp rebuild --target=3.0.10 --arch=x64 --dist-url=https://atom.io/download/electron
```

The raw error I got is

```
C:\Users\...\node_modules\nan\nan.h(1081): error C2660: 'v8::String
::WriteUtf8': function does not take 5 arguments (compiling source file ..\src\combined.cpp) [C:\...\xxx.vcxproj]
```

I find that the latest nan will call `WriteUtf8` with 5 arguments if node version is larger than 10

https://github.com/nodejs/nan/blob/55923c824e4c148b127faec9b57d5b914f6a9ab9/nan.h#L1081

However, `WriteUtf8` only accepts 4 arguments in node 10

https://v8docs.nodesource.com/node-10.6/d2/db3/classv8_1_1_string.html#a99aec0d3b418feca1f2a2ba79416c60b

Any ideas? Thanks.

Flarna:
The 4 param version has been deprecated in v8 in favor of the 5 param version and this has been backported to node 10 (see https://github.com/nodejs/node/pull/22531). Therefore the linked v8 docs don't match to the actual variant of v8 used by node.

See https://github.com/nodejs/node/blob/4128793a150dbcc63e03738ddd30f1768608aa11/deps/v8/include/v8.h#L2656

I assume that the same did not happen in electron and as a result some ifdefs are wrong in this setup.

Flarna:
Actually it seems that the backport of the v8 change was done with Node 10.10.0. Therefore if someone uses Nan 2.12.0 with Node 10.0.0 till 10.9.0 build will fail.

Flarna:
Caused by #825 - seems it's a decision between build errors and deprecation warnings.

Flarna:
@amb26 I guess you mean compiling electron with NodeJs 10.10.0. Or some native node extension?

refack:
@Flarna following your suggestion from https://github.com/mapbox/node-sqlite3/pull/1093#pullrequestreview-185457481 I got this situation:
* `nan@2.11.1` in node-sqlite3 on `NODE_VERSION="6" ELECTRON_VERSION="3.0.6"` compiles:
  https://travis-ci.org/mapbox/node-sqlite3/jobs/468473910

[...]

PR #833, opened 2018-12-17
Fix build breakage with Node.js 10.0.0-10.9.0.

bnoordhuis:
The five argument WriteUtf8() method was introduced in Node.js v10.0.0,
it doesn't exist in older 10.x releases.

Use the four argument overload and a bunch of pragmas to silence the
compiler warnings.

Fixes: https://github.com/nodejs/nan/issues/832
Refs: https://github.com/nodejs/nan/pull/825

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -1077,11 +1077,27 @@ class Utf8String {
         }
         const int flags =
             v8::String::NO_NULL_TERMINATION | imp::kReplaceInvalidUtf8;
-#if NODE_MAJOR_VERSION >= 10
+#if NODE_MAJOR_VERSION >= 11
         length_ = string->WriteUtf8(v8::Isolate::GetCurrent(), str_, static_cast<int>(len), 0, flags);
 #else
-        length_ = string->WriteUtf8(str_, static_cast<int>(len), 0, flags);
+        // See https://github.com/nodejs/nan/issues/832.
+        // Disable the warning as there is no way around it.
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4996)
 #endif
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
+        length_ = string->WriteUtf8(str_, static_cast<int>(len), 0, flags);
+#ifdef __GNUC__
+#pragma GCC diagnostic pop
+#endif
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
+#endif  // NODE_MAJOR_VERSION < 11
         str_[length_] = '\0';
       }
     }


nan.h
```diff
@@ -1077,10 +1077,23 @@ class Utf8String {
         }
         const int flags =
             v8::String::NO_NULL_TERMINATION | imp::kReplaceInvalidUtf8;
-#if NODE_MAJOR_VERSION >= 10
-        length_ = string->WriteUtf8(v8::Isolate::GetCurrent(), str_, static_cast<int>(len), 0, flags);
-#else
+        // See https://github.com/nodejs/nan/issues/832.
+        // Disable the warning as there is no way around it.
+        // TODO(bnoordhuis) Use isolate-based version in Node.js v12.
```
Flarna: I think the isolate version of `WriteUtf8` is already available since v11 so `#if NODE_MAJOR_VERSION >= 11` should be fine here.

Flarna:
@bnoordhuis Do you think it would be a good idea to extend the test matrix with the initial node.js versions (e.g. add `TRAVIS_NODE_VERSION="10.0.0"`,...)?

kkoopa:
We could do that for 10, but doing it for others would lead to too large a test matrix.

Ideally it would not be necessary. I still do not see why changes of this nature are introduced in minor versions of Node, but that ship has sailed so we have to make the best of it.

Flarna:
I would vote to add it for the "active" NodeJs versions, e.g. 10 and 11. I don't expect that anyone will touch older versions but I could imagine that the same deprecation mechanism used in Node 10 could happen again in 11 to prepare for 12.
Maybe remove versions like 5 from the matrix to reduce load?

nan.h
MaartenBent: What about
```c++
#if !NODE_VERSION_AT_LEAST(10,9,0)
```
It will use the isolate-based version for 10.10 and higher.

Flarna: @MaartenBent There are situations where the node version used to build differs from that one used to execute the program. Therefore if you build with 10.10.0 but execute with 10.9.0 it will not work as the new API is not there. This is not intended as usually a native addon binary works for all minors of a major version.

MaartenBent: You're right, I didn't consider ABI compatibility.

bnoordhuis: > I think the isolate version of WriteUtf8 is already available since v11 so #if NODE_MAJOR_VERSION >= 11 should be fine here.

Right, that's true. Updated.

bnoordhuis:
@kkoopa LGTY? Would be best to put out a bug fix release ASAP.

kkoopa:
Yes, I will look through it and likely publish tonight when I get home.

kkoopa:
Done, thank you both for triaging this.

PR #842, opened 2019-03-15
remove usage of deprecated IsNearDeath

Flarna:
Remove use of `PersistentBase::IsNearDeath()` as it has been deprecated in V8 and it would causes compilation warnings in Node.JS 12.

I have not added any `ifdefs` here as it was used only within `asserts`.

Refs: https://github.com/nodejs/node/pull/26630

### DIFF
--- a/nan_object_wrap.h
+++ b/nan_object_wrap.h
@@ -21,7 +21,6 @@ class ObjectWrap {
       return;
     }
 
-    assert(persistent().IsNearDeath());
     persistent().ClearWeak();
     persistent().Reset();
   }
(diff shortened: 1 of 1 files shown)


nolanmar511:
I would love to see this change merged into master (if possible).

PersistentBase::IsNearDeath() has been removed in the version of v8 used in the v8-canary build of Node 12, and it would be nice to be able to test with the v8-canary build with the version of nan that can be installed from github.

bnoordhuis:
@kkoopa See above. Do you want to do a release for this? It's an otherwise harmless change.

kkoopa:
Yes, I was planning to do so this week.

PR #841, opened 2019-03-14
[WIP] spec: run tests on Electron as well

MarshallOfSound:
This is a WIP while I figure it out, opening the PR to trigger builds

PR #848, opened 2019-04-20
add electron to travis test matrix

Flarna:
Seems that #841 got stuck so tried it by myself to add electron to travis test matrix.


.travis.yml
```diff
@@ -30,6 +33,7 @@ install:
   - if [[ $TRAVIS_OS_NAME == "linux" ]]; then export CXX=g++-4.8; fi
   - $CXX --version
   - if [[ $TRAVIS_NODE_VERSION == "0.8" ]]; then npm config set strict-ssl false; fi
-  - if [[ $(echo "$TRAVIS_NODE_VERSION < 4" | bc -l) ]]; then npm install npm@2 && mv node_modules npm && npm/.bin/npm --version && npm/.bin/npm install; else npm --version && npm install; fi
-  - node_modules/.bin/node-gyp rebuild --directory test
-script: node_modules/.bin/tap --gc test/js/*-test.js
+  - if [[ -z "$ELECTRON_VERSION" && $(echo "$TRAVIS_NODE_VERSION < 4" | bc -l)  == "1" ]]; then npm install npm@2 && mv node_modules npm && npm/.bin/npm --version && npm/.bin/npm install; else npm --version && npm install; fi
+  - if [[ -z "$ELECTRON_VERSION" ]]; then node_modules/.bin/node-gyp rebuild --directory test; else node_modules/.bin/node-gyp rebuild --target=v$ELECTRON_VERSION --dist-url=https://atom.io/download/electron --directory test; fi
+script:
+  - if [[ -z "$ELECTRON_VERSION" ]]; then node_modules/.bin/tap --gc test/js/*-test.js; fi
```
mkrufky: It looks like this is just a build check, if we're not going to actually run the node.js/nan unit tests.

Personally, I'd love to see electron added to the test matrix, but it would make a lot more sense if we could actually run the tests.  I don't kno wif there's a straight forward way to do this, or if we need to wrap them, maybe even create a separate set of tests for electron.....

Flarna: Yes, just build tests.
Honestly speaking I have no idea what's actually needed to run tests with electron but I fear it's not that easy as run nodejs which is available on the CI machines.

Flarna:
rebased, added electron 5.0.1 and updated other electron version to the current latest version.
Seems travis is not canceling builds if a new commit is made on the same branch. is this intended?

bnoordhuis:
Oh wait, maybe I misread the diff the first time around. The `node-gyp rebuild` command at least should run and against electron's headers. Currently that step is skipped too.

Flarna:
@bnoordhuis if I understand the travis log correct it calls `node_modules/.bin/node-gyp rebuild --directory test; else node_modules/.bin/node-gyp rebuild --target=v$ELECTRON_VERSION --dist-url=https://atom.io/download/electron --directory test` which should be what we want.

Flarna:
and looking into the variant for 5.0.1 it seems there is some more work to do in NAN as there are several deprecation warnings. But not sure if this should be done in this PR.

bnoordhuis:
Sorry, you're right. Travis folds the build output but unfolding didn't show anything yesterday. It's there now though. Okay, objection withdrawn. :-)

kkoopa:
Thank you. Hopefully this will help in avoiding regressions.

PR #850, opened 2019-04-25
Add node 12 support

Flarna:
No functional change, just update readme, CI and add `NODE_12_0_MODULE_VERSION`.

Refs: #849

PR #853, opened 2019-05-02
add windows test support to Travis

msimerson:
This is test PR, adding windows build support to Travis. It's an alternative for #850 to waiting for appveyor support.  IME, Travis windows builds are a fair bit slower than appveyor, but I've also had plenty of false positives on appveyor tests, so it's a mixed bag.

This PR should kick off a Travis build and we can see if it works or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os:
   - linux
   - osx
+  - windows
 language: cpp
 addons:
   apt:


msimerson:
the builds fail on Travis windows. I don't have the time or inclination to track it down. Someone that has an interest in this though could try running a shell script instead of the series of CLIs, and then add a custom path for building on Travis Windows.

Autoclose me in 30 days, if no progress.

bnoordhuis:
nvm doesn't support Windows and has no plans to do so ([ref](https://github.com/nvm-sh/nvm#important-notes)) so I'm not surprised it doesn't work out of the box. [nvm-windows](https://github.com/coreybutler/nvm-windows) will probably get you further although you probably also need to clean up the build rules.

I'll close this out for now but let me know if you want to continue working on it.

PR #857, opened 2019-05-28
spec: add support for enabling gc at runtime in tests

MarshallOfSound:
This allows folks running the `nan` tests in their own CI to run it without `--expose_gc` being supported as an incoming CLI flag.

The backstory here is I'm setting up some infra to run nan's test suite on Electron CI so that we know when things are gonna break as typically nowadays we are on the fringe of V8 👍

### DIFF
--- a/test/js/gc-fn.js
+++ b/test/js/gc-fn.js
@@ -1,0 +1,10 @@
+if (global.gc) {
+  module.exports = global.gc
+  return
+}
+
+var v8 = require('v8');
+var vm = require('vm');
+
+v8.setFlagsFromString('--expose_gc');
+module.exports = vm.runInNewContext('gc');
--- a/test/js/gc-fn.js
+++ b/test/js/gc-fn.js
@@ -1,6 +1,6 @@
 if (global.gc) {
-  module.exports = global.gc
-  return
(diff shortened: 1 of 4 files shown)


test/js/gc-fn.js
```diff
@@ -0,0 +1,10 @@
+if (global.gc) {
+  module.exports = global.gc
+  return
```
bnoordhuis: Tiniest of slightest of micro-nits: can you add semicolons here?

test/js/weak2-test.js
```diff
@@ -22,10 +22,10 @@ test('weak2', function (t) {
   });
 
   // run weak callback, should dispose
-  gc();
+  require('./gc-fn')();
 
   // do not run weak callback
-  gc();
+  require('./gc-fn')();
```
bnoordhuis: Not a blocker but I would like it if you `, gc = require('./gc-fn')` at the top of the file.

(Also applies to the other tests, of course.)

test/js/gc-fn.js
MarshallOfSound: Sure sorry, I was working in a repo without semicolons when I wrote this originally 😆

PR #859, opened 2019-06-11
fix: Install Node independent from AppVeyor images

felixrieseberg:
Even though Node.js v12 is in the test matrix, it's not being tested. The cause is AppVeyor, which hasn't added v12 to its images just yet.

Luckily, there's a way to have AppVeyor install whatever Node version we want – using `Update-NodeJsInstallation`, which is present in all images.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ environment:
 # Install scripts. (runs after repo cloning)
 install:
   # Get the latest stable version of Node 0.STABLE.latest
-  - ps: Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:nodejs_version) $env:PLATFORM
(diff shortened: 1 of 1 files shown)


kkoopa:
I learn something new every day. Thanks.

mathiask88:
This should be a bit faster:
```YAML
  - ps: |
        try {
          Install-Product node $env:nodejs_version
        } catch {
          echo "Unable to install node $env:nodejs_version, trying update..."
          Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:nodejs_version)
        }
```
So AppVeyor can use the preinstalled major version of node if it exists and fall back to download/install the bleeding edge else.

kkoopa:
Seems reasonable and should indeed be faster. AppVeyor is slow enough as it is due to the long build queues.

PR #876, opened 2019-11-12
Add data to SetMethod and SetPrototypeMethod + documentation + test

drorgl:
This is a similar to pr [602](https://github.com/nodejs/nan/pull/602), I've included tests and documentation changes.

### DIFF
--- a/doc/methods.md
+++ b/doc/methods.md
@@ -478,10 +478,12 @@ Signature:
 ```c++
 void Nan::SetMethod(v8::Local<v8::Object> recv,
                     const char *name,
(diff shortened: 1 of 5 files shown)


drorgl:
Is it possible to re-run just the failing job? I rebuilt it inside a clean container with no problems with v0.10, I want to make sure its not a temporary travis problem.

kkoopa:
Don't know an easy way of rerunning just one job, but it does look like some glitch in Travis.

kkoopa:
Thank you.

PR #880, opened 2019-11-30
Add variant of Nan::New<v8::String> which takes a C++17 string_view

robinchrist:
This PR adds `Nan::New<v8::String>(std::string_view)`, which makes it easier to handle compile time known strings.

In order to maintan backwards compatibility to C++11, a `NAN_HAS_CPLUSPLUS_17` macro is added (which will also be needed for my next PRs), the code is wrapped into `NAN_ENABLE_STRING_VIEW` guards. If the user wishes to disable the usage of string_view, it can be disabled via adding a define `-DNAN_ENABLE_STRING_VIEW=false`, otherwise it is automatically enabled if it's supported by the C++ runtime (C++17 and up)

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -53,6 +53,10 @@
 
 #define NAN_HAS_CPLUSPLUS_17 (__cplusplus >= 201703L)
 
+#ifndef NAN_ENABLE_STRING_VIEW
+# define NAN_ENABLE_STRING_VIEW NAN_HAS_CPLUSPLUS_17
+#endif
+
 #include <uv.h>
 #include <node.h>
 #include <node_buffer.h>
@@ -73,6 +77,9 @@
 # include <queue>
 # include <string>
 # include <vector>
+#endif
+#if NAN_ENABLE_STRING_VIEW
+# include <string_view>
 #endif
 
 // uv helpers
--- a/nan.h
+++ b/nan.h
@@ -51,6 +51,8 @@
 # error This version of node/NAN/v8 requires a C++11 compiler
 #endif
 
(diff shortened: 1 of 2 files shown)


robinchrist:
@bnoordhuis 
> 2\. `-DNAN_ENABLE_STRING_VIEW=false` won't undefine the macro. `defined(NAN_ENABLE_STRING_VIEW)` is true because the macro is non-empty, it evaluates to the string `"false"`.

Hmm, I'm not sure whether I can follow.
I have prepared a minimum repro here: https://godbolt.org/z/DET2bw

Without any additional `NAN_ENABLE_STRING_VIEW` flags:
Using `-std=c++11` -> `Output without string_view!`
Using `-std=c++17` -> `Output using string_view!`

Using `-std=c++17 -DNAN_ENABLE_STRING_VIEW=false` -> `Output without string_view!`
Using `-std=c++17 -DNAN_ENABLE_STRING_VIEW=true` -> `Output using string_view!`
Using `-std=c++11 -DNAN_ENABLE_STRING_VIEW=true` -> Compile error, as expected

Or should I change the meaning of `NAN_ENABLE_STRING_VIEW`?
Basically, instead of "By setting `NAN_ENABLE_STRING_VIEW` you force the inclusion of `string_view`" we say "By setting `NAN_ENABLE_STRING_VIEW`, you can enable or disable `string_view`, but if it's not compatible, it's disabled anyways"?

bnoordhuis:
Ah, I think I know where the misunderstanding comes from: `-DNAN_ENABLE_STRING_VIEW=false` is syntax that's accepted by npm and node-gyp but a) it does something different, and b) it's not what you meant. Okay, objection withdrawn.

Issue #884, opened 2020-02-11
silently failing make


majioa:
I'm trying to run make from the root pf the project but it is failed to lint the files silently:

    $ PYTHON=/usr/bin/python3 make
[...]

PR #885, opened 2020-02-11
Make cpplint.py compatible with python 3

bnoordhuis:
Commit af24c7162be from 2018 allegedly also did that but it dies
silently for me and others with python 3.7.

This commit fixes the incompatibilities and adds an excepthook so
you get an indication of what the issue is.

Fixes: https://github.com/nodejs/nan/issues/884

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ LINT_SOURCES = \
 	test/cpp/wrappedobjectfactory.cpp \
 	node_modules/node-gyp/gyp/data/win/large-pdb-shim.cc
 
-FILTER = -whitespace/parens
+FILTER = -build/include_subdir,-whitespace/parens
 
 .PHONY: lint
 
--- a/cpplint.py
+++ b/cpplint.py
@@ -44,30 +44,48 @@ same line, but it is far from perfect (in either direction).
 import codecs
 import copy
 import getopt
+import glob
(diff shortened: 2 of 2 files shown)


kkoopa:
This results in `<class 'TypeError'> write() argument must be str, not bytes <traceback object at 0x10179ecc8>`. Running with Python 2, shows that it dies just before outputting this error message: `Skipping input 'node_modules/node-gyp/gyp/data/win/large-pdb-shim.cc': Can't open for reading`. Now, this is a `str`, not `bytes`, but apparently there could have been bytes written to stdout somewhere earlier. `2to3` did not seem to include any fix for this. This will be a minor pain to correctly fix, since toy languages do not have strong typing.

bnoordhuis:
Are you saying it's failing for you locally with python 2? With what OS is that? Windows?

kkoopa:
Python 3 on a not-proper-unix Mac. Anyway, latest cpplint from upstream seems to work for both 2 and 3, along with some new additions, let's just update. https://raw.githubusercontent.com/cpplint/cpplint/master/cpplint.py

bnoordhuis:
That works for me too. Updated, PTAL.

kkoopa:
Works for me with both Python2 and Python 3.

PR #887, opened 2020-03-02
test: create HandleScope in weak callback

nornagon:
By analogy with https://chromium.googlesource.com/v8/v8/+/8c8bd658c6f3a765a73ac49e57b7b138a67db635/test/cctest/test-api.cc#13418


kkoopa:
Thank you both.

PR #888, opened 2020-03-06
use GetBackingStore() instead GetContents()

Flarna:
Use `GetBackingStore()` instead `GetContents()` for v8 >= 8.0 as `GetContents()` has been deprecated. Node.js 14 will use 8.1.

It would be possible to use `GetBackingStore()` already with 7.9 but this would lead to issues with Node.js 13 as it uses 7.8 till 13.2.0.

Additionally update testmatrix to use latest electron versions and add electron 8.

Refs: https://github.com/nodejs/node/issues/32108
Refs: https://github.com/nodejs/node/pull/30339


nan_typedarray_contents.h
```diff
@@ -31,7 +31,13 @@ class TypedArrayContents {
       v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
 
       length = byte_length / sizeof(T);
-      data   = static_cast<char*>(buffer->GetContents().Data()) + byte_offset;
+// actually it's 7.9 here but this would lead to ABI issues with NodeJs 13
+// using 7.8 till 13.2.0
```
bnoordhuis:
```suggestion
// Actually it's 7.9 here but this would lead to ABI issues with Node.js 13
// using 7.8 till 13.2.0.
```

PR #889, opened 2020-04-21
WIP: Add Node 14 to test matrix, update electron versions

Flarna:
Welcome Node 14. I think it should work out of the box as adaptions for v8 8.1 have been done already in #888.


Flarna:
Seems there are some more adaptions needed as there are deprecation warnings.

kkoopa:
I shall try and have a look at the warnings.

kkoopa:
I did not see any deprecation warnings in the logs from the CI. Just warnings about the attribute `deprecated` not being recognized originating upstream.

Flarna:
Correct, I misinterpreted this. Maybe a newer compiler should be used for node 13 onwards.

Would be nice to get a nan release which includes node 14 support.

kkoopa:
Right, probably it wants a newer compiler. Did not get any warnings on Node 13 when testing locally. I shall publish a new version tonight.

Flarna:
https://github.com/nodejs/node/blob/v14.0.0/BUILDING.md tells that at least GCC 6.3 is needed and VS2017 on windows. So it works better as expected as here still 4.8 is used which was Node 4 as far as I remember.

Issue #909, opened 2021-03-03
nan is affected by the update to V8 8.9 in Node.js


targos:
See https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2639/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/nan_v2_14_2/

Flarna:
Is this the v8 version which will be in Node 16? Or is it expected that there are more API changes till then?

targos:
Not sure. We might still be able to update V8 once more before Node 16.

PR #910, opened 2021-03-03
tests: adapt for v8 8.9

Flarna:
V8 API for `ScriptOrigin` constructor has changed. Adapt test to compile with v8 8.9.

Currently only `ScriptOrigin` instances are consumed by NAN therefore no adaptions in the NAN API are needed to keep build/tests working.

fixes: #909

### DIFF
--- a/test/cpp/nannew.cpp
+++ b/test/cpp/nannew.cpp
@@ -246,7 +246,12 @@ NAN_METHOD(testScript) {
 
   t.plan(6);
 
+#if defined(V8_MAJOR_VERSION) && (V8_MAJOR_VERSION > 8 ||                      \
+  (V8_MAJOR_VERSION == 8 && defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 9))
+  ScriptOrigin origin(New("foo").ToLocalChecked(), 5);
+#else
   ScriptOrigin origin(New("foo").ToLocalChecked(), New(5));
+#endif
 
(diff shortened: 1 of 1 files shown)


Flarna:
hmm, CI seems to be on vacation.

kkoopa:
Yes, do not know what is up with Travis. AppVeyor runs https://ci.appveyor.com/project/RodVagg/nan

Flarna:
Forced push to trigger CI again but without success for travis.

AppVeyor runs but it fails for 0.8 with
```
'npm' is not recognized as an internal or external command,
operable program or batch file.
```

lognaturel:
Thanks for pushing this forward, @Flarna and @kkoopa! I've landed here following a dependency chain that needs updating. Do you know whether non-test updates will be needed to support V9.0 which ships with Node 16?

kkoopa:
I have not tested or looked into it, but if it builds without errors originating from NAN itself, everything should be good.

lognaturel:
Thanks, @kkoopa, that makes sense. I'll aim to try it and report back.

Flarna:
I use latest NAN release since a while without issues. But there are for sure areas of NAN I don't use there.

But it would be still nice if this PR could be merged (or a variant of it) to get node/citgm green again and early hints for future issues like moving to v8 9.1 or in node 17.

lognaturel:
🚀 🚀 🚀

PR #913, opened 2021-05-19
Add a note about overriding the Handle*Callbacks

mmomtchev:
The initial statement can be somewhat misleading because the class will take care of running in the correct async context only if the callbacks are not overriden.


abarisain:
just got bitten by this, thanks for your PR, you saved me hours of pain :)

hope it gets merged!

PR #916, opened 2021-07-02
Fix the example.

sunoru:
The async_pi_estimate example could not compile.


kkoopa:
Thank you for your contribution.

PR #926, opened 2021-12-16
docs: license format and contributors

eddiemonge:
The license file differed enough that licensed did not recognize it.

The link to the contributors was also outdated.


kkoopa:
Thank you.

Issue #924, opened 2021-11-08
Support Node.js 17


matteodisabatino:
In the Readme there is no mention to Node.js v17. Is it officially supported?

matteodisabatino:
Sorry, may I have any news=

kkoopa:
The test suite passes, so I presume that everything works.

PR #928, opened 2022-01-02
add node 17 to the test matrix

mkrufky:
relates to: https://github.com/nodejs/nan/issues/924


kkoopa:
Travis has not worked ever since they changed their policies on OS projects and public repositories, but I guess it does not hurt to add it to the config, it simply never runs. Luckily AppVeyor still works.

PR #940, opened 2022-05-25
package.json: fix supported Node.js versions in package description

mkrufky:
Oops!  It looks like we may have forgotten to update this field for a while...

from:
```
"description": "Native Abstractions for Node.js: C++ header for Node 0.8 -> 14 compatibility"
```

to:
```
"description": "Native Abstractions for Node.js: C++ header for Node 0.8 -> 18 compatibility"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "nan",
   "version": "2.16.0",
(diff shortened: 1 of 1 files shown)


kkoopa:
Ah, right. Should not warrant a new release, but please merge and I will try to remember updating it in the future.

Issue #942, opened 2022-08-04
nan_callbacks.h:55:23: error: ‘AccessorSignature’ is not a member of ‘v8’


Abdullah-Albinali:
i am not sure i am  posting this in the correct place, but i am facing the below issue and i am not sure how to proceed further on this.

appreciate any push in the right direction.

```
⠙ Building module: node-pty, Completed: 0In file included from ../../nan/nan.h:180,
                 from ../src/unix/pty.cc:20:
../../nan/nan_callbacks.h:55:23: error: ‘AccessorSignature’ is not a member of ‘v8’
   55 | typedef v8::Local<v8::AccessorSignature> Sig;
      |                       ^~~~~~~~~~~~~~~~~
../../nan/nan_callbacks.h:55:40: error: template argument 1 is invalid
   55 | typedef v8::Local<v8::AccessorSignature> Sig;
      | 
```

westlakem:
Same problem when trying to compile for Electron v20

nikitakot:
same problem, using `electron-rebuild` to rebuild a native module which uses `nan` for Electron v20 and getting `nan_callbacks.h(55,23): error C2039: 'AccessorSignature': is not a member of 'v8'` and similar errors related to `AccessorSignature`

```
"nan": "2.16.0
```

RaisinTen:
A PR has already been sent to fix this - https://github.com/nodejs/nan/pull/941.

chouzz:
> A PR has already been sent to fix this - #941.

Confirmed, this PR fix my problem when rebuild electron with node-pty.

Abdullah-Albinali:
confirmed as well! thank!

RaisinTen:
Let's keep this open for now. It will get closed automatically when the PR lands!

alejandroclaro:
In case someone is looking for workarounds until this is fixed, take a look to: https://github.com/electron/electron/issues/35193

RichardFoss:
I am using a midi package. Recently I updated to node v 18.10.0 and nw v0.69.1. When I rebuild:
nw-gyp rebuild --target=0.69.1 --arch=x64
I get:
../../nan/nan_callbacks.h:55:23: error: no member named 'AccessorSignature' in namespace 'v8'
typedef v8::Localv8::AccessorSignature Sig;
~~~~^
I have looked at various resolutions to the issue in this thread. I have tried putting this into the midi package.json:
"overrides": {
[...]

PR #943, opened 2022-09-06
chore: overload deprecated AccessorSignatures

VerteDinde:
AccessorSignatures was removed in https://chromium-review.googlesource.com/c/v8/v8/+/3654096 . This PR overloads the SetAccessor method, removing the default parameters, and also includes the original method with the signature param removed. This should allow older versions of Node to use `signature`, but fixes nan for newer versions of Node and Electron

Supersedes: #941
Fixes: https://github.com/nodejs/nan/issues/942

@kkoopa This PR attempts to address the feedback left here: https://github.com/nodejs/nan/pull/941#issuecomment-1162385143 If this doesn't encompass what you were thinking, let me know, happy to make adjustments!

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -2509,15 +2509,15 @@ inline void SetPrototypeMethod(
 
 //=== Accessors and Such =======================================================
 
-inline void SetAccessor(
+NAN_DEPRECATED inline void SetAccessor(
     v8::Local<v8::ObjectTemplate> tpl
   , v8::Local<v8::String> name
   , GetterCallback getter
-  , SetterCallback setter = 0
-  , v8::Local<v8::Value> data = v8::Local<v8::Value>()
-  , v8::AccessControl settings = v8::DEFAULT
-  , v8::PropertyAttribute attribute = v8::None
-  , imp::Sig signature = imp::Sig()) {
+  , SetterCallback setter
+  , v8::Local<v8::Value> data
+  , v8::AccessControl settings
+  , v8::PropertyAttribute attribute
+  , imp::Sig signature) {
   HandleScope scope;
 
   imp::NativeGetter getter_ =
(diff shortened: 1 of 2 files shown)


alexciarlillo:
I am trying this out since the branch for #941 (which was working) no longer exists. I am wondering where the choice of node module version 18 comes from. I am on node 16.15.0 and trying to compile a module against electron 20 (also 16.15.0) and the version check here doesn't work for me. If I modify it to `NODE_16_0_MODULE_VERSION` it works as expected.

alexciarlillo:
I'm not sure if I got that error but IIRC I was also still getting

```
error: ‘AccessorSignature’ is not a member of ‘v8’
```

Sorry I should have kept better track and posted more details. I have just been in dependency hell for like a week trying to do an electron upgrade and my brain was a bit fried. But I can confirm this fix doesn't work for me on Node 16.15.x where the previous one did.

seb-mtl:
Same here, it reports this one as still being used:

https://github.com/VerteDinde/nan/blob/1719071c1b74e569f7e6f57c589fc9e5020a3f6e/nan.h#L2546-L2556

```cpp
    npm ERR! In file included from ../../nan/nan.h:180:
    npm ERR! ../../nan/nan_callbacks.h:55:23: error: no member named 'AccessorSignature' in namespace 'v8'
    npm ERR! typedef v8::Local<v8::AccessorSignature> Sig;
    npm ERR!                   ~~~~^
    npm ERR! In file included from ../src/functions.cpp:20:
    npm ERR! In file included from ../src/functions.h:4:
    npm ERR! ../../nan/nan.h:2546:8: error: no matching member function for call to 'SetAccessor'
    npm ERR!   tpl->SetAccessor(
```

Built on macOS for:
Local node: v16.15.0
Electron v19.0.15

kkoopa:
Yes, something like this was what I had in mind, API-wise.

kkoopa:
As for the `AccessorSignature`, having a dummy one as `Data` should work.
```diff
@@ -53,7 +53,9 @@ typedef void(*IndexQueryCallback)(

 namespace imp {
 #if (NODE_MODULE_VERSION < NODE_18_0_MODULE_VERSION)
-NAN_DEPRECATED typedef v8::Local<v8::AccessorSignature> Sig;
+typedef v8::Local<v8::AccessorSignature> Sig;
+#else
+typedef v8::Local<v8::Data> Sig;
 #endif

 static const int kDataIndex =                    0;
```

miniak:
Can we expedite merging this fix and releasing a new version? We cannot build our node modules using NaN for Electron 20, which has been out for over a month

segmentedcontrol:
@kkoopa 
image

VerteDinde:
@kkoopa Done, thanks! Added the dummy sig as suggested.

MRayermannMSFT:
@VerteDinde this branch appears to work well for me now. :)

michalzaq12:
image

miniak:
FYI Electron 21 was released in the meantime

devinbinnie:
Any idea when this will get merged? Trying to upgrade my app to Electron v20 and am currently blocked on this :(

alejandroclaro:
Any update?

alejandroclaro:
In case someone is looking for a temporal workaround: https://github.com/electron/electron/issues/35193

bnoordhuis (commented):
I might be missing something but why replicate SetAccessor() in its entirety when the only thing that's different is the disappearance of the `signature` argument?

RaisinTen:
> I might be missing something but why replicate SetAccessor() in its entirety when the only thing that's different is the disappearance of the `signature` argument?

@bnoordhuis I think that's because of @kkoopa's review comment in the previous PR - https://github.com/nodejs/nan/pull/941#issuecomment-1162385143.

kkoopa:
Sorry, completely forgot about this during vacation. I'll go through it now.

kkoopa:
Yes, this looks like what I had in mind and it seems to pass the tests. I need to add some more entries to the test matrix and do some maintenance before I can push a new release. Thank you all.

kkoopa:
New release published. Hopefully this resolves all problems.

segmentedcontrol:
all hail benjamin!

MRayermannMSFT:
thanks-a-bunch-thankyou

PR #945, opened 2022-11-18
Add Node 19 to CI and update README

UNIDY2002:
Additional changes: set build image of AppVeyor CI to Visual Studio 2017

This is mainly because of some usages of c++17 features in Node 19 headers.

Since I have no access to the Travis CI logs, I am not sure whether the CI passes in Travis.

PR #949, opened 2023-04-25
Add Node.js 20

westy92:
I attempted to copy https://github.com/nodejs/nan/pull/945.


westy92:
AppVeyor succeeded. ✅ 
https://ci.appveyor.com/project/RodVagg/nan/builds/46876091

westy92:
What is the process for getting a release with this included?

kkoopa:
Well, since it does not actually do anything, there will not be a release in the foreseeable future. The version detection is only used for handling specific logic, which does not apply to Node 20 as far as I am aware. The current version of NAN exhibits identical behavior. I have not personally tried Node 20, but unless there is something specifically broken, of which I am not aware, ecerything ahould work fine with the current version.
Thank you for the contribution though, it would still have had to be added at some point.

PR #950, opened 2023-05-06
new: Add CMake as an alternative build tool

zenon8adams:
This commit integrates CMake as an alternative build tool for the project. Users can now choose to use CMake to configure and build the project, in addition to the existing Makefile.


kkoopa:
Thank you, this is a nice addition. Just need to remember to keep this in sync with the Makefile.

PR #956, opened 2023-09-12
Cast `v8::Object::GetInternalField()` return value to `v8::Value`

codebytere:
Refs https://github.com/nodejs/node/pull/48943
Refs https://chromium-review.googlesource.com/c/v8/v8/+/4834471

Electron is hitting build failures on this as a result of the above CL - This PR fixes these issues by casting `v8::Object::GetInternalField()` return values to `v8::Value`.


kkoopa:
Thank you.

PR #960, opened 2023-11-01
chore: use new constructor for ScriptOrigin when >= 17.x

codebytere:
Refs https://chromium-review.googlesource.com/c/v8/v8/+/3395880.

Chromium did an advance deprecation removal of the `ScriptOrigin` ctor without an Isolate.

This allows Electron to remove a [patch](https://github.com/electron/electron/blob/61374019c03ab1af8584550277318ba7c2adb0ca/patches/nan/use_new_constructor_for_scriptorigin_when_17_x.patch#L1-L29)

### DIFF
--- a/test/cpp/news.cpp
+++ b/test/cpp/news.cpp
@@ -115,7 +115,7 @@ NAN_METHOD(NewScript) {
 
 NAN_METHOD(NewScript2) {
   v8::ScriptOrigin origin(
-#if NODE_MODULE_VERSION >= NODE_18_0_MODULE_VERSION
+#if NODE_MODULE_VERSION >= NODE_17_0_MODULE_VERSION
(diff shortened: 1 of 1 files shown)


kkoopa:
Why was the check originally for Node 18? I am AFK for about a week and cannot look into it until later.

codebytere:
@kkoopa that's when the new constructor was initially slated to be removed, but they moved up the date after all uses in Chromium got removed it seems.

codebytere:
@kkoopa I'd still love a look at this when you can!

codebytere:
They've now removed this ctor entirely:  https://chromium-review.googlesource.com/c/v8/v8/+/5539888

So we'll likely need to pend this until Node.js contains the version of V8 that removes it as well.

PR #963, opened 2023-11-14
fix: remove AllCan Read/Write

MarshallOfSound:
Refs https://chromium-review.googlesource.com/c/v8/v8/+/5006387

Fixes native module compilation against node 21 / incoming v8 changes

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -47,6 +47,7 @@
 #define NODE_18_0_MODULE_VERSION 108
 #define NODE_19_0_MODULE_VERSION 111
 #define NODE_20_0_MODULE_VERSION 115
+#define NODE_21_0_MODULE_VERSION 120
 
 #ifdef _MSC_VER
 # define NAN_HAS_CPLUSPLUS_11 (_MSC_VER >= 1800)
@@ -2517,7 +2518,9 @@ NAN_DEPRECATED inline void SetAccessor(
   , GetterCallback getter
   , SetterCallback setter
   , v8::Local<v8::Value> data
+#if (NODE_MODULE_VERSION < NODE_21_0_MODULE_VERSION)
   , v8::AccessControl settings
+#endif
   , v8::PropertyAttribute attribute
   , imp::Sig signature) {
   HandleScope scope;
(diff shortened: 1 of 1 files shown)


kkoopa:
Thank you. I am concerned about the signature changes. It seems like it should be possible to paper over the differences without breaking the NAN API. Assuming that v8::AccessControl is just a straight enum, it should be replaceable by an int. SetAccessor could be deprecated (again...) and a new one introduced without the settings parameter, shoving default for old versions. Now, due to the bad choice of using default arguments, C++ overload resolution becomes a mess I no longer remember. Might need to make a bunch of overloads for varying number of arguments instead, which is how it should have been written in the first place, since default arguments are bad design and should not be used, but I digress.

MarshallOfSound:
@kkoopa I could move `settings` after `attribute` and default it correctly in all cases, that should avoid the signature being meaningfully different 🤔 Although annoying because it wouldn't match the argument order to v8 but not the end of the world if the nan helper is deprecated anyway

kkoopa:
Yes, that might work. All I want is to avoid breaking NANs API directly while retaining the same observable behavior across versions.

I remember dealing with something similar in one of these default parameter functions previously. I ended up adding n overloads of increasing arity, where n was the number of default parameters, which happens to be the correct way of doing it in the first place. Default parameters in C++ are garbage.

MarshallOfSound:
Turns out the actual enum hasn't been removed, so the nan signature can be the same we just need to safely ifdef the impl.

Ref: https://source.chromium.org/chromium/chromium/src/+/main:v8/include/v8-object.h;l=173;bpv=1;bpt=1?q=v8::AccessControl&ss=chromium

Should be safe now

kkoopa:
Wonderful. That makes it a lot easier. Now, I would still like to have `Nan::SetAccessor` as it is in this PR deprecated and a new one added which does not have the `settings` parameter. Then, for old versions, it would set `settings` to `Default` when forwarding. This serves to maintain the identical behavior across versions: if one can't specify AccessControl, none shall do so. It also signals that a change has happened.

PR #966, opened 2024-02-22
Fix builds for Electron 29

indutny-signal:
`ObjectTemplate::SetAccessor` signature has changed in the V8 >= 12, and no longer accepts `v8::AccessControl` parameter. The `v8::AccessControl` enum is in fact a single value now so no public users of nan would be broken if we just stop passing it down to V8.

### DIFF
--- a/nan.h
+++ b/nan.h
@@ -2550,7 +2550,9 @@ NAN_DEPRECATED inline void SetAccessor(
     , getter_
     , setter_
     , obj
+#if !defined(V8_MAJOR_VERSION) || V8_MAJOR_VERSION < 12
     , settings
+#endif
     , attribute
 #if (NODE_MODULE_VERSION < NODE_16_0_MODULE_VERSION)
     , signature
@@ -2596,7 +2598,9 @@ inline void SetAccessor(
     , getter_
     , setter_
     , obj
+#if !defined(V8_MAJOR_VERSION) || V8_MAJOR_VERSION < 12
     , settings
+#endif
     , attribute
   );
 }


kkoopa:
Thank you. I'll get this out next week when I am home.

indutny-signal:
Yay! Thank you so much!

chrbongardt:
Hey @kkoopa  could we please get a release with this fix <3

nan.h
```diff
@@ -2550,7 +2550,9 @@ NAN_DEPRECATED inline void SetAccessor(
     , getter_
     , setter_
     , obj
+#if !defined(V8_MAJOR_VERSION) || V8_MAJOR_VERSION < 12
```
deepak1556, line 2553: Hi! Electron 28 ships with V8 12.0 and it still supports the `AccessControl` argument. Using nan `2.19.0` will break module compilation for these users. User can workaround by strict versioning to `2.18.0` instead of `^2.18.0`. Should the check be based on module version instead, ex: https://github.com/electron/electron/blob/main/patches/nan/api_remove_allcan_read_write.patch ?

indutny-signal: Oh gosh, I think that'd be better indeed. Did V8 make a breaking API change within the same major release? How come 12.0 isn't the same as 12.2?

deepak1556: Asked about it in the relevant V8 CL https://chromium-review.googlesource.com/c/v8/v8/+/5006387/comment/65b3eecb_f7eb5d8f/

kkoopa:
I can push out a minor update with a fix quickly, but let us wait and see what the V8 response is.

nan.h
LeszekSwirski: I think there's some confusion here about V8's versioning -- V8 doesn't use semver, so major vs minor versions don't carry different meaning for API changes. Rather, for historical reasons, it's synchronised to the chromium release cycle, where V8 12.2 is the V8 version that ships with Chrome 122. It's a bit unfortunate but we're stuck with it for now. We do have a rolling stability guarantee that functions will get a deprecation warning for one release before being removed (see https://v8.dev/docs/api), and this was unfortunately accidentally not followed in this case, but that doesn't prevent API changes across multiple "minor" version bumps.

PR #971, opened 2024-06-26
Change methodswithdata tests line endings to LF

dsanders11:
These two files got committed with CRLF line endings, while the rest of the repo uses LF line endings. This PR changes the line endings in these two files to LF to be consistent.


kkoopa:
I just ran into this. Better late than never. Thanks.

Issue #973, opened 2024-08-21
support for electron 32


steinmetz85:
With electron 32.0.1 and its underlying v8 version 12.8, there are mulitple errors when building NAN modules:

nan.h(700,39): error C2039: "IdleNotificationDeadline" ist kein Member von "v8::Isolate"
nan.h(2560,8): error C2039: "SetAccessor" ist kein Member von "v8::ObjectTemplate"
nan.h(2654,15): error C2039: "SetAccessor" ist kein Member von "v8::Object"
nan.h(2730,56): error C2440: "<function-style-cast>": "initializer list" kann nicht in "v8::NamedPropertyHandlerConfiguration" konvertiert werden
nan_scriptorigin.h(19,1): error C2664: "v8::ScriptOrigin::ScriptOrigin(v8::Local<v8::Value>,int,int,bool,int,v8::Local<v8::Value>,bool,bool,bool,v8::Local<v8::Data>)" : Konvertierung von Argument 1 von "v8::Isolate *" in "v8::Local<v8::Value>" nicht möglich

agracio:
In Electron v31 the following lines appeared as warnings

```
node_modules\nan\nan.h(700,39): warning C4996: 'v8::Isolate::IdleNotificationDeadline' : Use MemoryPressureNotification() to influence the GC schedule. 
node_modules\nan\nan.h(2654,15): warning C4996: 'v8::Object::SetAccessor': Use SetNativeDataProperty instead 
node_modules\nan\nan.h(2730,56): warning C4996: 'v8::NamedPropertyHandlerConfiguration ::NamedPropertyHandlerConfiguration': Provide interceptor callbacks with new signatures instead (NamedPropertyXxxCallback)
node_modules\nan\nan_scriptorigin.h(19,23): warning C4996: 'v8::ScriptOrigin::ScriptOrigin': Use constructor without the isolate. 
node_modules\nan\nan_scriptorigin.h(23,23): warning C4996: 'v8::ScriptOrigin::ScriptOrigin': Use constructor without the isolate. 
node_modules\nan\nan_scriptorigin.h(30,23): warning C4996: 'v8::ScriptOrigin::ScriptOrigin': Use constructor without the isolate. 

```

In Electron v32  it now throws these errors:

```
node_modules\nan\nan.h(700,39): error C2039: 'IdleNotificationDeadline': is not a member of 'v8::Isolate' 
[...]

PR #974, opened 2024-08-24
New breakage

kkoopa:
Resolving some breaking V8 changes reported in #973


kkoopa:
This appears unresolvable due to https://issues.chromium.org/issues/348660658

Issue #936, opened 2022-04-20
`Nan::SetAccessor()` deprecation warning with V8 10.1 and nan 2.15.0


mscdex:
```
../../../../node_modules/nan/nan.h: In function ‘void Nan::SetAccessor(v8::Local<v8::ObjectTemplate>, v8::Local<v8::String>, Nan::GetterCallback, Nan::SetterCallback, v8::Local<v8::Value>, v8::AccessControl, v8::PropertyAttribute, Nan::imp::Sig)’:
../../../../node_modules/nan/nan.h:2551:16: warning: ‘void v8::ObjectTemplate::SetAccessor(v8::Local<v8::Name>, v8::AccessorNameGetterCallback, v8::AccessorNameSetterCallback, v8::Local<v8::Value>, v8::AccessControl, v8::PropertyAttribute, v8::Local<v8::AccessorSignature>, v8::SideEffectType, v8::SideEffectType)’ is deprecated: Do signature check in accessor [-Wdeprecated-declarations]
 2551 |     , signature);
      |                ^
In file included from /home/foo/.cache/node-gyp/18.0.0/include/node/v8-function.h:15,
                 from /home/foo/.cache/node-gyp/18.0.0/include/node/v8.h:33,
                 from /home/foo/.cache/node-gyp/18.0.0/include/node/node.h:63,
                 from ../src/binding.cc:13:
/home/foo/.cache/node-gyp/18.0.0/include/node/v8-template.h:838:8: note: declared here
  838 |   void SetAccessor(
      |        ^~~~~~~~~~~
```

mkrufky:
@mscdex can you confirm that this is resolved?

mkrufky:
@mscdex Can you confirm that this is resolved in the main branch?

alejandroclaro:
There is now a similar problem with Electron >= 20

see: https://github.com/electron/electron/issues/35193

danielweck:
Electron29 / NodeJS 20 breakage:

```
../node_modules/nan/nan.h:2548:8: error: no matching member function for call to 'SetAccessor'
  tpl->SetAccessor(
  ~~~~~^~~~~~~~~~~
/Users/danielweck/Library/Caches/node-gyp/29.0.0/include/node/v8-template.h:806:8: note: candidate function not viable: no known conversion from 'v8::AccessControl' to 'PropertyAttribute' for 5th argument
  void SetAccessor(
       ^
[...]

PR #982, opened 2024-12-10
fix: electron v28 build failure due to setAcessor not having compatible signature

JCMais:
Fixes https://github.com/nodejs/nan/issues/936

Let me know if there is a better way to fix this! This seems to be working for Electron v28, but I did not run the remaining tests I have.


JCMais:
> Note that this issue is also reproducible with nodejs 22.12.0.

hmmm, in that case maybe we should use the v8 version itself in the directive?

agracio:
PR https://github.com/nodejs/nan/pull/1004 fixes Electron 28 build.

PR #932, opened 2022-02-17
package.json: Upgrade node-gyp

cclauss:
Failure to upgrade is causing a ton of pain at https://github.com/nodejs/node-gyp/issues?q=is%3Aissue+label%3A%22ERR%21+node-gyp+-v+%3C%3D+v5.1.0%22+

PR #983, opened 2025-02-05
AppVeyor: Test and support versions 8, 10, 12, 14, 16, 17, 18, 19, 20, 21, and 22

cclauss:
# Tested versions 8, 10, 12, 14, 16, 17, 18, 19, 20, 21, and 22.

Test supported Node.js versions before unsupported versions because each test takes ~5 minutes.
Test on the most current Visual Studio possible for each nodejs_version.
Test on the oldest ___supported___ Python version when possible (currently py39).
For all supported Node.js versions also test on the newest Python (currently py313).
* https://devguide.python.org/versions
# Test results: https://ci.appveyor.com/project/RodVagg/nan

Visual Studio 2017 image tops out at Python 3.8 which is end-of-life.
Visual Studio 2019 and 2022 images top out at Python 3.13.

`node-gyp` issues related to `nan`: https://github.com/nodejs/node-gyp/issues?q=label%3Anodejs%2Fnan

The Node.js v23 tests are commented out waiting for:
* #979 or similar.

~The Python 3.12 and 3.13 tests are commented out waiting for:~
*  ~#985 because of:~
    * ~nodejs/node-gyp#2869~

@kkoopa Your review, please.  These changes enable us to focus on the two issues above.

PR #985, opened 2025-02-07
package.json: Upgrade node-gyp to support Python >= v3.12

cclauss:
https://github.com/nodejs/node-gyp?tab=readme-ov-file#installation says
> [!Important]
> Python >= v3.12 requires `node-gyp` >= v10

Fixes: #986
* #986

`package.json`:
```diff
-    "node-gyp": "~8.4.1",
+    "node-gyp": "~v10.3.1",
```
Provides compatibility with Python 3.12, 3.13, and later.

https://github.com/nodejs/node-gyp/releases
* nodejs/node-gyp#2869
* https://devguide.python.org/versions

Related to:
* #932

@mkrufky Your review, please.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "commander": "^2.8.1",
     "glob": "^5.0.14",
     "request": "=2.81.0",
-    "node-gyp": "~8.4.1",
+    "node-gyp": "~v10.3.1",
     "readable-stream": "^2.1.4",
     "tap": "~0.7.1",
     "xtend": "~4.0.0"


kkoopa:
Looking at the test output from this, there are a lot of failures on ancient versions. Am I correct in assuming that this is expected. The reason for keeping as old a node-gyp as possible has been to enable testing on these ancient versions. I am prepared to drop them from testing if necessary.

cclauss:
Let's move forward with #987 which adds the GitHub Actions test.   Merging that one will autoclose this PR and two others.

Once this is merged, we can then rework the AppVeyor tests in #983 to provide guidance on older versions of Node.js

PR #993, opened 2025-03-13
Let's see if CI passes on Node.js v23

cclauss:
@agracio @kkoopa Your review, please.


agracio:
I am not a member of `nan` team 
I only participate in topics that are related to Electron builds 😄

agracio:
Personally I use nan to build Node.js and  node-electron native modules for my projects. Building Node.js v23 native module works perfectly fine but that does not mean that there no underlying issues just like the failing tests are showing.

PR #979, opened 2024-10-18
Support Node 23

kkoopa:
V8 Was bumped, this would be a patch release, since I already claimed support for Node 23 in 2.22.0.
Also fixes #978.

PR #997, opened 2025-05-08
test node 24

nigrosimone:


cclauss:
Related to:
* #979
* #993

kkoopa:
The PR itself looks good, but is there any point of testing on Node 24 if not even 23 is working?

nigrosimone:
I use latest version on node 23 and works for me 🤨

cclauss:
@kkoopa Please enable the tests to run so we can see the results.  In 6 months v24 will become the LTS release so we should encourage the community to get these tests to pass.

kkoopa:
That is an excellent point. Will do.

PR #1002, opened 2025-09-12
use ArrayBuffer->Data() when available

zbjornson:
This is faster than accessing the BackingStore. API is available in V8 10.5+ per https://github.com/nodejs/node/issues/32226#issuecomment-1185423020.

Manually tested in Node.js v18.20 and v20.

### DIFF
--- a/nan_typedarray_contents.h
+++ b/nan_typedarray_contents.h
@@ -31,9 +31,13 @@ class TypedArrayContents {
       v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
 
       length = byte_length / sizeof(T);
+#if (V8_MAJOR_VERSION >= 10 ||                                                 \
+     (V8_MAJOR_VERSION == 10 &&                                                \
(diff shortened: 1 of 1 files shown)


nan_typedarray_contents.h
```diff
@@ -31,9 +31,14 @@ class TypedArrayContents {
       v8::Local<v8::ArrayBuffer> buffer = array->Buffer();
 
       length = byte_length / sizeof(T);
+#if (V8_MAJOR_VERSION >= 10 ||                                                 \
+     (V8_MAJOR_VERSION == 10 &&                                                \
+      (defined(V8_MINOR_VERSION) && V8_MINOR_VERSION >= 5)) ||                 \
+       (NODE_MAJOR_VERSION == 18 && NODE_MINOR_VERSION >= 8))
```
Flarna, line 37: I think this means a native addon built with 18.8.0 or newer wont work with older 18 node.js versions.
Does this fit the NAN compatibilty schema?
Maybe go for >= 20?

PR #1022, opened 2026-09-14
fix: support V8 15.3 and later

codebytere:
Building against V8 15.3 or later fails in two places. `PropertyCallbackInfo::HolderV2()` and `Object::SetPrototypeV2()` are gone: 14.7 renamed them back to `Holder()` / `SetPrototype()` (https://crrev.com/c/7624359) and 15.3 removed the V2 names (https://crrev.com/c/8234700). And `SetNativeDataProperty` plus the named/indexed property handler configurations only take setter callbacks with `const PropertyCallbackInfo<v8::Boolean>&` now: those were added in 14.8 (https://crrev.com/c/7689002, https://crrev.com/c/7693693) and the `PropertyCallbackInfo<void>` overloads went away in 15.3 (https://crrev.com/c/8236212).

This uses `Holder()` / `SetPrototype()` from 14.7 on (14.6 and earlier 14.x keep the V2 calls), and adds an `imp::PropertyCallbackValue<T>` mapping so that `Nan::PropertyCallbackInfo<void>` and `Nan::ReturnValue<void>` wrap the `v8::Boolean`-typed V8 objects on 14.8+, with the accessor, named and indexed setter wrappers declared with the matching V8 signature. Nothing changes for addons: setters still receive `Nan::PropertyCallbackInfo<void>`, and `info.GetReturnValue().Set(...)` in a setter stays the no-op it has been since 14.6.

Tests: the `test/` addons build with no new warnings and the JS tests pass against Node.js 18.20.8, 20.20.2, 22.23.2, 24.21.0, 26.8.2 and 27.0.0-v8-canary20260904 (V8 15.5.8); on that canary the unmodified tree fails with exactly the two errors above. cpplint is clean with the repo's pre-commit settings.

---

Disclosure: the code and this description were written by Claude Code, directed and reviewed by @codebytere.


kkoopa:
Thank you, this looks good and all the tests seem to pass. I will make a new release.

